PR #5, opened 2017-03-27
implement quota setting

majewsky:
- [x] Scrapers should try to rectify a frontend/backend quota mismatch.
  - [x] add tests
- [x] implement PUT on project
  - [x] add tests
- [x] implement PUT on domain
  - [x] add tests

**DO NOT MERGE** without my consent.

### DIFF
--- a/pkg/api/api_test.go
+++ b/pkg/api/api_test.go
@@ -35,8 +35,8 @@ import (
 )
 
 func init() {
-	limes.RegisterQuotaPlugin(&test.Plugin{StaticServiceType: "shared"})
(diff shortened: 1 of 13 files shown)


majewsky:
This is good to go from the Limes side, but there is a sort of race condition in there if we merge this like it is: If Elektra updates the quota in the backend, and then Limes scrapes that project again *before* pulling the new approved quota from Elektra, Limes will write the outdated approved quota back into the backend.

This can be rectified by adjusting Elektra such that it pushes approved quotas into Limes immediately, thus avoiding the inconsistent state between both databases.

PR #22, opened 2017-04-20
initial manila quota plugin

auhlig:
First take on manila quota plugin.

### DIFF
--- a/pkg/plugins/manila.go
+++ b/pkg/plugins/manila.go
@@ -87,7 +87,7 @@ func (p *manilaPlugin) Scrape(driver limes.Driver, domainUUID, projectUUID strin
 	}
 
 	var result gophercloud.Result
-	var totalShareUsage, totalSnapshotUsage, totalShareNetworksUsage = int64(0), 0, 0
+	var totalShareUsage, totalSnapshotUsage, totalShareNetworksUsage = uint64(0), uint64(0), uint64(0)
 
 	//Get absolute quota limits per project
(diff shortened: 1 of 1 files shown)


pkg/plugins/manila.go
```diff
@@ -0,0 +1,213 @@
...
+	if err != nil {
+		return nil, err
+	}
+	var manilaQuotaData struct {
+		QuotaSet struct {
+			Gigabytes         int `json:"gigabytes"`
+			Shares            int `json:"shares"`
+			SnapshotGigabytes int `json:"snapshot_gigabytes"`
+			Snapshots         int `json:"snapshots"`
+			ShareNetworks     int `json:"share_networks"`
```
majewsky: Please use `int64` here instead of `int` to avoid the cast at the end.

pkg/plugins/manila.go
```diff
@@ -0,0 +1,213 @@
...
+	//Get usage of shares per project
+	url = client.ServiceURL("shares", "detail") + fmt.Sprintf("?project_id=%s&all_tenants=1", projectUUID)
+	_, err = client.Get(url, &result.Body, nil)
+	if err != nil {
+		return nil, err
+	}
+
+	var manilaShareUsageData struct {
+		Shares []struct {
+			Size int `json:"size"`
```
majewsky: `uint64` instead of `int` for the same reason (same for the other size summations down below)

auhlig:
Done @majewsky. Please review again.

PR #24, opened 2017-04-20
ignore baremetal flavors in nova capacity plugin

### LINKED ISSUE #23: narrow down flavor list in nova capacity plugin
Use-case: Consider only VM flavors, not baremetal flavors.

For each flavor, query Nova for `GET /v2/:tenant_id/flavors/:flavor_id/os-extra_specs`. Returns a JSON like 

```go
type ExtraSpecsBody struct {
  ExtraSpecs map[string]string `json:"extra_specs"`
}
```

Extend the `Nova` subkey of the CapacitorConfiguration with a `map[string]string` such that only filters with the configured extra specs are considered for calculating the maximum ephemeral disk size.

auhlig:
closes #23

### DIFF
--- a/pkg/plugins/capacity_nova.go
+++ b/pkg/plugins/capacity_nova.go
@@ -27,10 +27,15 @@ import (
 	"github.com/gophercloud/gophercloud/openstack/compute/v2/flavors"
 	"github.com/gophercloud/gophercloud/pagination"
 	"github.com/sapcc/limes/pkg/limes"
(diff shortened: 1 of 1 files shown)


auhlig:
You mean evaluate if the extra specs contain `{ "vmware:hv_enabled" : 'True' }` instead of the current check or what else should be done?

majewsky:
Yes, with the concrete keys and values being a configuration parameter.

PR #25, opened 2017-04-26
configurable extra specs in nova capacity plugin

### LINKED ISSUE #23: narrow down flavor list in nova capacity plugin

auhlig:
fixes #23

### DIFF
--- a/pkg/plugins/capacity_nova.go
+++ b/pkg/plugins/capacity_nova.go
@@ -24,6 +24,7 @@ import (
 
 	"strings"
 
+	"fmt"
 	"github.com/gophercloud/gophercloud"
 	"github.com/gophercloud/gophercloud/openstack"
 	"github.com/gophercloud/gophercloud/openstack/compute/v2/flavors"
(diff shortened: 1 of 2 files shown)


pkg/plugins/capacity_nova.go
```diff
@@ -114,10 +116,18 @@ func (p *capacityNovaPlugin) Scrape(driver limes.Driver) (map[string]map[string]
...
-			if len(extras.ExtraSpecs) != 0 {
-				maxFlavorSize = math.Max(maxFlavorSize, float64(element.Disk))
+			//consider only flavors as defined in extra specs
+			var extraSpecs map[string]string
+			if p.cfg.Nova.ExtraSpecs != nil {
+				extraSpecs = p.cfg.Nova.ExtraSpecs
+			}
+
+			for key, value := range extraSpecs {
+				if value == strings.Replace(extras.ExtraSpecs[key], "'", "", -1) {
```
majewsky: I'd rather keep the `extras.ExtraSpecs[key]` as-is, to avoid surprises.

PR #26, opened 2017-04-27
add swift capacity plugin

### LINKED ISSUE #18: capacity plugin for Swift
- Query Prometheus to fetch the capacity metric submitted by swift-health-statsd. (An appropriate plugin ID would be `swift-health-statsd`.)
- Get Prometheus server URL from config.yaml.
- Maybe the Prometheus client library already has some facility to execute PromQL queries and parse the responses?

auhlig:
fixes #18

### DIFF
--- a/pkg/plugins/capacity_swift.go
+++ b/pkg/plugins/capacity_swift.go
@@ -1,0 +1,99 @@
+/*******************************************************************************
+*
+* Copyright 2017 SAP SE
+*
+* Licensed under the Apache License, Version 2.0 (the "License");
+* you may not use this file except in compliance with the License.
+* You should have received a copy of the License along with this
+* program. If not, you may obtain a copy of the License at
+*
+*     http://www.apache.org/licenses/LICENSE-2.0
+*
+* Unless required by applicable law or agreed to in writing, software
+* distributed under the License is distributed on an "AS IS" BASIS,
+* WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+* See the License for the specific language governing permissions and
+* limitations under the License.
+*
+*******************************************************************************/
+
+package plugins
+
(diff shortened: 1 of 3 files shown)


pkg/plugins/capacity_swift.go
```diff
@@ -0,0 +1,99 @@
...
+	client, err := prometheus.New(config)
+	if err != nil {
+		util.LogDebug("Could not create Prometheus client with URL: %s", prometheusAPIURL)
+		return nil, err
+	}
+	return prometheus.NewQueryAPI(client), nil
+}
+
+func (p *capacitySwiftPlugin) ID() string {
+	return "swift-health-statsd"
```
majewsky: The plugin ID looks good, but the filename and class name should reflect that this capacity plugin is specifically targeted at swift-health-statsd, rather than working with every Swift out there.

pkg/plugins/capacity_swift.go
```diff
@@ -0,0 +1,99 @@
...
+	var value model.Value
+	var resultVector model.Vector
+
+	value, err = client.Query(context.Background(), "min(swift_cluster_storage_capacity_bytes_gauge < inf)", time.Now())
+	if err != nil {
+		return nil, err
+	}
+	resultVector, ok := value.(model.Vector)
+	if !ok {
+		util.LogDebug("Could not get value for query min(swift_cluster_storage_capacity_bytes_gauge < inf) from Prometheus")
```
majewsky: 1. I think this can easily be a `LogError`.
2. Consider de-duplicating the query string into a variable and interpolating it here as `%s`.
3. The log message should indicate that the problem is a type mismatch (all other errors are caught in the `err != nil` check).

auhlig:
Makes sense. Done @majewsky.

PR #30, opened 2017-05-10
cinder: add filter for volume_backend_name

auhlig:
Checklist:

### DIFF
--- a/pkg/plugins/capacity_cinder.go
+++ b/pkg/plugins/capacity_cinder.go
@@ -54,6 +54,11 @@ func (p *capacityCinderPlugin) Scrape(driver limes.Driver) (map[string]map[strin
 	}
 
 	var result gophercloud.Result
+	var volumeBackendName string
+
+	if p.cfg.Cinder.VolumeBackendName != "" {
+		volumeBackendName = p.cfg.Cinder.VolumeBackendName
+	}
 
 	//Get absolute limits for a tenant
 	url := client.ServiceURL("scheduler-stats", "get_pools") + "?detail=True"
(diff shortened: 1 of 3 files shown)


pkg/plugins/capacity_cinder.go
```diff
@@ -54,18 +54,25 @@ func (p *capacityCinderPlugin) Scrape(driver limes.Driver) (map[string]map[strin
 	}
 
 	var result gophercloud.Result
+	var volumeBackendName string
+
+	if p.cfg.Cinder.VolumeBackendName != "" {
+		volumeBackendName = p.cfg.Cinder.VolumeBackendName
+	}
```
majewsky: Or just `volumeBackendName := p.cfg.Cinder.VolumeBackendName`.

pkg/plugins/capacity_cinder.go
```diff
@@ -97,7 +104,11 @@ func (p *capacityCinderPlugin) Scrape(driver limes.Driver) (map[string]map[strin
 
 	//add results from scheduler-stats
 	for _, element := range limitData.Pools {
-		totalCapacity += uint64(element.Capabilities.TotalCapacity)
+		if (volumeBackendName != "") && (element.Capabilities.VolumeBackendName == volumeBackendName) {
+			totalCapacity += uint64(element.Capabilities.TotalCapacity)
+			util.LogDebug("Considering %s with volume_backend_name %s", element.Name, element.Capabilities.VolumeBackendName)
+		}
+		util.LogDebug("Not considering %s with volume_backend_name %s", element.Name, element.Capabilities.VolumeBackendName)
```
majewsky: This skips all volume types if the `volume_backend_name` is not configured. Also, it prints the `not considering` message even when considering the volume type.

auhlig: I need more coffee

auhlig:
Better now @majewsky?

PR #33, opened 2017-05-19
Resolving flavors for compute instance subresource

reimannf:
Basicly we need to discuss 2 things: 
* Flavors are cached and not invalidated, as changing **should** not be possible
* Access to cache is not thread safe as the current assumption is that the nova plugin will only scrape on project at once

Checklist:

### DIFF
--- a/docs/operators/config.md
+++ b/docs/operators/config.md
@@ -150,9 +150,9 @@ The `instances` resource supports subresource scraping. Subresources bear the fo
 | `id` | string | nova server UUID |
 | `name` | string | nova server name |
 | `status` | string | https://wiki.openstack.org/wiki/VMState |
-| `ram` | string | ram configured in flavor |
-| `vcpu` | string | vcpu configured in flavor |
-| `disk` | string | root disk size configured in flavor |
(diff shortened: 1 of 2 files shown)


auhlig:
To stop the CI from complaining do a `gofmt pkg/plugins/nova.go`. In gogland right click on file, go-tools, `go fmt file`. In case of additional imports `goimports pkg/plugins/nova.go` @reimannf

docs/operators/config.md
```diff
@@ -147,10 +147,12 @@ The `instances` resource supports subresource scraping. Subresources bear the fo
...
-| `id` | string ||
-| `name` | string ||
-| `status` | string ||
-| `flavor_id` | string | provisional; will be replaced by `cores` and `ram` attributes soon |
+| `id` | string | nova server UUID |
+| `name` | string | nova server name |
+| `status` | string | https://wiki.openstack.org/wiki/VMState |
+| `ram` | int | ram configured in flavor |
+| `vcpu` | int | vcpu configured in flavor |
+| `disk` | int | root disk size configured in flavor |
```
majewsky: Which unit?

PR #35, opened 2017-06-01
Relabel metrics from `cluster` to `os_cluster`

reimannf:
It turned out that cluster is too generic for the configured OpenStack
clusters. E.g Prometheus federation might add the same label as well.
To avoid further confusions rename the label to os_cluster.

Checklist:

~~- [ ] If this PR is about a plugin, I tested the plugin against an OpenStack cluster.~~


majewsky:
Muss ja, ne? :)

PR #43, opened 2018-01-25
add capacity plugin for Manila

majewsky:
According to @Carthaca's specifications.

Checklist:


majewsky:
Maurice gave a 👍 in Slack, so pushing this forward.

PR #47, opened 2018-02-28
[capacity_manila] filter by default share-type

Carthaca:
I just copied the way of providing a url param from https://github.com/sapcc/limes/blob/05ecf7cde6734a445cf2614fd8fec346ae89474e/pkg/plugins/cinder.go#L92 and extracted the param from the python-manilaclient api debug

i.e. not tested, but I'm quite certain it works if there are no syntax errors


majewsky:
Ok sure. Please test in staging and press the deploy-all button in the pipeline if it looks good.

PR #51, opened 2018-05-28
proposal: rate limit api extension

auhlig:
As just discussed, this is the API extension proposal for rate limits to limes
Introduces the new field `rates` which hold information for project specific rate limits and resides on the same level as `resources`, as a rate limit definition differs from the quota/usage definition.
Also would allow us to limit the response just to rates or resources later on.
A rate limit would consists of the `targetTypeURI` and a list of `actions` with their `limit`.  Both keys as per [CADF standard](https://www.dmtf.org/sites/default/files/standards/documents/DSP2038_1.1.0.pdf). 
See below example.
Initial Elektra plugin for configuring rate limits might be based an a searchable list of `targetTypeURI`s and pre-configured `limits`.
If you're okay with this spec, I'd go further and start implementation.

What do you think @majewsky @reimannf?


auhlig:
Defaults would be configured per cluster via limes seed: $secrets/master/global/values/limes.yaml . 
They apply to each and every project in the cluster. Configuring it per domain might also make sense, so we'd have a hierarchy of cluster, domain, project, in which (a) project limits would overwrite domain, cluster limits or (b) project limits can never exceed domain or cluster limits. (cluster/domain limit as global limit to protect backend)
Good point. Introducing `backend_limit` to indicate a mismatch between limes and backend makes sense IMO.

majewsky:
> Good point. Introducing `backend_limit` to indicate a mismatch between limes and backend makes sense IMO.

What does "mismatch" mean here? The backends pull the rate limits from Limes, so there is no deviation over time. And to be able to report a `backend_limit`, we would need to have an API to query the rate limit in the backend, which is not planned afaik.

What I think Falk meant was a `default_limit` field which behaves towards `limit` in the same manner as `backend_quota` towards `quota`: It only appears when the values differ.

auhlig:
I understood it the same way. I guess "mismatch" was the wrong wording. "deviation" describes it better.

reimannf:
Ok. So we agree to seed defaults on cluster level. Project level deviations from the default will result in an additional field `default_limit` to indicate that.

Can we scratch the idea of domain rate limits for now?

auhlig:
Agreed. Domain rate limits are dismissed for now.

auhlig:
travis' gofmt hates me. Haven't touched these files and local gofmt is happy @majewsky 
```
 => gofmt
pkg/audit/rabbit.go
pkg/collector/scrape_test.go
```

pkg/reports/project.go
```diff
@@ -175,77 +262,84 @@ func GetProjects(cluster *limes.Cluster, domainID int64, projectID *int64, dbi d
...
+						Limit: *limit,
+					}
+				}
+				rate.Actions[*action].Name = *action
+				rate.Actions[*action].Limit = *limit
+
+				// show only rates
+				if filter.onlyRates {
+					service.Resources = nil
+				}
```
majewsky: Putting this here means that `?rates=only` only works correctly for services that have rate limits. If a service only has resources, you will never get to this point, so the resources would end up in the result even for `?rates=only`.

I suggest you wrap the entire first query in `if !filter.onlyRates { ... }` and adjust the second query such that it also collects all the auxiliary attributes that the first query populates (project name, parent UUID, scraping timestamps).

pkg/reports/project.go
```diff
@@ -256,8 +350,120 @@ func GetProjects(cluster *limes.Cluster, domainID int64, projectID *int64, dbi d
...
+
+	result := make(ProjectRateLimits, len(limits))
+	for _, ttu := range targetTypeURIs {
+		limits[ttu].Actions = sortRateLimitActions(limits[ttu].Actions)
+		result[ttu] = limits[ttu]
+	}
+	return result
+}
+
+func sortRateLimitActions(actions RateLimitActions) RateLimitActions {
```
majewsky: This attempts to sort a map, but Go maps AFAIK are not defined to return their elements in insertion order upon iteration.

auhlig:
TODO:
- [ ] add `default_limit` field to indicate deviation from default rate limits

auhlig:
rebased on master pulling in latest changes regarding audit

auhlig:
squashed into single commit

pkg/reports/project.go
auhlig: done

pkg/reports/project.go
auhlig: yep 🤦‍♂

auhlig:
Note for tomorrows code review: 
- Expose service-specific [global rate limits](https://github.com/sapcc/limes/blob/ratelimits/pkg/core/config.go#L135) via `GET /v1/clusters/:cluster_id`?

docs/users/api-v1-specification.md
```diff
@@ -200,6 +229,65 @@ exists and is non-zero), then resources with `usage > quota` will display an add
...
 `usage` values directly instead of using this field.
 
+### Rate limits
+
+ If the `?rates` query parameter is given, rate limits for the service are included in the response.
+Additionally the response can be limited to only include rate limits by providing the query parameter `?rates=only`.
+Rate limits can be used to control the traffic received by an API in order to prevent service capacities from being exhausted.
+They can be set on 2 levels:
+(1) On *cluster* level in order to ensure a service does not receive more request than it can handle
+(2) Per *project* in order to ensure a fair usage among projects within a cluster
```
majewsky: This is not how enumerations work in Markdown. Needs to be:

1. like
2. this

docs/users/api-v1-specification.md
```diff
@@ -200,6 +229,65 @@ exists and is non-zero), then resources with `usage > quota` will display an add
...
+ A prerequisite to traffic control is a normative scheme of classification for actions initiated by users.
+The [CADF specification](https://www.dmtf.org/sites/default/files/standards/documents/DSP2038_1.1.0.pdf) established such a classification for the OpenStack ecosystem.
+In terms of rate limits, requests sent to and OpenStack API are characterized by their `target_type_uri` and `action`.
+The `target_type_uri` represents the request path against which the activity was performed and the `action` the activity that was performed.
+
+ A rate limit entity in limes, as shown at the end of this subsection, consist of a `target_type_uri`, a list of `actions`, a `limit` and a `defaultLimit`.
+The field `limit` defines the maximum number of requests within a rate limit window per project.
+While `defaultLimits` is used to reflect the default rate limit set on cluster level.
+
+ A rate limit entity in limes, as shown at the end of this subsection, consist of a `target_type_uri` which characterizes the request path and a list of `actions`.
```
majewsky: This sentence is duplicate with the last paragraph.

pkg/reports/filter.go
```diff
@@ -32,6 +33,21 @@ import (
...
+//NewFilter returns a new Filter.
+func NewFilter(serviceTypes, resourceNames []string, withRates, onlyRates, withSubresources bool) Filter {
+	return Filter{
+		serviceTypes:     serviceTypes,
+		resourceNames:    resourceNames,
+		withRates:        withRates,
+		onlyRates:        onlyRates,
+		withSubresources: withSubresources,
+	}
 }
```
majewsky: I'd rather have callsites just construct a Filter literal, so that the field names are seen explicitly.

pkg/api/projects.go
```diff
@@ -279,42 +273,73 @@ func (p *v1Provider) putOrSimulatePutProjectQuotas(w http.ResponseWriter, r *htt
...
+					continue
+				}
+
+				rl.Limit = req.NewValue
+				rl.Unit = string(req.NewUnit)
+				rateLimitsToUpdate = append(rateLimitsToUpdate, &rl)
+				servicesToUpdate[srv.Type] = true
+			}
 		}
 	}
```
majewsky: This only updates existing rate limits. You also need to create missing rate limits. (This was not a concern for resources because those are all created by the scraper already.)

pkg/api/quota_updater.go
```diff
@@ -230,6 +284,12 @@ func (u QuotaUpdater) validateQuota(srv limes.ServiceInfo, res limes.ResourceInf
 	return u.validateProjectQuota(domRes, *projRes, newQuota)
 }
 
+func (u QuotaUpdater) validateRateLimit(srv limes.ServiceInfo, rateLimit *limes.ProjectRateLimitActionReport, newRateLimitValue uint64, newRateLimitUnit limes.Unit) *core.QuotaValidationError {
+	oldLimit := uint64(0)
+	oldUnit := limes.UnitRequestsPerHour
+	return u.validateAuthorizationRateLimit(oldLimit, newRateLimitValue, oldUnit, newRateLimitUnit)
+}
+
```
majewsky: Should only allow to set rate limits that are already mentioned in the config (i.e. only existing targettypeuri/action pairs).

pkg/api/quota_updater.go
```diff
@@ -230,6 +284,12 @@ func (u QuotaUpdater) validateQuota(srv limes.ServiceInfo, res limes.ResourceInf
 	return u.validateProjectQuota(domRes, *projRes, newQuota)
 }
 
+func (u QuotaUpdater) validateRateLimit(srv limes.ServiceInfo, rateLimit *limes.ProjectRateLimitActionReport, newRateLimitValue uint64, newRateLimitUnit limes.Unit) *core.QuotaValidationError {
+	oldLimit := uint64(0)
+	oldUnit := limes.UnitRequestsPerHour
```
majewsky: Why is the old value always `0r/h`?

pkg/api/quota_updater.go
```diff
@@ -366,6 +473,24 @@ func (u QuotaUpdater) WriteSimulationReport(w http.ResponseWriter) {
 		return resName1 < resName2
 	})
 
+	sort.Slice(result.UnacceptableRateLimits, func(i, j int) bool {
+		srvType1 := result.UnacceptableRateLimits[i].ServiceType
+		srvType2 := result.UnacceptableRateLimits[j].ServiceType
+		if srvType1 != srvType2 {
+			return srvType1 > srvType2
```
majewsky: This should probably be `<` instead of `>`.

pkg/api/quota_updater.go
```diff
@@ -478,4 +617,51 @@ func (u QuotaUpdater) CommitAuditTrail(token *gopherpolicy.Token, r *http.Reques
...
+						"action":          action,
+					}
+					if u.ScopeType() == "project" {
+						if invalid {
+							lowPrivilegeRaiseProjectFailureCounter.With(labels).Inc()
+						} else {
+							lowPrivilegeRaiseProjectSuccessCounter.With(labels).Inc()
+						}
+					}
+				}
```
majewsky: We don't allow low-privilege raise for rate limits as of yet.

pkg/reports/project.go
```diff
@@ -31,136 +31,164 @@ import (
...
+				resReport.Annotations = behavior.Annotations
 
-			if resourceName != nil {
-				if cluster.HasResource(*serviceType, *resourceName) {
-					subresourcesValue := ""
-					if subresources != nil {
-						subresourcesValue = *subresources
-					}
+				if clusterCanBurst {
+				}
```
majewsky: Can be removed.

pkg/reports/project.go
```diff
@@ -31,136 +31,164 @@ import (
...
+				if projectHasBursting && clusterCanBurst && quota != nil && usage != nil {
+					if *usage > *quota {
+						resReport.BurstUsage = *usage - *quota
 					}
-					if projectHasBursting && clusterCanBurst && quota != nil && usage != nil {
-						if *usage > *quota {
-							resource.BurstUsage = *usage - *quota
+				}
+				serviceReport.Resources[*resourceName] = resReport
```
majewsky: This line can probably be deleted. It was already done by Find().

pkg/reports/project.go
```diff
@@ -31,136 +31,164 @@ import (
...
-					}
-					project.Services[*serviceType] = service
+			if resReport != nil && resourceName != nil {
+				subresourcesValue := ""
+				if subresources != nil {
+					subresourcesValue = *subresources
 				}
-			}
+				behavior := cluster.BehaviorForResource(*serviceType, *resourceName, domain.Name+"/"+projectName)
+				resReport.Scaling = behavior.ToScalingBehavior()
```
majewsky: This is already done by Find().

pkg/reports/project.go
```diff
@@ -31,136 +31,164 @@ import (
...
+				resReport.PhysicalUsage = physicalUsage
+				resReport.BackendQuota = nil //See below.
+				resReport.Subresources = limes.JSONString(subresourcesValue)
+				resReport.Annotations = behavior.Annotations
```
majewsky: This is already done by Find().

docs/users/api-v1-specification.md
auhlig: ✅

docs/users/api-v1-specification.md
auhlig: ✅

pkg/reports/filter.go
auhlig: ✅

pkg/api/quota_updater.go
auhlig: ✅

pkg/reports/project.go
auhlig: ✅

pkg/api/quota_updater.go
auhlig: ✅

pkg/reports/project.go
auhlig: ✅

pkg/reports/project.go
auhlig: ✅

pkg/reports/project.go
auhlig: ✅

pkg/api/quota_updater.go
auhlig: ✅

pkg/api/quota_updater.go
auhlig: ✅

pkg/api/projects.go
auhlig: Note for code review: See [here](https://github.com/sapcc/limes/blob/920b0f834583265c6e3e8511e14847048e5275ab/pkg/api/projects.go#L313-L340).

auhlig: ✅

auhlig:
This is ready for review @majewsky 🙏 

Also added some tests for `PUT /v1/domains/$domain/projects/$project` requests (expected success + failure) [here](https://github.com/sapcc/limes/blob/2f7e47dc90868ce9db5e1c16e28835884efcaf2f/pkg/api/api_test.go#L1450-L1544).

pkg/api/quota_updater.go
```diff
@@ -261,6 +343,25 @@ func (u QuotaUpdater) validateAuthorization(oldQuota, newQuota, lprLimit uint64,
...
+func (u QuotaUpdater) validateAuthorizationRateLimit(oldLimit, newLimit uint64, oldUnit, newUnit limes.Unit) *core.QuotaValidationError {
+	if oldLimit >= newLimit || oldUnit.IsGreaterThanOrEqual(newUnit) {
+		if u.CanLower {
+			return nil
+		}
+		return &core.QuotaValidationError{
+			Status:  http.StatusForbidden,
+			Message: "user is not allowed to lower rate limits",
+		}
+	}
```
majewsky: This logic would consider a change from `1r/s` to `1000r/m` a decrease because `r/s >= r/m`. Is that what we want?

pkg/reports/project.go
```diff
@@ -31,136 +31,187 @@ import (
...
+			}
+			return nil
+		})
+		if err != nil {
+			return nil, err
+		}
+
+		//Enrich the report with the default rate limits.
+		for _, projectReport := range projects {
+			for _, serviceReport := range projectReport.Services {
```
majewsky: Request for a test-case: With `?rates=only`, if a project uses the default rate limits everywhere, I believe that this right now outputs an entirely empty report. There are no `project_rate_limits` entries, so `projectReport.Services` is empty in this line, which will cause this loop to be skipped over.

Ideas:

- With `?rates=only`, do not skip the `projectReportQuery`, but set a filter to ensure that only services, but no resources are returned, like here: https://github.com/sapcc/limes/blob/33cae537ba4d87c6da50a53bf6f1bd884a0fdae9/pkg/api/api_test.go#L1067-L1072 This would ensure that the services are all initialized when we enter into the `projectReportRateLimitQuery`. This comment also applies to `reports.GetClusters`. Over there, this approach completely removes the need for the newly-added query 6.
- Consider filling the default rate limits in `projects.Find` when a service is first created. Then have `projectReportRateLimitQuery` overwrite the default rate limits with the project's custom values as necessary.

pkg/api/projects.go
```diff
@@ -279,42 +273,71 @@ func (p *v1Provider) putOrSimulatePutProjectQuotas(w http.ResponseWriter, r *htt
...
-			res.Quota = req.NewValue
-			resourcesToUpdate = append(resourcesToUpdate, &res)
-			servicesToUpdate[srv.Type] = true
+			for targetTypeURI, reqs := range rateLimitRequests {
+				for action, rl := range reqs {
+					dbRateLimit := searchProjectRateLimit(rateLimits, targetTypeURI, action)
+					if dbRateLimit == nil {
+						dbRateLimit = &db.ProjectRateLimit{
+							TargetTypeURI: targetTypeURI, Action: action, ServiceID: srv.ID,
+						}
```
majewsky: This won't work because you're still using UPDATE. The update will succeed, but without affecting any rows. You need an INSERT here, i.e. `_, err := dbi.Insert(newRateLimit)`.

Request for a test-case: PUT a rate limit where no `project_rate_limits` entry does not exist yet.

pkg/reports/cluster.go
```diff
@@ -426,10 +467,10 @@ func makeClusterFilter(tableWithClusterID string, clusterID *string) map[string]
 
 type clusters map[string]*limes.ClusterReport
 
-func (c clusters) Find(config core.Configuration, clusterID string, serviceType, resourceName *string) (*limes.ClusterReport, *limes.ClusterServiceReport, *limes.ClusterResourceReport) {
+func (c clusters) Find(config core.Configuration, clusterID string, serviceType, resourceName, targetTypeURI, action *string) (*limes.ClusterReport, *limes.ClusterServiceReport, *limes.ClusterResourceReport, *limes.ClusterRateLimitReport) {
```
majewsky: I would not do rate-limit stuff in here. The purpose of those `Find()` methods is to reduce code duplication, but there is only one spot where rate limits get created in `GetClusters`.

pkg/api/quota_updater.go
auhlig: Comparing rate limits is tricky as units cannot be converted: `1 r/m != 60 r/s`.

A possible solution would be to not allow users to change the unit from the default one but only the value.

pkg/reports/cluster.go
auhlig: I moved it there, since I'm using it 2 times in `GetClusters(..)` to increase readability and reduce code duplication

majewsky: I'm only seeing one instance that's not giving `nil, nil` in the trailing arguments:
```
$ grep -n '\.Find' pkg/reports/cluster.go | grep -v 'nil, nil)'
429:                                                    _, _, _, rateLimitReport := clusters.Find(config, clusterConfig.ID, &svcConfig.Type, nil, &rateLimit.TargetTypeURI, &action.Name)
```

pkg/api/quota_updater.go
majewsky: Another option: Do not have separate "raise" and "lower" permissions altogether. We could introduce a new oslo.policy rule like "project:set_rate_limit".

pkg/reports/project.go
auhlig: This works already and there's a test for a `GET /v1/domain/$domainID/projects/$projectID?rates=only` where no project specific rate limits are set and only the default rate limits are returned. See [test](https://github.com/sapcc/limes/blob/ratelimits/pkg/api/api_test.go#L1123-L1129) and [response](https://github.com/sapcc/limes/blob/ratelimits/pkg/api/fixtures/project-get-paris-only-default-rates.json).

pkg/api/quota_updater.go
auhlig: I'd like that even more. Going with that.

pkg/reports/project.go
majewsky: Ah sorry, I didn't realize you're doing a left outer join in the second query.

pkg/api/projects.go
auhlig: Thanks for spotting this!

pkg/api/quota_updater.go
auhlig: Done via 9482933606cfcd6bac56424241b3164c855cd1e7

pkg/api/projects.go
auhlig: Fixed via 9f80036a8fca9c9c3478ec3ac0b64a9b4fd3a8a0

auhlig: The `project_editor` is permitted to set rate limits now.

pkg/reports/cluster.go
auhlig: I'm using `clusters.Find(..)` [here](https://github.com/sapcc/limes/blob/ratelimits/pkg/reports/cluster.go#L424) and [here](https://github.com/sapcc/limes/blob/ratelimits/pkg/reports/cluster.go#L429). 

Just the explain the rationale for putting the doing the rate limit in `clusters.Find(..)`: 
I think it makes the loop clearer and is consistent to the pattern used in `projects.Find(..)`.
When looking at the code, the 1st snippet seems a lot more readable to me than the 2nd:
```
_, _, _, rateLimitReport := clusters.Find(config, clusterConfig.ID, &svcConfig.Type, nil, &rateLimit.TargetTypeURI, &action.Name)
if rateLimitReport != nil && &action.Limit != nil && &action.Unit != nil {
  rateLimitReport.Actions[action.Name].Limit = action.Limit
  rateLimitReport.Actions[action.Name].Unit = limes.Unit(action.Unit)
}
```

compared to having it all in the loop:
```
var rateLimit *limes.ClusterRateLimitReport
if targetTypeURI != nil && action != nil {
  rateLimit, exists = service.Rates[*targetTypeURI]
  if !exists {
    rateLimit = &limes.ClusterRateLimitReport{
      TargetTypeURI: *targetTypeURI,
      Actions:       make(limes.ClusterRateLimitActionReports),
    }
  }
  if _, exists := rateLimit.Actions[*action]; !exists {
    rateLimit.Actions[*action] = &limes.ClusterRateLimitActionReport{Name: *action}
  }
  service.Rates[*targetTypeURI] = rateLimit
}

if rateLimitReport != nil && &action.Limit != nil && &action.Unit != nil {
  rateLimitReport.Actions[action.Name].Limit = action.Limit
  rateLimitReport.Actions[action.Name].Unit = limes.Unit(action.Unit)
}
```

However, I don't have a strong opinion here. If you still want it changed, I'll do so.

auhlig:
... and another round @majewsky.

docs/example-policy.json
```diff
@@ -8,14 +8,15 @@
...
-  "project:lower":         "rule:project_editor",
-  "project:discover":      "rule:domain_editor",
+  "project:list":           "rule:domain_viewer",
+  "project:show":           "rule:project_viewer",
+  "project:edit":           "rule:project_editor",
+  "project:sync":           "rule:project_editor",
+  "project:raise":          "rule:domain_editor",
+  "project:raise_lowpriv":  "rule:project_editor",
+  "project:lower":          "rule:project_editor",
+  "project:set_rate_limit": "rule:project_editor",
```
majewsky: I know it's just an example, but that should be `rule:domain_editor` or even `rule:cluster_admin`. Project admins should not be able to set arbitrarily large rate limits without approval.

pkg/reports/cluster.go
majewsky: I cannot stop having a strong opinion, so I would appreciate if you put it all in the loop. :)

docs/example-policy.json
auhlig: Adjusted the example

PR #52, opened 2018-06-11
proposal for inconsistency detection API

majewsky:
@reimannf Please have a look, but do not merge yet. My intention is for Talal to use this branch for the implementation, too.


docs/users/api-v1-specification.md
```diff
@@ -343,6 +340,81 @@ all domains in all clusters (and will thus be the same for every cluster listed)
...
+      "quota": 5,
+      "backend_quota": 7
+    },
+    ...
+  ]
+}
+```
+
+Each entry in those three lists concerns exactly one resource in one project or domain. If multiple resources in the
+same project are inconsistent, they will appear as multiple entries. Like in the example above, the same project and
```
reimannf: Why are multiple resources, within the same project/domain represented as multiple entries? Why not follow the same structure as in `GET /v1/domains/:domain_id/projects/:project_id`

majewsky: The primary grouping here should be by inconsistency type here (IMO). Given that, I find it hard to add additional grouping by service/resource without making the structure quite hard to navigate. Also, I think the design principle of "making dissimilar things appear dissimilar" applies here, although I'm aware of the irony of it (since I'm always complaining about OpenStack being inconsistent with itself).

majewsky:
@MTalalAnwar Very nice! Since this is your first larger piece of Go code, I expected to write quite a few review comments regarding coding style, but you did very well in emulating the structure and style of the existing code base.

The only thing missing is some tests. Can you please add to `pkg/api/api_test.go`? You'll probably want to make a new test that does not start from `pkg/api/fixtures/start_data.sql`, but uses a different set of data that has inconsistencies in it.

PR #56, opened 2018-07-06
Marshalling support for ServiceQuotas

BugRoger:
Another addition to make reuse easier. This time marshalling the `ServiceQuota`.

### DIFF
--- a/pkg/api/input.go
+++ b/pkg/api/input.go
@@ -35,6 +35,40 @@ type ServiceQuotas map[string]ResourceQuotas
 //JSON request bodies in PUT requests.
 type ResourceQuotas map[string]limes.ValueWithUnit
 
+//MarshalJSON implements the json.Marshaler interface.
+func (sq ServiceQuotas) MarshalJSON() ([]byte, error) {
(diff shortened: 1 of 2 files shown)


pkg/api/input.go
```diff
@@ -35,6 +35,41 @@ type ServiceQuotas map[string]ResourceQuotas
 //JSON request bodies in PUT requests.
 type ResourceQuotas map[string]limes.ValueWithUnit
 
+//MarshalJSON implements the json.Marshaler interface.
+func (sq ServiceQuotas) MarshalJSON() ([]byte, error) {
+	type resourceQuota struct {
+		Name  string      `json:"name"`
+		Quota uint64      `json:"quota"`
+		Unit  *limes.Unit `json:"unit"`
```
majewsky: Why do you need a pointer here? I usually use a plain `limes.Unit` with the `omitempty` tag for JSON, since countable resources have `UnitNone = ""`.

BugRoger: Good point. I just copy/pasted the data structure from below. Removed it and the reassignment.

PR #57, opened 2018-07-23
proposal: CADF audit trail

MTalalAnwar:
&ast;*In Limes terminology: event = quota change*

Currently Limes logs all API requests to the standard output and standard error streams.

For PUT requests, Limes logs each individual quota change in the following format:

```bash
AUDIT: set quota compute.instances = 0 -> 50 for domain dd124178371a432499f7f03bddf11652 by user 53c978a396444c2c97499b70fdc3679c (D123456)
```

Hermes workflow: audit events from Keystone Middleware > oslo messaging bus held by RabbitMQ > Logstash transforms (adds domain and project metadata) events > ElasticSearch holds the events > Hermes API displays the events through HTTP requests or Elektra as front-end.

The proposal:

1. All Limes API `PUT` requests at the domain and project level get intercepted.
2. Transform these requests into a CADF formatted event.
3. Send these events to an Elasticsearch setup; for our purpose we send events to Hermes' Elasticsearch setup.
4. Audit trail can be viewed through Elektra with Hermes as the back-end.

Step 1 was already implemented in Limes. I have done a crude (not all the info is transformed to CADF) implementation of Step 2.

Current implementation of Step 2 looks like this:

```bash
2018/07/23 14:50:55 AUDIT: {"domain_id":"dd124178371a432499f7f03bddf11652","service":"compute","resource":"cores","newQuota":85}
2018/07/23 14:50:55 AUDIT: {"domain_id":"dd124178371a432499f7f03bddf11652","service":"compute","resource":"instances","newQuota":40}
2018/07/23 14:50:55 AUDIT: {"domain_id":"dd124178371a432499f7f03bddf11652","service":"object-store","resource":"capacity","newQuota":41943040000}
```

Additionally, I have also written an initial draft of the audit trail guide/spec available at `./docs/users/audit.md`. My intention is to create a guide that the reader can use to follow the audit trail and extract the required information from the audit event without having to go through the official (read: lengthy) CADF spec.


notque:
Looks good so far. First guess is entry point is a rabbitmq for limes picked up by our logstash keeping things consistent, and handling our rolling indexes properly.  

Getting the fields right which you've properly specified in the audit.md

But it all makes sense, great work sorting it out so far.

docs/users/audit.md
```diff
@@ -0,0 +1,91 @@
...
+    "host": {
+      "address": "0.0.0.0",
+      "agent": "curl/7.54.0"
+    },
+    "project_id": "example"
+  },
+  "target": {
+    "typeURI": "service/compute/ram/quota",
+    "id": "example-target-id"
+  },
```
majewsky: This should include the old and new quota values, to answer questions like "who approved that much quota?".

majewsky:
Looks good to me otherwise up to this point.

docs/users/audit.md
MTalalAnwar: You're right. I didn't add it yet because I modelled the example event structure after the CADF spec and Hermes' current events. Also, I wanted to be sure whether `target` was the correct place for this info.

majewsky:
Notes from our telco:
- also record failed change requests (as a next step)
- observer is the limes-api process, so `observer.typeURI = "service/resources"`

Open questions:
- typeURI grouping: `service/compute/ram/quota` or `service/quota/compute/ram`? (important because filtering on this field works only by prefix)
- Does the Limes chart deploy its own rabbitmq, or does Hermes deploy a general-purpose rabbitmq that Limes consumes?

pkg/api/domains.go
```diff
@@ -210,10 +212,37 @@ func (p *v1Provider) PutDomain(w http.ResponseWriter, r *http.Request) {
...
+					NewQuota: newQuota,
+				},
+				RequestPath: r.URL.String(),
+			}
+
+			//apply default values to the event
+			auditEvent.ApplyDefaults()
+
+			//add the event to the audit trail
+			auditTrail.Add(auditEvent)
```
majewsky: To decrease the chance of misusing the API of `auditTrail`, I think that AuditTrail.Add should call ApplyDefaults on the event. Then all of this can be simplified to a single
```go
auditTrail.Add(util.CADFevent {
  ...
})
```

pkg/api/domains.go
```diff
@@ -250,10 +279,36 @@ func (p *v1Provider) PutDomain(w http.ResponseWriter, r *http.Request) {
...
```
majewsky: Also, while I appreciate your desire to add comments to the code, please avoid comments that say the same thing as the code itself, like in this case. An API with speaking names often removes the need for comments.

The only time I deviate from this rule is for the header comments of paragraphs when applying the "code in paragraphs" rule. (See point no. 7 on [this page](https://www.perl.com/pub/2005/07/14/bestpractices.html/) for further details.)

pkg/util/rabbit.go
```diff
@@ -0,0 +1,71 @@
...
+
+	// declare a queue to hold and deliver messages to consumers.
+	q, err := ch.QueueDeclare(
+		"events", // name of the queue
+		false,    // durable: msgs should survive cluster reset
+		false,    // autodelete when unused
+		false,    // exclusive: gives error if other consumers subscribe to queue
+		false,    // no-wait: RabbitMQ should wait if a queue is already set up
+		nil,      // arguments for advanced config
+	)
```
majewsky: I haven't checked what these switches mean in detail. Are we sure that we want all of them to be false?

pkg/api/domains.go
MTalalAnwar: Done. The audit event creation and applying of defaults is now more clean and simplified.

pkg/audit/rabbit.go
```diff
@@ -0,0 +1,99 @@
...
+	duration := time.Second
+	for {
+		err := action()
+		if err != nil {
+			logg.Error("RabbitMQ -- %s", err)
+			if duration > 5*time.Minute {
+				duration = 5 * time.Minute
+			} else {
+				duration *= 2
+			}
```
majewsky: Once `duration` reaches 5 minutes, this concrete algorithm will cause it to oscillate between 5 and 10 minutes. The following variant does not exhibit this behavior:
```go
duration *= 2
if duration > 5 * time.Minute {
  duration = 5 * time.Minute
}
```

MTalalAnwar: updated and merged.

PR #59, opened 2018-08-07
My take on the structure of limes audit message

notque:
Muhammed, I've went ahead and went through the process of checking it, and updating it where it made sense to me. This includes moving the quota information into the attachments where it belongs. 

The naming changes were just preference as I did them, they aren't important, but the entire spec that we use for cadf is added here with omit empty.

Please look over it, thanks!


pkg/audit/log.go
```diff
@@ -45,101 +45,106 @@ var observerUUID string
...
-			NewQuota: newQuota,
-			Unit:     resUnit,
+		Target: Resource{
+			TypeURI:   fmt.Sprintf("service/%s/%s/quota", srvType, resName),
+			ID:        targetID,
+			ProjectID: targetID,
+			Attachments: Attachment{
+				Name:    "payload",
+				TypeURI: "mime:application/json",
+				Content: fmt.Sprintf("OldQuota:%d,NewQuota:%d,Unit:%s", resQuota, newQuota, resUnit),
```
majewsky: This doesn't look like `application/json` to me.

notque:
LGTM @MTalalAnwar

PR #60, opened 2018-08-13
Support audit events for rejected quota changes

talal:
Progress:

* Audit events for token authentication failure is a bit tricky, need to work on that.
* The rest of the error types for which we need audit events have been implemented.

### DIFF
--- a/pkg/api/domains.go
+++ b/pkg/api/domains.go
@@ -108,14 +108,9 @@ func (p *v1Provider) DiscoverDomains(w http.ResponseWriter, r *http.Request) {
 //PutDomain handles PUT /v1/domains/:domain_id.
 func (p *v1Provider) PutDomain(w http.ResponseWriter, r *http.Request) {
 	requestTime := time.Now().Format("2006-01-02T15:04:05.999999+00:00")
-	token := p.CheckToken(r)
-	canRaise := token.Check("domain:raise")
-	canLower := token.Check("domain:lower")
-	if !canRaise && !canLower {
-		token.Require(w, "domain:raise") //produce standard Unauthorized response
-		return
-	}
+	var auditTrail audit.Trail
 
+	token := p.CheckToken(r)
 	cluster := p.FindClusterFromRequest(w, r, token)
 	if cluster == nil {
 		return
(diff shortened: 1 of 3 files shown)


pkg/api/domains.go
```diff
@@ -137,6 +132,12 @@ func (p *v1Provider) PutDomain(w http.ResponseWriter, r *http.Request) {
 	}
 	serviceQuotas := parseTarget.Domain.Services
 
+	canRaise := token.Check("domain:raise")
+	canLower := token.Check("domain:lower")
+	if !canRaise && !canLower {
+		token.Require(w, "domain:raise") //produce standard Unauthorized response
+		return
+	}
```
majewsky: This check needs to happen before `Find{Cluster,Domain}FromRequest`. Otherwise, we may leak information about the existence of domains to unauthorized users. Also, I'm not sure why this was reordered.

pkg/api/projects.go
```diff
@@ -210,6 +205,13 @@ func (p *v1Provider) PutProject(w http.ResponseWriter, r *http.Request) {
 	}
 	serviceQuotas := parseTarget.Project.Services
 
+	canRaise := token.Check("project:raise")
+	canLower := token.Check("project:lower")
+	if !canRaise && !canLower {
+		token.Require(w, "project:raise") //produce standard Unauthorized response
+		return
+	}
+
```
majewsky: Same as above.

pkg/audit/log.go
```diff
@@ -104,77 +105,102 @@ type Host struct {
...
-//NewEvent takes the necessary parameters from an API request and returns a new audit event.
-func NewEvent(
-	t *gopherpolicy.Token, r *http.Request, requestTime, dbDomainID, dbProjectID,
-	srvType, resName string, resUnit limes.Unit, resQuota, newQuota uint64,
-) CADFEvent {
-	targetID := dbProjectID
-	if dbProjectID == "" {
-		targetID = dbDomainID
+//EventParams to-do
+type EventParams struct {
```
majewsky: "to-do"

talal:
Changed as requested.

majewsky:
Please merge when ready.

PR #66, opened 2018-10-22
update for changed CFM API

majewsky:
Checklist:

- [ ] ~I updated the documentation to describe the semantical or interface changes I introduced.~


reimannf:
We should increase the cache timeout here to at least 30 mins: https://github.com/sapcc/limes/blob/81be6d3fc0336ff3dc89d1e952aa61f32ef3e788/pkg/plugins/cfm.go#L89

majewsky:
> We should increase the cache timeout here to at least 30 mins
This was clarified in person. The cache only applies to the shareserver list. For individual shareservers, we don't need to cache since those are scraped in 30-minute intervals anyway.

PR #73, opened 2018-12-18
Burst audit trail

talal:


pkg/api/projects.go
```diff
@@ -360,8 +362,8 @@ func (p *v1Provider) putOrSimulatePutProjectQuotas(w http.ResponseWriter, r *htt
 }
 
 func (p *v1Provider) putOrSimulateProjectAttributes(w http.ResponseWriter, r *http.Request, simulate, hasBursting bool) {
-	//TODO generate audit events when changing Project.HasBursting
-
+	requestTime := time.Now().Format("2006-01-02T15:04:05.999999+00:00")
```
majewsky: The exact time format string is an implementation detail of the audit trail and should not be in here. `BurstEventParams` should take a `time.Time` instead of a string for `Time`.

pkg/audit/event.go
```diff
@@ -0,0 +1,285 @@
...
+	ServiceType  string
+	ResourceName string
+	OldQuota     uint64
+	NewQuota     uint64
+	QuotaUnit    limes.Unit
+	RejectReason string
+}
+
+//newEvent takes the necessary parameters and returns a new audit event.
+func (p ResourceEventParams) newEvent() CADFEvent {
```
majewsky: There is a lot of code duplication between the two `newEvent` methods. What about the following structure?

```go
type EventParams struct {
  //fields that are shared between BurstEventParams and ResourceEventParams
  Token *gopherpolicy.Token
  Request *http.Request
  ReasonCode int
  ...
  Target EventTarget
}

type EventTarget interface {
  RenderTarget() Resource
}

//EventTarget has two implementors, QuotaEventTarget and BurstEventTarget, e.g.
type BurstEventTarget struct {
  ProjectID string
  DomainID string
  NewStatus bool
}
```

So `QuotaEventTarget` and `BurstEventTarget` would each render the `target` resource only, and `EventParams` renders the rest when it gets converted into `CADFEvent`.

PR #76, opened 2019-01-14
add metrics for quota auto-approval

talal:
FYI @reimannf

### DIFF
--- a/pkg/api/quota_updater.go
+++ b/pkg/api/quota_updater.go
@@ -26,6 +26,7 @@ import (
 	"strings"
 	"time"
 
+	"github.com/prometheus/client_golang/prometheus"
 	"github.com/sapcc/go-bits/gopherpolicy"
 	"github.com/sapcc/go-bits/respondwith"
 	"github.com/sapcc/limes"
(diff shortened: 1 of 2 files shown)


talal:
metrics names changed as requested.

pkg/api/quota_updater.go
```diff
@@ -411,6 +412,42 @@ func (u QuotaUpdater) CommitAuditTrail(token *gopherpolicy.Token, r *http.Reques
 
 	for srvType, reqs := range u.Requests {
 		for resName, req := range reqs {
+			// auto-approval metrics
```
majewsky: Also here.

pkg/api/quota_updater.go
```diff
@@ -411,6 +412,42 @@ func (u QuotaUpdater) CommitAuditTrail(token *gopherpolicy.Token, r *http.Reques
...
+			if u.CanRaiseLP && !u.CanRaise {
+				if u.ScopeType() == "domain" {
+					domainLabels := prometheus.Labels{
+						"os_cluster": u.Cluster.ID,
+						"domain_id":  string(u.Domain.ID),
+						"service":    srvType,
+						"resource":   resName,
+					}
+					lowPrivilegeRaiseDomainSuccessCounter.With(domainLabels).Add(0)
+					lowPrivilegeRaiseDomainFailureCounter.With(domainLabels).Add(0)
```
majewsky: Let's do it without the `.Add(0)` initialization here. I don't think we need zero-initialized counters here, and it would help keep the number of timeseries way down.

pkg/api/quota_updater.go
```diff
@@ -411,6 +412,42 @@ func (u QuotaUpdater) CommitAuditTrail(token *gopherpolicy.Token, r *http.Reques
...
+					if invalid {
+						lowPrivilegeRaiseDomainFailureCounter.With(domainLabels).Inc()
+					} else {
+						lowPrivilegeRaiseDomainSuccessCounter.With(domainLabels).Inc()
+					}
+				} else {
+					projectLabels := prometheus.Labels{
+						"os_cluster": u.Cluster.ID,
+						"domain_id":  string(u.Project.DomainID),
+						"project_id": string(u.Project.ID),
```
majewsky: @reimannf Do we want to also have the domain/project names as labels like we have in the data metrics (`limes_project_quota` etc.)?

reimannf: we can skip the domain/project labels

PR #77, opened 2019-01-16
add metrics for object count per swift container

talal:
FYI @reimannf


talal:
tests were successful.

PR #81, opened 2019-02-07
support multiple constraints files at once

majewsky:
Rationale for this is explained in the accompanying PR in the secrets repo.


majewsky:
Come to think of it, I don't need a code change in Limes for what I want to do.

PR #85, opened 2019-04-04
availability zone support for subresources

talal:
Checklist:

FYI @reimannf

### DIFF
--- a/pkg/plugins/nova.go
+++ b/pkg/plugins/nova.go
@@ -27,6 +27,7 @@ import (
 
 	"github.com/gophercloud/gophercloud"
 	"github.com/gophercloud/gophercloud/openstack"
+	"github.com/gophercloud/gophercloud/openstack/compute/v2/extensions/availabilityzones"
 	"github.com/gophercloud/gophercloud/openstack/compute/v2/extensions/limits"
 	"github.com/gophercloud/gophercloud/openstack/compute/v2/extensions/quotasets"
 	"github.com/gophercloud/gophercloud/openstack/compute/v2/flavors"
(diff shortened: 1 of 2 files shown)


pkg/plugins/nova.go
```diff
@@ -246,6 +246,18 @@ func (p *novaPlugin) Scrape(provider *gophercloud.ProviderClient, eo gophercloud
...
 	}
 
+	//Server represents a server/instance in the OpenStack cloud.
+	//We use our own struct type instead of the Server type provided by
+	//gophercloud because we need some fields that it doesn't provide.
+	type Server struct {
+		ID               string                 `json:"id"`
+		Name             string                 `json:"name"`
+		Status           string                 `json:"status"`
+		Image            map[string]interface{} `json:"-"`
```
majewsky: If this is `json:"-"`, then how is image information extracted?

talal: copy-pasted from https://github.com/gophercloud/gophercloud/blob/master/openstack/compute/v2/servers/results.go#L184

pkg/plugins/nova.go
```diff
@@ -473,8 +476,11 @@ func (p *novaPlugin) findOSType(provider *gophercloud.ProviderClient, eo gopherc
 	return osType, nil
 }
 
-func makeIntPointer(value int) *int {
-	return &value
+//ServerWithAZ represents a server/instance in the OpenStack cloud with the
+//ServerAvailabilityZoneExt.
+type ServerWithAZ struct {
```
majewsky: This type should be private. In fact, I would lean towards inlining it altogether into line 263 since it's only used once:
```go
var instances []struct {
  servers.Server
  availabilityzones.ServerAvailabilityZoneExt
}
```

talal: done.

PR #88, opened 2019-04-18
nova: add ip address support for instances

talal:
Checklist:

Doc update will follow after code approval.

FYI @reimannf

### DIFF
--- a/pkg/plugins/nova.go
+++ b/pkg/plugins/nova.go
@@ -20,6 +20,7 @@
 package plugins
 
 import (
+	"encoding/json"
 	"fmt"
 	"regexp"
 	"sort"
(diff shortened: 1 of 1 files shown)


pkg/plugins/nova.go
```diff
@@ -270,11 +271,25 @@ func (p *novaPlugin) Scrape(provider *gophercloud.ProviderClient, eo gophercloud
...
+					data, err := json.Marshal(instance.Addresses)
+					if err != nil {
+						logg.Error("error while trying to parse ip address data for instance %q: %v", instance.ID, err)
+					} else {
+						err = json.Unmarshal(data, &ipAddresses)
+						if err != nil {
+							logg.Error("error while trying to parse ip address data for instance %q: %v", instance.ID, err)
+						}
+					}
+				}
```
majewsky: To avoid the duplicate `logg.Error()`, you can use the following pattern:

```go
data, err := json.Marshal(instance.Addresses)
if err == nil {
  err = json.Unmarshal(data, &ipAddresses)
}
if err != nil {
  logg.Error(...)
}
```

pkg/plugins/nova.go
```diff
@@ -498,3 +513,76 @@ func (opts novaQuotaUpdateOpts) ToComputeQuotaUpdateMap() (map[string]interface{
...
+type novaServerIPAddresses []novaServerIPData
+
+type novaServerIPData struct {
+	Address string `json:"address"`
+	Type    string `json:"type"`
+	Target  string `json:"target,omitempty"`
+}
+
+//UnmarshalJSON implements the json.Unmarshaler interface.
+func (nsip *novaServerIPAddresses) UnmarshalJSON(b []byte) error {
```
majewsky: This part is really messy because it still has all those runtime type checks in it. Please first unmarshal in a data structure representing the data returned from Nova, then translate that into the appropriate data structure for our own subresource.

docs/users/audit.md
```diff
@@ -93,7 +93,7 @@ The table below should help you understand what the different fields in an audit
 | `event.initiator.domain_id` | ID of domain where the initiating user was authorized. |
 | `event.initiator.project_id` | ID of project where the initiating user was authorized. |
 | `event.initiator.host.address` | Host's address from where the quota change was requested. |
-| `event.initiator.host.agent` | Curl or Elektron depending on where the request came from. |
+| `event.initiator.host.agent` | Curl or Elektra depending on where the request came from. |
```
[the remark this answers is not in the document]
talal: Ahh, I thought that perhaps I made a typo by mistaking it for the _other Electron_ at the time of the initial write-up.

pkg/plugins/nova.go
talal: done

pkg/plugins/nova.go
talal: This is already implemented. The incoming `b []byte` is [first unmarshalled](https://github.com/sapcc/limes/blob/656488c2711bbc5e0220001f3667a48ef2c0027a/pkg/plugins/nova.go#L525-L529) into a  `map[string]interface{}`, which is the data structure used by Nova's server type ([reference](https://github.com/gophercloud/gophercloud/blob/b9b92a82580638d091517faa0a8a9655657ed263/openstack/compute/v2/servers/results.go#L192)).

All these type checks are necessary because the `interface{}` has nested data structures which need to be checked individually.

majewsky: We're talking past each other, I guess. In line 276, you marshal from `map[string]interface{}` into `[]byte`. Then in line 526, you unmarshal back into `map[string]interface{}` (effectively canceling out the first operation), and then you gnaw at that opaque structure with dynamic type checks. Why don't you unmarshal into a more useful data structure, e.g.
```go
var addresses map[string][]struct {
  MAC     string `json:"OS-EXT-IPS-MAC:mac_addr"`
  Type    string `json:"OS-EXT-IPS:type"`
  Address string `json:"addr"`
}
```
Then this can be transformed into `type novaServerIPAddresses` without any dynamic type checking magic. Doesn't even need an extra `UnmarshalJSON` function.

PR #90, opened 2019-07-24
Switch to Hermes' libraries

talal:


pkg/audit/log.go
```diff
@@ -40,34 +44,100 @@ func init() {
...
+					c.trail.events = c.trail.events[i:]
+					eventPublishSuccessCounter.With(labels).Inc()
+				} else {
+					eventPublishFailedCounter.With(labels).Inc()
+					//we don't want retry.ExponentialBackoff here because it might keep
+					//the connection open for a long time. A small interval sleep and
+					//one retry will suffice.
+					time.Sleep(30 * time.Second)
+					err = rabbit.PublishEvent(ch, c.Config.RabbitMQ.QueueName, &e)
+					if err != nil {
```
majewsky: When `err == nil`, the event should be removed from the queue here.

pkg/audit/log.go
```diff
@@ -40,34 +44,100 @@ func init() {
...
+			}.RetryUntilSuccessful(func() error {
+				conn, ch, err = connectToRabbitMQ(c.Config.RabbitMQ.URL, c.Config.RabbitMQ.QueueName)
+				return err
+			})
+
+			c.trail.mux.Lock()
+			for i, e := range c.trail.events {
+				err = rabbit.PublishEvent(ch, c.Config.RabbitMQ.QueueName, &e)
+				if err == nil {
+					c.trail.events = c.trail.events[i:]
```
majewsky: I'm weary of how this interacts with the ongoing iteration over `c.trail.events` in line 119. I would suggest the following alternative strategy:

```go
for len(c.trail.events) > 0 {
  nextEvent := c.trail.events[0]
  //...try to send it...
  //if successful:
  c.trail.events = c.trail.events[1:]
}
```

pkg/audit/log.go
```diff
@@ -39,34 +41,92 @@ func init() {
...
-//Add adds an event to the audit trail.
-func (t *Trail) Add(p EventParams) {
-	event := p.newEvent()
-	t.events = append(t.events, event)
+//Start starts the audit trail by initializing the EventSinkPerCluster
+//and starting separate Commit() goroutines per Cluster.
+func Start(configPerCluster map[string]core.CADFConfiguration) {
+	EventSinkPerCluster = make(map[string]chan<- cadf.Event)
+	for clusterID, config := range configPerCluster {
+		eventSink := make(chan cadf.Event)
```
majewsky: Please add some buffering to this channel (by giving a [second argument to `make`](https://golang.org/pkg/builtin/#make)), to reduce the risk that the RabbitMQ sending blocks the HTTP API from responding to requests.

PR #94, opened 2019-09-09
Free capacity per availability zone

talal:
Checklist:

### DIFF
--- a/docs/users/api-v1-specification.md
+++ b/docs/users/api-v1-specification.md
@@ -528,7 +528,7 @@ allocating quota to domains.
 The `per_availability_zone` key will only be supplied when capacity data can be
 measured for each availability zone separately. This data reflects the actual
 underlying hardware capacity, meaning that there may be disparity between this
-and the cluster level information.
+and the cluster-level information.
 
 When `raw_capacity` is given, it means that this resource is configured with an overcommitment. The `capacity` key will
 show the overcommitted capacity (`raw_capacity` times overcommitment factor).
--- a/docs/users/api-v1-specification.md
+++ b/docs/users/api-v1-specification.md
@@ -444,16 +444,18 @@ Returns 200 (OK) on success. Result is a JSON document like:
             {
               "name": "cores",
               "capacity": 1000,
-              "per_availability_zone": {
-                "az-one": {
+              "per_availability_zone": [
+                {
+                  "name": "az-one",
                   "capacity": 500,
                   "usage": 0
                 },
-                "az-two": {
+                {
+                  "name": "az-two",
                   "capacity": 500,
                   "usage": 2
                 }
-              },
+              ],
               "domains_quota": 100,
               "usage": 2
             },
@@ -462,16 +464,18 @@ Returns 200 (OK) on success. Result is a JSON document like:
               "unit": "MiB",
               "capacity": 1048576,
               "raw_capacity": 524288,
-              "per_availability_zone": {
-                "az-one": {
+              "per_availability_zone": [
+                {
+                  "name": "az-one",
                   "capacity": 262144,
                   "usage": 2048
                 },
-                "az-two": {
+                {
+                  "name": "az-two",
                   "capacity": 262144,
                   "usage": 0
                 }
-              },
+              ],
               "domains_quota": 204800,
               "usage": 2048,
               "burst_usage": 128
--- a/docs/users/api-v1-specification.md
+++ b/docs/users/api-v1-specification.md
@@ -6,6 +6,34 @@ catalog under the service type `resources`.
 Where permission requirements are indicated, they refer to the default policy. Limes operators can configure their
 policy differently, so that certain requests may require other roles or token scopes.
 
+* [Request headers](#request-headers)
+  * [X\-Auth\-Token](#x-auth-token)
+  * [X\-Limes\-Cluster\-Id](#x-limes-cluster-id)
+* [GET /v1/domains/:domain\_id/projects](#get-v1domainsdomain_idprojects)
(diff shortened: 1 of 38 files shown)


docs/users/api-v1-specification.md
```diff
@@ -387,7 +415,7 @@ In contrast to project data, `scraped_at` is replaced by `min_scraped_at` and `m
 ## GET /v1/clusters/:cluster\_id
 ## GET /v1/clusters/current
 
-Query data for clusters. Requires a cloud-admin token. Arguments:
+Query data for clusters. `:cluster_id` is optional for cloud admins. Cloud admin token shows all clusters. With any other token, only that token's cluster is shown. Arguments:
```
majewsky: > With any other token, only that token's cluster is shown.
That should be "only that token's cluster *may be* shown", i.e. requests for other clusters will return 403.

docs/users/api-v1-specification.md
```diff
@@ -415,6 +443,16 @@ Returns 200 (OK) on success. Result is a JSON document like:
...
               "name": "cores",
               "capacity": 1000,
+              "per_availability_zone": {
+                "az-one": {
+                  "capacity": 500,
+                  "physical_usage": 0
+                },
+                "az-two": {
+                  "capacity": 500,
+                  "physical_usage": 2
```
majewsky: Should that not be just `usage`?

docs/users/api-v1-specification.md
```diff
@@ -472,6 +520,9 @@ administrators, in which case a `comment` string will be present (such as for `o
 output above). The capacity is only informational: Cloud admins can choose to exceed the reported capacity when
 allocating quota to domains.
 
+The `per_availability_zone` key will only be supplied when a capacity is known. It shows the `capacity` and `physical_usage`
+for a single resource per availability zone.
```
majewsky: > will only be supplied when a capacity is known
Should be "will only be supplied *when capacity can be measured for each AZ separately*".

docs/users/api-v1-specification.md
talal: Yes, changed.

pkg/api/clusters.go
```diff
@@ -63,15 +63,29 @@ func (p *v1Provider) ListClusters(w http.ResponseWriter, r *http.Request) {
...
+	showBasic := !token.Check("cluster:show")
 
 	clusterID := mux.Vars(r)["cluster_id"]
+	currentClusterID := p.Cluster.ID
 	if clusterID == "current" {
-		clusterID = p.Cluster.ID
+		clusterID = currentClusterID
+	}
+	if showBasic && (clusterID != currentClusterID) {
+		http.Error(w, "querying data for a cluster other than the current cluster requires a cloud-admin token", http.StatusUnauthorized)
```
majewsky: I would just say `Forbidden` in these cases, similar to how Limes responds in this situation right now.

Also, the status must be 403 (`http.StatusForbidden`). 401 is for failed authentication ("we don't know who that is"), and 403 is for failed authorization ("we know who that is, and he may not pass").

Same in line 84 below.

report_cluster.go
```diff
@@ -51,6 +51,7 @@ type ClusterResourceReport struct {
 	Capacity      *uint64    `json:"capacity,omitempty"`
 	RawCapacity   *uint64    `json:"raw_capacity,omitempty"`
 	Comment       string     `json:"comment,omitempty"`
+	CapacityPerAZ JSONString `json:"per_availability_zone,omitempty"`
```
majewsky: Please use a struct type with the fields as defined in the spec, e.g. by moving `type core.AvailabilityZoneCapacityData` into this module under an appropriate name.

We only use `JSONString` for subresources and subcapacities because those structures vary between plugins, but that is not the case here.

pkg/db/migrations.go
```diff
@@ -127,34 +127,40 @@ var SQLMigrations = map[string]string{
...
-	"009_add_project_resources_physical_usage.up.sql": `
+	"010_add_project_resources_physical_usage.up.sql": `
 		ALTER TABLE project_resources ADD COLUMN physical_usage BIGINT DEFAULT NULL;
 	`,
-	"010_add_project_rate_limits.down.sql": `
+	"011_add_project_rate_limits.down.sql": `
 		DROP TABLE project_rate_limits;
 	`,
-	"010_add_project_rate_limits.up.sql": `
+	"011_add_project_rate_limits.up.sql": `
```
majewsky: Existing migrations may not be touched. You need to add your migration at the bottom as number 11, otherwise this will break on existing databases (i.e. in production).

pkg/plugins/capacity_nova.go
```diff
@@ -175,30 +219,45 @@ func (p *capacityNovaPlugin) Scrape(provider *gophercloud.ProviderClient, eo gop
...
 
 	if maxFlavorSize != 0 {
-		instanceCapacity := uint64(math.Min(float64(10000*azCount), float64(totalLocalGb)/maxFlavorSize))
-		capacity["compute"]["instances"] = core.CapacityData{Capacity: instanceCapacity}
+		totalInstances := calculateInstanceAmount(azCount, totalLocalGb, maxFlavorSize)
+		instancesPerAZ := make(map[string]*core.AvailabilityZoneCapacityData)
+		for az, localGb := range localGbPerAZ {
+			instancesPerAZ[az] = &core.AvailabilityZoneCapacityData{
+				Capacity: calculateInstanceAmount(azCount, localGb.Capacity, maxFlavorSize),
+				Usage:    calculateInstanceAmount(azCount, localGb.Usage, maxFlavorSize),
```
majewsky: Since this is only for a single AZ, `azCount` should probably be replaced by 1, but we need to check how this affects the numbers that come out in the end.

PR #100, opened 2020-01-20
manila: use the new netapp exporter metric format

talal:
Checklist:


talal:
Prometheus queries tested in QA. All good.

PR #101, opened 2020-04-20
check lbaasv2 support

notandy:
This change introduces a pre-flight check for the neutron plugin that checks
if the lbaasv2 extension is enabled for setting lbaasv2 related quotas.

TLDR: Now works lbaasv2-less neutrons, since lbaasv2 is dropped with train.

Checklist:

### DIFF
--- a/pkg/plugins/neutron.go
+++ b/pkg/plugins/neutron.go
@@ -22,6 +22,7 @@ package plugins
 import (
 	"github.com/gophercloud/gophercloud"
 	"github.com/gophercloud/gophercloud/openstack"
+	"github.com/gophercloud/gophercloud/openstack/common/extensions"
 	"github.com/sapcc/limes"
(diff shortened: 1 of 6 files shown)


pkg/plugins/neutron.go
```diff
@@ -129,7 +131,24 @@ func init() {
...
+
+	// LBaaSv2 supported?
+	r := extensions.Get(client, "lbaasv2")
+	switch r.Result.Err.(type) {
+	case gophercloud.ErrDefault404:
+		p.hasLBaaS = false
+	case nil:
+		p.hasLBaaS = true
+	default:
+		return r.Result.Err
```
majewsky, line 147: A more descriptive error message would be nice.

```suggestion
		return fmt.Errorf("cannot check for lbaasv2 support in Neutron: %s", r.Result.Err.Error())
```

PR #104, opened 2020-09-10
Introduce cronus endpoint plugin

kayrus:
Checklist:


majewsky (changes requested):
As already hinted at on Slack, this cannot go in as-is because a quota plugin doesn't make sense without quota. We need to make some sort of adjustments in the Limes core for usage-only services, and also think on a fundamental level about how to model services where usage is a counter, not a gauge.

PR #105, opened 2020-10-20
API proposal for generalization of rate limits

majewsky:
@reimannf As always with large API changes, I'm open to feedback.

@auhlig Also from you, since this affects the [openstack-rate-limit-middleware](https://github.com/sapcc/openstack-rate-limit-middleware). There are several backwards-incompatible changes in here to make my future life easier, but from a cursory study of the middleware code, it shouldn't be too hard to adapt it to the proposed API.

1. Since we will in the future have several rates that are not for API requests and thus don't fit in the CADF framework, we don't replicate the target_type_uri/action structure on the Limes API level anymore. Instead, `rates` are a flat list within each service just like `resources`, and we encode the target_type_uri/action pair into the rate name where necessary.

2. The existing `unit` field for rate limits is split into `unit` and `window`. With this change, `unit` has the same meaning as for resources and also makes sense for usage values. On API request rates, the `unit` field is omitted, just like for countable resources.

3. For rates where we know how to scrape usage, `usage_as_bigint` is added to the rate and `rates_scraped_at` is added to the service.

4. In general, a bunch of wording surrounding rates is being clarified and codified in the spec.

### DIFF
--- a/docs/users/api-v1-specification.md
+++ b/docs/users/api-v1-specification.md
@@ -11,9 +11,10 @@ policy differently, so that certain requests may require other roles or token sc
   * [X\-Limes\-Cluster\-Id](#x-limes-cluster-id)
 * [GET /v1/domains/:domain\_id/projects](#get-v1domainsdomain_idprojects)
 * [GET /v1/domains/:domain\_id/projects/:project\_id](#get-v1domainsdomain_idprojectsproject_id)
+  * [Quota/usage for resources](#quota-usage-for-resources)
   * [Subresources](#subresources)
   * [Quota bursting details](#quota-bursting-details)
-  * [Rate limits](#rate-limits)
+  * [Rate limits and throughput tracking](#rate-limits-and-throughput-tracking)
     * [Default rate limits](#default-rate-limits)
 * [GET /v1/domains](#get-v1domains)
 * [GET /v1/domains/:domain\_id](#get-v1domainsdomain_id)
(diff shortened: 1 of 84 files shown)


docs/users/api-v1-specification.md
```diff
@@ -255,59 +250,71 @@ exists and is non-zero), then resources with `usage > quota` will display an add
...
+ms    - millisecond
+s     - second
+m     - minute
+h     - hour
+```
+
+If Limes knows how to track usage for a certain rate, the field `usage_as_bigint` will be shown (like in the large
+example above). The value of this field is an ever-increasing counter, guaranteed to never reset. Most JSON parser
+libraries parse integers into 64-bit-wide types, a size which can be reasonably expected to overflow esp. for rates
+relating to data throughput. Therefore the `usage_as_bigint` field (as indicated by the name) is set up like a bigint
```
reimannf, line 299: Would the `usage_as_bigint` inherit the `unit` from the rate or is this implicit?

reimannf (approved):
Only a little remaining question but lgtm. we should keep in mind that barbican already jumped on the rate-limit-middleware boat.

auhlig (approved):

docs/users/api-v1-specification.md
majewsky: Yes, `usage` is in terms of the `unit`. That's why I split the existing `unit` (e.g. `r/m`) into `unit` and `window`.

majewsky: Clarified in d99c541c42.

PR #110, opened 2021-04-01
manila: don't set replica quota for non-replication enabled types

Carthaca:
use implicit defaults from the manila service instead

Checklist:

I need help, I have no experience with go, the travis errors are real.
Also https://github.com/sapcc/limes/pull/110/files#diff-b0174593825027531cd9d5cf073b71f27d92f07d70c3b454901d10b889f49463R160-R163 can go now, our manila is on victoria release now globally.

### DIFF
--- a/pkg/plugins/manila.go
+++ b/pkg/plugins/manila.go
@@ -267,8 +267,8 @@ func (p *manilaPlugin) SetQuota(provider *gophercloud.ProviderClient, eo gopherc
 			Gigabytes:         quotas[p.makeResourceName("share_capacity", stName)],
 			Snapshots:         quotas[p.makeResourceName("share_snapshots", stName)],
 			SnapshotGigabytes: quotas[p.makeResourceName("snapshot_capacity", stName)],
-			Replicas:          0,
-			ReplicaGigabytes:  0,
+			Replicas:          nil,
+			ReplicaGigabytes:  nil,
 			ShareNetworks:     nil,
 		}
 		if p.hasReplicaQuotas && shareType.ReplicationEnabled {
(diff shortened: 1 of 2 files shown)


pkg/plugins/manila.go
```diff
@@ -316,8 +316,8 @@ type manilaQuotaSetDetail struct {
 	Shares            manilaQuotaDetail `json:"shares"`
 	SnapshotGigabytes manilaQuotaDetail `json:"snapshot_gigabytes"`
 	Snapshots         manilaQuotaDetail `json:"snapshots"`
-	ReplicaGigabytes  manilaQuotaDetail `json:"replica_gigabytes"`
-	Replicas          manilaQuotaDetail `json:"share_replicas"`
+	ReplicaGigabytes  manilaQuotaDetail `json:"replica_gigabytes,omitempty"`
+	Replicas          manilaQuotaDetail `json:"share_replicas,omitempty"`
```
majewsky, line 320: This data structure is only used for `Scrape`, not for `SetQuota`.

pkg/plugins/manila.go
```diff
@@ -267,8 +267,8 @@ func (p *manilaPlugin) SetQuota(provider *gophercloud.ProviderClient, eo gopherc
 			Gigabytes:         quotas[p.makeResourceName("share_capacity", stName)],
 			Snapshots:         quotas[p.makeResourceName("share_snapshots", stName)],
 			SnapshotGigabytes: quotas[p.makeResourceName("snapshot_capacity", stName)],
-			Replicas:          0,
-			ReplicaGigabytes:  0,
+			Replicas:          nil,
+			ReplicaGigabytes:  nil,
```
majewsky, line 271: I don't understand why that would make a difference. These fields are marked as `json:"-"` and thus don't get serialized. Also, this change produces compiler errors for me because `nil` is not a suitable value for `uint64`.

majewsky (changes requested):
👍 on the documentation fix, but for all I can tell, the code already has the intended behavior.

majewsky:
The problem was fixed independently in 73b8759b4ed7567044e35cc6bdb1a0a8efa21a93.

PR #111, opened 2021-05-26
Use quota_usage endpoint in Octavia API instead of counting.

velp:
This PR changes scraping for Octavia quota usage metrics to use new endpoint `quota_usage` instead of counting resources one by one.

Based on https://github.com/sapcc/octavia/pull/16

**⚠️DO NOT MERGE THIS PR BEFORE OCTAVIA'S ONE⚠️**

Checklist:


majewsky:
Looks good. I will apply this when the quota usage endpoint is rolled out to prod.

PR #113, opened 2021-06-17
Ironic: use `resource_class` field for matching nodes to flavors

talal:
Checklist:


majewsky (approved):
Awesome, thanks! Please go ahead with the validation in QA.

PR #120, opened 2021-12-16
Use environment variables for db, api, collector configuration options

talal:
Checklist:


pkg/core/config.go
```diff
@@ -24,30 +24,26 @@ import (
...
 )
 
 //Configuration contains all the data from the configuration file.
 type Configuration struct {
-	Database  db.Configuration       `yaml:"database"`
-	Clusters  map[string]*Cluster    `yaml:"-"`
-	API       APIConfiguration       `yaml:"api"`
-	Collector CollectorConfiguration `yaml:"collector"`
+	Clusters map[string]*Cluster
+	API      APIConfiguration
```
majewsky, line 42: Why do we leave the APIConfiguration in `type Configuration`? From what I can tell, we only need the fields of API configuration in `taskServe()`. The only thing that gets passed down is `config.API.PolicyEnforcer`, but we can also have `taskServe()` give the Enforcer to the API directly.

One of my goals here is to have `type Configuration` contain only the `map[string]*Cluster`, so that it eventually gets obsoleted entirely when we remove multi-cluster support next year and there is only one `*Cluster` left in the end`. And I really think we can get there.

pkg/core/config.go
```diff
@@ -347,14 +335,35 @@ func NewConfiguration(path string) (cfg Configuration) {
...
+	}
+	allowedOrigins := strings.ReplaceAll(os.Getenv("LIMES_API_CORS_ALLOWED_ORIGINS"), " ", "")
+	apiCfg.CORS.AllowedOrigins = strings.Split(allowedOrigins, "||")
+	exceptCodeStrings := strings.Split(os.Getenv("LIMES_API_REQUEST_LOG_EXCEPT_STATUS_CODES"), ",")
+	exceptCodes := make([]int, 0, len(exceptCodeStrings))
+	for _, v := range exceptCodeStrings {
+		v := strings.TrimSpace(v)
+		code, err := strconv.Atoi(v)
+		if err != nil {
+			logg.Fatal(err.Error())
```
majewsky, line 355: The error message should indicate where the parse error occurred:
```suggestion
			logg.Fatal("while parsing LIMES_API_REQUEST_LOG_EXCEPT_STATUS_CODES: %s", err.Error())
```

majewsky (changes requested):

majewsky (approved):

PR #121, opened 2022-02-08
plugins discovery: convert to gophercloud

SuperSandro2000:
Checklist:

  - I compared the called API endpoints in mitmproxy and the repsonse from the domains endpoint and they where all identically


pkg/plugins/capacity_cinder.go
```diff
@@ -166,8 +144,8 @@ func (p *capacityCinderPlugin) Scrape(provider *gophercloud.ProviderClient, eo g
 		}
 
 		azCapaData := capaData[resourceName].CapacityPerAZ[poolAZ]
-		azCapaData.Capacity += uint64(element.Capabilities.TotalCapacityGb)
-		azCapaData.Usage += uint64(element.Capabilities.AllocatedCapacityGb)
+		azCapaData.Capacity += uint64(element.Capabilities.TotalCapacityGB)
+		azCapaData.Usage += uint64(element.Capabilities.ProvisionedCapacityGB)
```
majewsky, line 148: This changes from evaluating `.pools[].capabilities.allocated_capacity_gb` to reading `.pools[].capabilities.provisioned_capacity_gb`, but this field is not provided by our Cinder installation. (At least according to my spot check in eu-de-1.)

The issue with some of these admin-level APIs is that different OpenStack installations may return different shapes of data based on what kind of driver they're using (e.g. for the case of Cinder, a VMware backend may return differently-shaped data than a Ceph backend).

Some of this variation is covered by the official API description. If the [OpenStack API ref for Cinder v2](https://docs.openstack.org/api-ref/block-storage/v2/index.html) mentions the `allocated_capacity_gb` field, the next step would be to get those fields covered in Gophercloud via PR over there. Otherwise, we will have to continue using our custom extraction data type here.

majewsky (changes requested):
Please split out the Cinder stuff into its own PR. This part needs some more work, the other parts are good to go in.

majewsky:
FYI, I have merged all commits except for the Cinder one.

PR #124, opened 2022-02-13
remove multi-cluster support

majewsky:
**Warning:** Do not merge without my approval. We need to remove the one remaining legacy cluster first.


docs/operators/config.md
```diff
@@ -625,35 +573,12 @@ scraping. Subresources bear the following attributes:
 | `status` | string | volume status [as reported by OpenStack Cinder](https://developer.openstack.org/api-ref/block-storage/v2/index.html#volumes-volumes) |
 | `size` | integer value with unit | volume size |
 
-# Available capacity plugins
+## Available capacity plugins
 
 Note that capacity for a resource only becomes visible when the corresponding service is enabled in the
 `clusters.$id.services` list as well.
```
SuperSandro2000, line 579: I think you missed something here.

majewsky: Fixed in ff15a82603.

SuperSandro2000 (approved):

pkg/api/core.go
```diff
@@ -111,7 +107,17 @@ func NewV1Router(cluster *core.Cluster, config core.Configuration, policyEnforce
 	r.Methods("POST").Path("/v1/domains/{domain_id}/projects/{project_id}/simulate-put").HandlerFunc(p.SimulatePutProject)
 	r.Methods("PUT").Path("/v1/domains/{domain_id}/projects/{project_id}").HandlerFunc(p.PutProject)
 
-	return sre.Instrument(r), p.VersionData
+	return sre.Instrument(forbidClusterIDHeader(r)), p.VersionData
+}
+
+func forbidClusterIDHeader(inner http.Handler) http.Handler {
```
reimannf, line 113: Is this enough for `limesctl` ?

majewsky: Good point. In fact, limesctl defaults to `X-Limes-Cluster-Id: current` when nothing else is specified. I had not anticipated this since this setting is rather pointless. Fixed in e3329caefe.

PR #125, opened 2022-02-14
Plugins cinder: convert to gophercloud

SuperSandro2000:
Checklist:

Sadly Gophercloud does not provide a map string for quota sets https://github.com/gophercloud/gophercloud/blob/master/openstack/blockstorage/extensions/quotasets/results.go#L163


majewsky:
> Sadly Gophercloud does not provide a map string for quota sets

`quotasets.Update()` takes an interface type though, so if you implement `quotasets.UpdateOptsBuilder` for a custom payload type, you can still use it.

majewsky (approved):

PR #133, opened 2022-03-03
Remove support for `X-Limes-Cluster-Id` header

talal:
Checklist:


majewsky:
I see where you're coming from, but I purposefully left this branch in because the cost of having one extra line of code and one testcase is justified to me when it means I don't have to hunt down users that send `X-Limes-Cluster-Id: current` and get them to change their code.

PR #134, opened 2022-03-04
Add support for dependent quota with OPA

SuperSandro2000:
Checklist:


limes.rego
```diff
@@ -0,0 +1,39 @@
+package limes
+
+# No NoneZeroDependendedQuota is set
+output = {"allow": true, "message": null} {
+	input.Behavior.NoneZeroDependedQuotaServiceType == ""
+	input.Behavior.NoneZeroDependedQuotaResourceName == ""
+}
+
+# NoneZeroDependendedQuota condition is met
+output = {"allow": true, "message": null} {
```
majewsky, line 10: As discussed on the phone, a more useful input/output setup for the policy would be to generate a DomainReport or ProjectReport representing the desired state of the domain or project after the requested quota has been applied (i.e. deep-copy the existing report and adjust `Quota` and `DesiredBackendQuota` to match the requested values).

With this change in input format, the output format should also change to report as many error messages at once as possible, to match the error reporting style of QuotaUpdater (where we support reporting one error message per requested quota). OPA Gatekeeper has a similar setup, wherein the policy returns a list of violations, and an empty list is interpreted as acceptance. With this pattern, the policy from the testcase could look like this:

```rego
package limes

violation[{ "msg": msg, "service": "shared", "resource": "things" }] {
  input.TargetProjectReport.Services["shared"].Resources["capacity"].Quota == 0
  msg := "must allocate shared/things quota before"
}
```

pkg/core/config.go
```diff
@@ -47,6 +48,8 @@ type ClusterConfiguration struct {
 	Subcapacities        map[string][]string              `yaml:"subcapacities"`
 	Authoritative        bool                             `yaml:"authoritative"`
 	ConstraintConfigPath string                           `yaml:"constraints"`
+	OpenPolicyAgentPath  string                           `yaml:"open_agent_policy"`
+	PreparedRegoQuery    rego.PreparedEvalQuery           `yaml:"-"`
```
majewsky, line 52: I'm not anticipating much overlap in the checks that OPA policies will be used for on project level and domain level, respectively. Therefore please split domain-level and project-level checks into two separate policies:
```suggestion
	OPA  OPAConfiguration                           `yaml:"opa"`
```

where
```go
type OPAConfiguration struct {
  ProjectQuotaPolicyPath string `yaml:"project_quota_policy_path"`
  ProjectQuotaQuery rego.PreparedEvalQuery `yaml:"-"`
  DomainQuotaPolicyPath string `yaml:"domain_quota_policy_path"`
  DomainQuotaQuery rego.PreparedEvalQuery `yaml:"-"`
}
```

majewsky (commented):
Some early remarks. I'm explicitly focusing on the large scale here.

pkg/api/quota_updater.go
```diff
@@ -452,6 +513,62 @@ func (u QuotaUpdater) validateAuthorization(srv limes.ServiceInfo, oldQuota, new
...
+		for _, resultViolationEntry := range resultViolations {
+			if resultViolation, ok := resultViolationEntry.(map[string]interface{}); ok {
+				if _, ok := resultViolation["msg"].(string); ok {
+					violations = append(violations, map[string]string{
+						"msg":      resultViolation["msg"].(string),
+						"service":  resultViolation["service"].(string),
+						"resource": resultViolation["resource"].(string),
+					})
+				} else {
+					logg.Error("%#v", resultViolation["msg"])
```
majewsky, line 548: The error message should describe what is happening, not least to make it easier to find the respective log in Kibana when a user reports the user-visible error message to us.

pkg/api/quota_updater.go
```diff
@@ -452,6 +513,62 @@ func (u QuotaUpdater) validateAuthorization(srv limes.ServiceInfo, oldQuota, new
...
+					return nil, &core.QuotaValidationError{
+						Status:  http.StatusInternalServerError,
+						Message: "OPA violation msg has wrong format (please report this problem!)",
+					}
+				}
+			} else {
+				logg.Error("OPA violation has wrong format: %s", reflect.TypeOf(resultViolation))
```
majewsky, line 555: You can simplify this with the `%T` format directive:
```suggestion
				logg.Error("OPA violation has wrong format: %T", resultViolation)
```

pkg/api/quota_updater.go
```diff
@@ -452,6 +513,62 @@ func (u QuotaUpdater) validateAuthorization(srv limes.ServiceInfo, oldQuota, new
...
+				return nil, &core.QuotaValidationError{
+					Status:  http.StatusInternalServerError,
+					Message: "OPA violation has wrong format (please report this problem!)",
+				}
+			}
+		}
+		return violations, nil
+	} else {
+		logg.Error("OPA returned unsupported data: %#s", results[0].Bindings)
```
majewsky, line 564: `#` is not supported for `%s` as far as I know. Did you mean `%#v`?

majewsky (changes requested):

pkg/api/api_test.go
```diff
@@ -109,6 +109,8 @@ func setupTest(t *testing.T, clusterName, startData string) (*core.Cluster, http
 	clusterConfig := core.ClusterConfiguration{
 		Auth: &core.AuthParameters{},
 	}
+	// dummy contains no violations to not be restricted to nonsense violations in all other tests
+	clusterConfig.SetupOPA("fixtures/dummy.rego")
```
majewsky, line 113: Why do we need a dummy file here? We should support having no policy at all for backwards-compatibility and simplicity.

pkg/core/config.go
SuperSandro2000: It's just easier to have everything in one file and spares us a lot of code duplication.

```rego
package limes

violation[{ "msg": msg, "service": "shared", "resource": "things" }] {
  input.targetdomainreport....
}

violation[{ "msg": msg, "service": "shared", "resource": "things" }] {
  input.targetprojectreport.....
}
```

majewsky: I'm willing to accept a bit of code duplication for increased clarity. When someone looks at the policy in our config two years down the line, I don't want there to be any unnecessary confusion between what is a domain-level policy and what is a project-level policy. Both look so similar (both in structure and in names like `targetdomainreport` vs. `targetprojectreport`) that they will be very likely confused.

pkg/api/api_test.go
SuperSandro2000: Now I am skipping the entire OPA section when the file is missing. Otherwise we get a nil point dereference [here](https://github.com/sapcc/limes/blob/888f039803e0f3975051109b2a868f0ce01b9222/pkg/api/quota_updater.go#L523) and if the rego file does not contain some boilerplate the we don't get any output and other errors trigger.

majewsky (approved):
I've added a small change to make the `opa` config section optional (otherwise the first deployment after merge will fail), and I've also added some basic documentation. With this, this is good to go in from my PoV.

PR #135, opened 2022-03-07
rewrite reports.GetProjects() into a streaming style to limit peak memory usage

majewsky:
As explained in the commit messages and code comments.

Checklist:

- [ ] ~If this PR is about a plugin, I tested the plugin against an OpenStack cluster.~
- [ ] ~I updated the documentation to describe the semantical or interface changes I introduced.~

### DIFF
--- a/pkg/api/projects.go
+++ b/pkg/api/projects.go
@@ -20,7 +20,9 @@
 package api
 
 import (
+	"bufio"
 	"database/sql"
+	"encoding/json"
 	"fmt"
 	"net/http"
 	"strings"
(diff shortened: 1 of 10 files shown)


pkg/api/projects.go
```diff
@@ -59,12 +61,74 @@ func (p *v1Provider) ListProjects(w http.ResponseWriter, r *http.Request) {
...
+	//
+	//The basic output structure looks like this:
+	//
+	//	{"projects":[ <REPORT> , <...> , <REPORT> ]}
+	//
+	//We delay the opening `{"projects":[` until we receive the first report, so
+	//that errors can be logged as a 5xx response if necessary. Upon getting the
+	//first report, we commit to the response being 200 and print reports as they
+	//come in. If we get to the end, we just need to write the trailing `]}` to
+	//complete the response.
```
SuperSandro2000, line 83: and what if an error happens in the middle?

pkg/api/projects.go
```diff
@@ -59,12 +61,74 @@ func (p *v1Provider) ListProjects(w http.ResponseWriter, r *http.Request) {
...
+			//the callback function above was never called
+			respondwith.JSON(w, http.StatusOK, map[string]interface{}{"projects": []*limes.ProjectReport{}})
+			return
+		}
+	}
+	if err != nil {
+		if outputStarted {
+			//we are already writing a 200 response, so we can only log the error,
+			//but we cannot send it to the client
+			logg.Error("late error during GET %s: %s", r.URL.String(), err.Error())
```
SuperSandro2000, line 126: Maybe we could append some invalid json at the end with the error message? Otherwise people might be confused why something aborted and start debugging in the wrong direction.

SuperSandro2000 (commented):
Do we have a test which returns such a big report?

majewsky:
> Do we have a test which returns such a big report?

The test cases only return up to 2 reports at once, but that exercises all the code paths of the current implementation.

pkg/api/projects.go
majewsky: Then the JSON will be incomplete since we don't write the trailing `]}`.

SuperSandro2000:
> The test cases only return up to 2 reports at once, but that exercises all the code paths of the current implementation.

I don't feel super confident that this is enough even if we test all code paths. With two reports not a lot can go wrong especially with consistent ordering but we also can't simulate a real chaotic production environment. I really want to avoid that we add some rare race condition which duplicates or eats entries.

Could we generate some semi-random dummy data to test this with 100 or 1000 reports to make sure that it works under lab conditions?

majewsky:
> Could we generate some semi-random dummy data to test this with 100 or 1000 reports to make sure that it works under lab conditions?

Test added in 5041408134.

SuperSandro2000 (approved):
Lets find the other bugs in staging and production 🍿

PR #136, opened 2022-03-08
Switch to golangci-lint

talal:
Checklist:


pkg/plugins/capacity_prometheus.go
```diff
@@ -101,23 +102,23 @@ func prometheusGetSingleValue(client prom_v1.API, queryStr string, defaultValue
...
 	}
 	resultVector, ok := value.(model.Vector)
 	if !ok {
-		//lint:ignore ST1005 Prometheus is a proper name
+		//nolint:stylecheck //Prometheus is a proper name
 		return 0, fmt.Errorf("Prometheus query failed: %s: unexpected type %T", queryStr, value)
 	}
 
+	//nolint:stylecheck //Since we use default case in the middle of the switch and
+	//because Prometheus is a proper name.
```
SuperSandro2000, line 115: So far the only bad thing about golangci. Either you go with an entire check group or you disable it completely.

PR #137, opened 2022-03-15
Add API endpoint for scrape errors

talal:
Checklist:

### DIFF
--- a/pkg/api/fixtures/scrape-error-list.json
+++ b/pkg/api/fixtures/scrape-error-list.json
@@ -1,0 +1,30 @@
+{
+  "scrape_errors": [
+    {
+      "project": {
+        "id": "uuid-for-dummy-project",
+        "name": "dummy-project",
+        "domain": {
+          "id": "uuid-for-dummy-domain",
+          "name": "dummy-domain"
+        }
+      },
+      "service_type": "shared",
+      "checked_at": 44,
+      "message": "could not scrape shared service"
+    },
+    {
+      "project": {
+        "id": "uuid-for-berlin",
+        "name": "berlin",
+        "domain": {
+          "id": "uuid-for-germany",
+          "name": "germany"
+        }
+      },
+      "service_type": "unshared",
+      "checked_at": 11,
+      "message": "could not scrape this specific unshared service"
+    }
+  ]
+}
--- a/pkg/api/fixtures/scrape-error-list.json
+++ b/pkg/api/fixtures/scrape-error-list.json
@@ -2,15 +2,16 @@
   "scrape_errors": [
     {
       "project": {
-        "id": "uuid-for-dummy-project",
-        "name": "dummy-project",
+        "id": "uuid-for-paris",
+        "name": "paris",
         "domain": {
-          "id": "uuid-for-dummy-domain",
-          "name": "dummy-domain"
+          "id": "uuid-for-france",
+          "name": "france"
         }
       },
+      "affected_projects": 3,
       "service_type": "shared",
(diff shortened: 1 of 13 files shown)


talal:
I've put the PR as draft as I still need to add rates scrape error reporting.

pkg/reports/error.go
```diff
@@ -0,0 +1,140 @@
...
+	"github.com/sapcc/limes/pkg/db"
+)
+
+var scrapeErrorsQuery = db.SimplifyWhitespaceInSQL(`
+	SELECT d.uuid, d.name, p.uuid, p.name, ps.type, ps.checked_at, ps.scrape_error_message
+	  FROM projects p
+	  LEFT OUTER JOIN domains d ON d.id = p.domain_id
+	  LEFT OUTER JOIN project_services ps ON ps.project_id = p.id
+	WHERE %s GROUP BY d.uuid, d.name, p.uuid, p.name, ps.type, ps.checked_at, ps.scrape_error_message
+	HAVING ps.scrape_error_message != ''
```
majewsky, line 39: - `LEFT OUTER JOIN` can be replaced by `JOIN` since the `ps.scrape_error_message != ''` condition makes it impossible to have partially-null result rows.
- `GROUP BY` is not necessary since no aggregate functions are used and the query will never produce multiple rows for one project service (with basically the same justification as in the previous point).
- 
```suggestion
	  JOIN domains d ON d.id = p.domain_id
	  JOIN project_services ps ON ps.project_id = p.id
	WHERE %s AND ps.scrape_error_message != ''
```

pkg/reports/error.go
```diff
@@ -0,0 +1,140 @@
...
+
+func getScrapeErrors(cluster *core.Cluster, dbi db.Interface, filter Filter, dbQuery string) ([]ScrapeError, error) {
+	fields := map[string]interface{}{"d.cluster_id": cluster.ID}
+
+	var result []ScrapeError
+	queryStr, joinArgs := filter.PrepareQuery(dbQuery)
+	whereStr, whereArgs := db.BuildSimpleWhereClause(fields, len(joinArgs))
+	err := db.ForeachRow(dbi, fmt.Sprintf(queryStr, whereStr), append(joinArgs, whereArgs...), func(rows *sql.Rows) error {
+		sErr := ScrapeError{}
+		var checkedAtAsTime *time.Time
```
majewsky, line 75: When initializing to a zero value, it's more idiomatic to just declare the variable in that type.
```suggestion
		var (
			sErr ScrapeError
			checkedAtAsTime *time.Time
		)
```

pkg/api/fixtures/scrape-error-list.json
```diff
@@ -0,0 +1,30 @@
...
+  "scrape_errors": [
+    {
+      "project": {
+        "id": "uuid-for-dummy-project",
+        "name": "dummy-project",
+        "domain": {
+          "id": "uuid-for-dummy-domain",
+          "name": "dummy-domain"
+        }
+      },
```
majewsky, line 11: I like the idea of grouping multiple similar errors together, but not having any actual project IDs in here would make it harder to investigate in cases where multiple projects are affected, but not all of them are. Please show the identity of one of the affected projects, and add a field like `affected_projects` containing the number of affected projects if that number is not 1. Then we can display in the UI something like `Project: myproject@mydomain (+10 more)`.

majewsky (changes requested):

majewsky (approved):
I fixed an error that CodeQL found. With that, this is good to go in.

PR #163, opened 2022-04-28
provide a stub implementation for a separate rate data API

majewsky:
During #135, I realized that having two types of differently-shaped data coming from different data sources (resource data and rate data) in the same API leads to an unnecessarily complex implementation and a lot of pain in general. Since the only user for the rate data is CBR (who collect Cronus rate usage data for billing purposes), we still have an opportunity to remake this API into something more maintainable.

As a first step, this PR provides new endpoints below `/rates/v1/` that implement reading of rate data via the existing report logic, but with `?rates=only` implied. This is not exactly what we want to have here in the end (there are still lots of `"resources": []` in the JSON outputs of these new endpoints), but it's enough to migrate CBR over to these new endpoints for their usecase. This will then enable a second step where we remove rate data support from the original endpoints and restructure the implementation accordingly.

### DIFF
--- a/docs/users/api-v1-specification.md
+++ b/docs/users/api-v1-specification.md
@@ -6,32 +6,6 @@ catalog under the service type `resources`.
 Where permission requirements are indicated, they refer to the default policy. Limes operators can configure their
 policy differently, so that certain requests may require other roles or token scopes.
 
-* [Request headers](#request-headers)
-  * [X\-Auth\-Token](#x-auth-token)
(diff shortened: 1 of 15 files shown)


majewsky:
Did you review the rest of the PR? I would like to move forward here since the bulk of the time will be spent waiting on our API consumers to use the new API.

SuperSandro2000:
I looked at it and didn't notice any gleaming errors.

SuperSandro2000 (approved):

PR #229, opened 2022-08-03
Move some settings into environment variables

SuperSandro2000:
Checklist:

TODO: update https://github.com/sapcc/limes/blob/master/docs/operators/config.md


docs/operators/config.md
```diff
@@ -41,21 +41,30 @@ Configuration options relating to the behavior of the collector service.
...
+| Variable | Required | Description |
+| --- | --- | --- |
+| `OS_AUTH_URL` | yes | URL for Keystone v3 API in this cluster. Should end in `/v3`. Other Keystone API versions are not supported. |
+| `OS_USERNAME` | yes | Limes service user. |
+| `OS_USER_DOMAIN_NAME` | yes | Domain containing Limes service user. |
+| `OS_PROJECT_NAME` | yes | Project where Limes service user has access. |
+| `OS_PROJECT_DOMAIN_NAME` | yes | Domain containing that project. |
+| `OS_PASSWORD` | yes | Password for Limes service user. |
+| `OS_REGION_NAME` | no | In multi-region OpenStack clusters, this selects the region to work on. |
+| `OS_INTERFACE` | no | The endpoint type Limes should use for the OpenStack services. |
```
majewsky, line 58: There are lots of other variations of configuration that are supported by gophercloud/utils/clientconfig. For other services, I usually point to the upstream documentation for openstackclient, see e.g. https://github.com/sapcc/tenso/blob/183c721f334e86cab344afb5c3018acc1c28226d/README.md?plain=1#L63

docs/operators/config.md
```diff
@@ -41,21 +41,30 @@ Configuration options relating to the behavior of the collector service.
...
+| `LIMES_AUTHORITATIVE` | no | If set to `true`, the collector will write the quota from its own database into the backend service whenever scraping encounters a backend quota that differs from the expectation. This flag is strongly recommended in production systems to avoid divergence of Limes quotas from backend quotas, but should be used with care during development. |
+| `LIMES_CONSTRAINTS_PATH` | no | Path to a YAML file containing the quota constraints for this cluster. See [*quota constraints*](constraints.md) for details. |
+| `LIMES_OPA_DOMAIN_QUOTA_POLICY_PATH` | yes, when `opa` section is given | Like `opa.project_quota_policy_path`, but this policy is used for validating domain quota changes instead. |
```
majewsky, line 61:
```suggestion
| `LIMES_OPA_DOMAIN_QUOTA_POLICY_PATH` | yes, when `opa` section is given | Like `LIMES_OPA_PROJECT_QUOTA_POLICY_PATH`, but this policy is used for validating domain quota changes instead. |
```

docs/operators/config.md
```diff
@@ -41,21 +41,30 @@ Configuration options relating to the behavior of the collector service.
...
+| `LIMES_OPA_PROJECT_QUOTA_POLICY_PATH` | yes, when `opa` section is given | Path to a Rego file. This policy will be evaluated whenever a user tries to change a project quota. If the policy reports any violations, the PUT will fail and the generated error messages will be reported to the user. See [../example-quota-policy.rego](../example-quota-policy.rego) for an example illustrating the expected structure of the policy. |
+
+The configuration file options relating to the targeted OpenStack cluster. A minimal config file could look like this:
```
majewsky, line 64: While we're at it, we should fix the broken grammar in this line.
```suggestion
Furthermore, a YAML file must be provided that describes things like the set of available backend services and the quota/capacity scraping behavior. A minimal config file could look like this:
```

pkg/api/audit.go
```diff
@@ -63,30 +61,18 @@ func StartAuditTrail(clusterID string, config core.CADFConfiguration) {
...
-		}
-		if config.RabbitMQ.Port == 0 {
-			config.RabbitMQ.Port = 5672
-		}
 		rabbitURI := url.URL{
 			Scheme: "amqp",
-			Host:   net.JoinHostPort(config.RabbitMQ.Hostname, strconv.Itoa(config.RabbitMQ.Port)),
-			User:   url.UserPassword(config.RabbitMQ.Username, string(config.RabbitMQ.Password)),
+			Host:   net.JoinHostPort(osext.GetenvOrDefault("LIMES_AUDIT_RABBITMQ_HOSTNAME", "localhost"), osext.GetenvOrDefault("LIMES_AUDIT_RABBITMQ_PORT", "5672")),
+			User:   url.UserPassword(osext.GetenvOrDefault("LIMES_AUDIT_RABBITMQ_USERNAME", "rabbitmq"), os.Getenv("LIMES_AUDIT_RABBITMQ_PASSWORD")),
```
majewsky, line 67: This does not match the documented behavior:
```suggestion
			User:   url.UserPassword(osext.GetenvOrDefault("LIMES_AUDIT_RABBITMQ_USERNAME", "guest"), osext.GetenvOrDefault("LIMES_AUDIT_RABBITMQ_PASSWORD", "guest")),
```

pkg/api/audit.go
```diff
@@ -63,30 +61,18 @@ func StartAuditTrail(clusterID string, config core.CADFConfiguration) {
...
 			Path:   "/",
 		}
 
 		go audittools.AuditTrail{
 			EventSink:           s,
 			OnSuccessfulPublish: onSuccessFunc,
 			OnFailedPublish:     onFailFunc,
-		}.Commit(rabbitURI, config.RabbitMQ.QueueName)
+		}.Commit(rabbitURI, osext.GetenvOrDefault("LIMES_AUDIT_QUEUE_NAME", "notifications.info"))
```
majewsky, line 75: This does not match the documented behavior:
```suggestion
		}.Commit(rabbitURI, osext.MustGetenv("LIMES_AUDIT_QUEUE_NAME"))
```

pkg/core/auth.go
```diff
@@ -54,33 +47,22 @@ func (auth *AuthParameters) Connect() error {
...
-		return fmt.Errorf("cannot initialize OpenStack client: %w", err)
+		logg.Fatal("cannot find OpenStack credentials: " + err.Error())
 	}
 
-	userAgent := fmt.Sprintf("%s@%s", bininfo.Component(), bininfo.VersionOr("rolling"))
-	auth.ProviderClient.UserAgent.Prepend(userAgent)
-
-	err = openstack.Authenticate(auth.ProviderClient, gophercloud.AuthOptions{
-		IdentityEndpoint: auth.AuthURL,
-		AllowReauth:      true,
```
majewsky, line 68: The new codepath does not set AllowReauth, which is an error for a long-running service. Compare e.g. https://github.com/sapcc/tenso/blob/183c721f334e86cab344afb5c3018acc1c28226d/internal/tenso/config.go#L69

pkg/core/cluster.go
```diff
@@ -189,9 +190,9 @@ func (c *Cluster) Connect() error {
 	}
 
 	//load quota constraints
-	if c.Config.ConstraintConfigPath != "" && c.QuotaConstraints == nil {
+	if os.Getenv("LIMES_CONSTRAINTS_PATH") != "" && c.QuotaConstraints == nil {
 		var errs []error
-		c.QuotaConstraints, errs = NewQuotaConstraints(c, c.Config.ConstraintConfigPath)
+		c.QuotaConstraints, errs = NewQuotaConstraints(c, os.Getenv("LIMES_CONSTRAINTS_PATH"))
```
majewsky, line 195: A double invocation of `os.Getenv()` is not good style. The value should be put into a variable first.

pkg/core/config.go
```diff
@@ -29,37 +29,35 @@ import (
...
 type OPAConfiguration struct {
-	ProjectQuotaPolicyPath string                 `yaml:"project_quota_policy_path"`
-	ProjectQuotaQuery      rego.PreparedEvalQuery `yaml:"-"`
-	DomainQuotaPolicyPath  string                 `yaml:"domain_quota_policy_path"`
-	DomainQuotaQuery       rego.PreparedEvalQuery `yaml:"-"`
+	ProjectQuotaPolicyPath string
+	ProjectQuotaQuery      rego.PreparedEvalQuery
+	DomainQuotaPolicyPath  string
+	DomainQuotaQuery       rego.PreparedEvalQuery
 }
```
majewsky, line 42: Since the paths are not part of the bigger YAML data structure anymore, we don't need to store them here anymore. This then enables a simplification of how this is set up at runtime vs. at test time. Please change `func SetupOPA` to take both paths as arguments, and then provide them directly at that function's callsites.

pkg/core/config.go
```diff
@@ -484,9 +447,12 @@ func (cluster ClusterConfiguration) validateConfig() (success bool) {
 		success = false
 	}
 
-	if cluster.CADF.Enabled && cluster.CADF.RabbitMQ.QueueName == "" {
+	if osext.GetenvBool("LIMES_AUDIT_ENABLE") && osext.GetenvBool("LIMES_AUDIT_QUEUE_NAME") {
 		missing("cadf.rabbitmq.queue_name")
 	}
```
majewsky, line 452: This is not required anymore when MustGetenv is used for `LIMES_AUDIT_QUEUE_NAME` like I suggested above.

majewsky (changes requested):

majewsky:
@SuperSandro2000 I made some changes on top of your changes to improve the structure of the code and docs. Also, I think I fixed a bug that the tests did not catch (see commit messages). Please counter-review.

majewsky (approved):
If you agree with my changes, this can be merged together with the Helm chart change and the secrets change.

docs/operators/config.md
```diff
@@ -1,61 +1,62 @@
...
 | `LIMES_API_POLICY_PATH` | `/etc/limes/policy.yaml` | Path to the oslo.policy file that describes authorization behavior for this service. Please refer to the [OpenStack documentation on policies][policy] for syntax reference. This repository includes an [example policy][ex-pol] that can be used for development setups, or as a basis for writing your own policy. For `:raise`, `:raise_lowpriv`, `:lower` and `:set_rate_limit` policies, the object attribute `%(service_type)s` is available to restrict editing to certain service types. |
+| `LIMES_OPA_DOMAIN_QUOTA_POLICY_PATH` | yes, when `opa` section is given | Like `LIMES_OPA_PROJECT_QUOTA_POLICY_PATH`, but this policy is used for validating domain quota changes instead. |
+| `LIMES_OPA_PROJECT_QUOTA_POLICY_PATH` | yes, when `opa` section is given | Path to a Rego file. This policy will be evaluated whenever a user tries to change a project quota. If the policy reports any violations, the PUT will fail and the generated error messages will be reported to the user. See [../example-quota-policy.rego](../example-quota-policy.rego) for an example illustrating the expected structure of the policy. |
```
SuperSandro2000, line 30:
```suggestion
| `LIMES_OPA_DOMAIN_QUOTA_POLICY_PATH` | yes, when `opa` section is given | Path to a Rego file. This policy will be evaluated whenever a user tries to change a domain quota. If the policy reports any violations, the PUT will fail and the generated error messages will be reported to the user. See [../example-quota-policy.rego](../example-quota-policy.rego) for an example illustrating the expected structure of the policy. |
| `LIMES_OPA_PROJECT_QUOTA_POLICY_PATH` | yes, when `opa` section is given | Like `LIMES_OPA_PROJECT_QUOTA_POLICY_PATH`, but this policy is used for validating project quota changes instead. |
```

majewsky: Done.

PR #275, opened 2022-11-15
move plugin-specific config types into the DiscoveryPlugins themselves

majewsky:
On its surface, this is just cleaner separation of concerns. The pkg/core module should not have to know any types that are only relevant to the internals of individual plugin implementations.

The bigger picture is that I want to get test coverage for the config loading and cluster assembling code, by also using it to set up test scenarios. This means constructing test plugins in the usual way, from configuration. Without this change, we would have to bleed test-only types into pkg/core, which I find excessively ugly.

This change is a proof-of-concept for the approach. If this proves to work, I will also adjust the QuotaPlugin and CapacityPlugin interfaces in the same way.

### DIFF
--- a/pkg/core/config.go
+++ b/pkg/core/config.go
@@ -31,6 +31,8 @@ import (
 	limesresources "github.com/sapcc/go-api-declarations/limes/resources"
 	"github.com/sapcc/go-bits/logg"
 	yaml "gopkg.in/yaml.v2"
+
+	"github.com/sapcc/limes/pkg/util"
 )
 
 // ClusterConfiguration contains all the configuration data for a single cluster.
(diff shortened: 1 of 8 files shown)


pkg/core/config.go
```diff
@@ -30,6 +30,7 @@ import (
 	limesrates "github.com/sapcc/go-api-declarations/limes/rates"
 	limesresources "github.com/sapcc/go-api-declarations/limes/resources"
 	"github.com/sapcc/go-bits/logg"
+	"github.com/sapcc/limes/pkg/util"
 	yaml "gopkg.in/yaml.v2"
```
SuperSandro2000, line 34:
```suggestion
	yaml "gopkg.in/yaml.v2"

	"github.com/sapcc/limes/pkg/util"
```

SuperSandro2000 (approved):
Please keep a fire extinguisher near in case of things suddenly combust into flames.

PR #301, opened 2023-02-14
sharev2: add "snapmirror_capacity" resource (usage reporting only)

majewsky:
@Carthaca Can you please check whether I understood the requirement correctly? Especially the part where I'm assuming that shares managed by Manila are _never_ tagged as `volume_type="dp"` and that Snapmirror backups are _always_ tagged as `volume_type="dp"`. Also, is `snapmirror_capacity` ok as a user-facing resource name? I'm willing to take your suggestions if you have a better idea.

### DIFF
--- a/internal/plugins/manila.go
+++ b/internal/plugins/manila.go
@@ -511,7 +511,7 @@ func (p *manilaPlugin) collectSnapmirrorUsage(project core.KeystoneProject, shar
 	//Manila itself, so we have to collect usage from NetApp metrics instead
 	defaultValue := float64(0)
 	queryStr := fmt.Sprintf(
-		`sum(max by (volume) (netapp_volume_total_bytes{project_id=%q,share_type=%q,volume_type="dp",volume_state="online",volume=~".*EC2BKP"}))`,
+		`sum(max by (share_id) (netapp_volume_total_bytes{project_id=%q,volume_type!="dp",share_type=%q,volume_state="online",snapshot_policy="EC2_Backups"}))`,
 		project.UUID, shareTypeName,
 	)
 	bytesTotal, err := client.GetSingleValue(queryStr, &defaultValue)
@@ -520,7 +520,7 @@ func (p *manilaPlugin) collectSnapmirrorUsage(project core.KeystoneProject, shar
 	}
 
 	queryStr = fmt.Sprintf(
-		`sum(max by (volume) (netapp_volume_used_bytes{project_id=%q,share_type=%q,volume_type="dp",volume_state="online",volume=~".*EC2BKP"}))`,
+		`sum(max by (share_id) (netapp_volume_used_bytes{project_id=%q,volume_type!="dp",share_type=%q,volume_state="online",snapshot_policy="EC2_Backups"}))`,
 		project.UUID, shareTypeName,
 	)
 	bytesUsed, err := client.GetSingleValue(queryStr, &defaultValue)
--- a/internal/plugins/manila.go
+++ b/internal/plugins/manila.go
@@ -478,7 +478,7 @@ func (p *manilaPlugin) collectPhysicalUsage(project core.KeystoneProject) (manil
 		//NOTE: The `max by (share_id)` is necessary for when a share is being
 		//migrated to another shareserver and thus appears in the metrics twice.
 		queryStr := fmt.Sprintf(
-			`sum(max by (share_id) (netapp_volume_used_bytes{project_id=%q,volume_type!="dp",share_type=%q}))`,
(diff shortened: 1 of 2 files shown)


internal/plugins/manila.go
```diff
@@ -463,25 +478,64 @@ func (p *manilaPlugin) collectPhysicalUsage(project core.KeystoneProject) (manil
...
+	client, err := p.PrometheusAPIConfig.Connect()
+	if err != nil {
+		return core.ResourceData{}, err
+	}
+
+	//snapmirror backups are only known to the underlying NetApp filer, not to
+	//Manila itself, so we have to collect usage from NetApp metrics instead
+	defaultValue := float64(0)
+	queryStr := fmt.Sprintf(
+		`sum(max by (volume) (netapp_volume_total_bytes{project_id="%s",share_type="%s",volume_type="dp",volume=~".*EC2BKP"}))`,
```
Carthaca, line 514: I noticed some downsides with this approach looking at the snapmirror targets:

- we can only use this data if source and target region are the same
- due to snapmirror automatically adjusting the size of the target, it is not transparent to the end users what sizes can be expected:
  -  e.g. have a look yourself at `(max(netapp_volume_total_bytes{volume_type="dp"}) by (share_id))/(max (netapp_volume_total_bytes{volume_type="rw"}) by (share_id))`. This is not always equal 1. Depending on the physical usage the targets may by be bigger or smaller.

I suggest to collect 'Usage' by looking at the snapmirror source, identifying the relevant volumes by `snapshot_policy="EC2_Backups"`, i.e.
`sum(max by (volume) (netapp_volume_total_bytes{project_id=%q,volume_type!="dp",share_type=%q,snapshot_policy="EC2_Backups"}))`

Btw: is there a reason you used `"%s"` here instead of `%q` like above? (TIL: I even looked it up what this formatting means in golang)

internal/plugins/manila.go
```diff
@@ -463,25 +478,64 @@ func (p *manilaPlugin) collectPhysicalUsage(project core.KeystoneProject) (manil
...
+		project.UUID, shareTypeName,
+	)
+	bytesTotal, err := client.GetSingleValue(queryStr, &defaultValue)
+	if err != nil {
+		return core.ResourceData{}, err
+	}
+
+	queryStr = fmt.Sprintf(
+		`sum(max by (volume) (netapp_volume_used_bytes{project_id="%s",share_type="%s",volume_type="dp",volume=~".*EC2BKP"}))`,
```
Carthaca, line 523: For the sake of consistency the same applies to 'PhysicalUsage':
- we expect same size for this anyhow subject to the things that have not been synced at the time (incidentally I just found some broken snapmirror relationships by comparing these sizes)
- this way we also have this value available in regions where the target region of the snapmirror is different to the source

i.e.
``sum(max by (volume) (netapp_volume_used_bytes{project_id=%q,volume_type!="dp",share_type=%q,snapshot_policy="EC2_Backups"}))``

Carthaca (changes requested):
Thanks for starting this.

I'm fine with the name `snapmirror` (mirroring volumes by using snapshots) - we already exposed this term with castellum.
The term `backup` may be used later when we have a real backup implementation.
And the other term I could think of `data protection` (that's what the volume type `dp` stands for) is not really more descriptive imho.

I commented the changes to collect data at the source inline. I think the benefit in being able to collect this quota if source and target region are different outweighs the heuristic in making assumptions about the target by only looking at the source. Regional quotas/pricing is not accurate, but this approach here is anyhow only meant to bill something at all and hopefully not a solution that gets more users.

And since we are touching this place I have an unrelated improvement suggestion:
please add `volume_state="online"` (or `volume_state!="offline"`) to all `netapp_volume_*` metrics. When moving volumes in/between filers we would see the same volume twice and `max by ...` alone may wrongly choose the offline volume.

internal/plugins/manila.go
majewsky: I have updated the PR as per your suggestions.

Regarding the part about things breaking when source and target region are not the same, I'm surprised that this is an issue. I'm aware that we discussed this when I was last involved in this topic (which, admittedly, was quite a while back). But my recollection was that labelling would be added to the mirrors to attribute them to a project in the target region, thus rendering them accessible to billing. Is that not the case anymore?

Carthaca: We had that idea, yes - but it would mean too much effort:

-  how to select the target projects
-  where to maintain a mapping of source and target projects
-  explain this to the customer

The initial reaction to this effort was: 'oh let's not bill those cross-region setups, then'

Carthaca (approved):

Carthaca:
What we now do, breaks down to: select all shares where the metrics are labeled with `snapshot_policy="EC2_Backups"`, and report the usage data of those additionally as `snapmirror_capacity`.

PR #312, opened 2023-03-17
remove OPA support

majewsky:
As discussed. The "require Swift quota when Keppel is used" check moves into the IsQuotaAcceptableForProject machinery, which I extend to allow cross-service quota consistency checks in the same way that the OPA policy used to allow.

Checklist:

- [ ] If this PR is about a plugin, I tested the plugin against an OpenStack cluster. --> Will do so on Monday, currently on the train without QA access.


internal/plugins/keppel.go
```diff
@@ -98,7 +99,10 @@ func (p *keppelPlugin) Scrape(project core.KeystoneProject) (result map[string]c
 
 // IsQuotaAcceptableForProject implements the core.QuotaPlugin interface.
 func (p *keppelPlugin) IsQuotaAcceptableForProject(project core.KeystoneProject, fullQuotas map[string]map[string]uint64) error {
-	//not required for this plugin
+	ourQuotas := fullQuotas[p.ServiceInfo().Type]
+	if fullQuotas["object-store"]["capacity"] == 0 && ourQuotas["images"] > 0 {
+		return fmt.Errorf("Keppel can only be used when nonzero Swift quota is configured")
```
SuperSandro2000, line 105:
```suggestion
		return fmt.Errorf("Keppel can only be used when a non-zero Swift quota is configured")
```

majewsky: Merriam-Webster says "nonzero": https://www.merriam-webster.com/dictionary/nonzero

Wiktionary says that "non-zero" is allowed as an alternate spelling: https://en.wiktionary.org/wiki/non-zero

SuperSandro2000 (approved):

PR #322, opened 2023-04-21
parametrize QuotaPlugin interface by service type

majewsky:
The QuotaPlugin interface used to be built on the assumption that `pluginTypeID == serviceType` for each plugin. This assumption is broken in the tests, where the same generic plugin gets instantiated multiple times to test different subvariants in the same test.

The plugin loading code is already aware of the distinction between `pluginTypeID` and `serviceType`, but while trying to convert the testcases to load plugins from config, I ran into some places where the plugin implementations themselves rely on `pluginTypeID == serviceType`. This commit removes this assumption, by passing in `serviceType` or `allServiceInfos` from outside where required.

### DIFF
--- a/internal/core/plugin.go
+++ b/internal/core/plugin.go
@@ -113,7 +113,12 @@ type QuotaPlugin interface {
 	Init(client *gophercloud.ProviderClient, eo gophercloud.EndpointOpts, scrapeSubresources map[string]bool) error
 
 	//ServiceInfo returns metadata for this service.
-	ServiceInfo() limes.ServiceInfo
+	//
+	//This receives the `serviceType` as an argument because it needs to appear
+	//in the ServiceInfo struct. But in general, a plugin cannot know which
+	//serviceType it was instantiated for (esp. in unit tests, where the generic
+	//test plugin is instantiated multiple times for different service types).
+	ServiceInfo(serviceType string) limes.ServiceInfo
 
 	//Resources returns metadata for all the resources that this plugin scrapes
 	//from the backend service.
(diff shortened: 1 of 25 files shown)


internal/core/plugin.go
```diff
@@ -133,7 +138,12 @@ type QuotaPlugin interface {
 	//to the quotas of other services. (For example, Keppel uses Swift as a
 	//storage backend, so nonzero Keppel quota requires nonzero Swift quota in
 	//Limes, to catch a common misconfiguration that confuses users.)
-	IsQuotaAcceptableForProject(project KeystoneProject, fullQuotas map[string]map[string]uint64) error
+	//
```
SuperSandro2000, line 141:
```suggestion

```

internal/plugins/keppel.go
```diff
@@ -98,9 +98,17 @@ func (p *keppelPlugin) Scrape(project core.KeystoneProject) (result map[string]c
...
-	ourQuotas := fullQuotas[p.ServiceInfo().Type]
-	if fullQuotas["object-store"]["capacity"] == 0 && ourQuotas["images"] > 0 {
+func (p *keppelPlugin) IsQuotaAcceptableForProject(project core.KeystoneProject, fullQuotas map[string]map[string]uint64, allServiceInfos []limes.ServiceInfo) error {
+	var ourQuotas, swiftQuotas map[string]uint64
+	for _, srv := range allServiceInfos {
+		switch srv.ProductName {
+		case "keppel":
+			ourQuotas = fullQuotas[srv.Type]
+		case "swift":
+			swiftQuotas = fullQuotas[srv.Type]
```
SuperSandro2000, line 108: Would we want a default case here? I am not sure.

internal/core/plugin.go
majewsky: This is not a leading newline. It separates two paragraphs.

internal/plugins/keppel.go
majewsky: We are only interested in these specitic services. Quotas of other services are not relevant to this method, so they get skipped over.

PR #332, opened 2023-06-02
capacity-manila: consider "hardware_state" for capacity counting

majewsky:
@chuan137 @Carthaca I saw from your meeting minutes that this is appearing in prod now, so I'm getting ready to start using it in Limes. Please let me know when you're done with the rollout.

Checklist:


Carthaca:
You are fast 👍 . Yes, I wanted to let you know, once this has been deployed everywhere (most likely Monday).

This is removing the entire pool, which means we don't just drop the capacity, but also the usage, right?
I think it would be good to keep the usage, though. That was agreed with Frank when we brainstormed about that feature. But I'm not sure what is the current agreement: I saw the issue with `usable < provisioned` coming up in ironic, that in theory could happen here then, as well ...

But in the end, just make CBR happy, please and have a look at consistency across the services :)

Carthaca (approved):
lgtm

`hardware_state` has been deployed to golden regions on Wednesday

PR #355, opened 2023-09-12
rewrite capacity scrape into a jobloop.ProducerConsumerJob

majewsky:
This allows each capacitor to run individually, and fail individually, without blocking the other capacitors from running.


internal/collector/capacity_scrape.go
```diff
@@ -0,0 +1,307 @@
...
+		return err
+	}
+	return tx.Commit()
+}
+
+func convertAZReport(capacityPerAZ map[string]*core.CapacityDataForAZ) limesresources.ClusterAvailabilityZoneReports {
+	//The initial implementation wrote limesresources.ClusterAvailabilityZoneReports into
+	//the CapacityPerAZJSON database field, even though
+	//map[string]*core.CapacityDataForAZ would have been more appropriate. Now we
+	//stick with it for compatibility's sake.
```
SuperSandro2000, line 298:
```suggestion
	//The initial implementation wrote limesresources.ClusterAvailabilityZoneReports into the CapacityPerAZJSON database field,
	// even though map[string]*core.CapacityDataForAZ would have been more appropriate.
	//Now we stick with it for compatibility's sake.
```

internal/collector/capacity_scrape.go
```diff
@@ -0,0 +1,307 @@
...
+			serviceIDForType[serviceType] = serviceID
+		}
+		return err
+	})
+	if err != nil {
+		return fmt.Errorf("cannot collect cluster service mapping: %w", err)
+	}
+
+	//collect ownership information for existing cluster_resources
+	query = `SELECT cs.type, cr.name, cr.capacitor_id FROM cluster_resources cr JOIN cluster_services cs ON cr.service_id = cs.id`
```
SuperSandro2000, line 159: We should move the sql queries like most others outside of the function and don't reuse the variable.

internal/collector/capacity_scrape_test.go
```diff
@@ -120,13 +131,20 @@ func Test_ScanCapacity(t *testing.T) {
...
-		UPDATE cluster_capacitors SET scraped_at = 5, next_scrape_at = 905 WHERE capacitor_id = 'unittest2';
+		UPDATE cluster_capacitors SET scraped_at = 7, next_scrape_at = 907 WHERE capacitor_id = 'unittest2';
 		UPDATE cluster_resources SET capacity = 23 WHERE service_id = 1 AND name = 'things';
 	`)
 
+	//move the clock forward by 300 seconds (the capacitor add step only triggers every five minutes)
+	//TODO: I hate this clock
+	for step := 1; step <= 300; step++ {
+		_ = test.TimeNow()
+	}
```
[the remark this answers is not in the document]
majewsky: This is not an error result, it's a time result that's discarded because I only want the clock stepping behavior.

internal/collector/capacity_scrape_test.go
```diff
@@ -138,24 +156,32 @@ func Test_ScanCapacity(t *testing.T) {
...
+		UPDATE cluster_capacitors SET scraped_at = 315, next_scrape_at = 1215 WHERE capacitor_id = 'unittest';
+		UPDATE cluster_capacitors SET scraped_at = 317, next_scrape_at = 1217 WHERE capacitor_id = 'unittest2';
+		UPDATE cluster_capacitors SET scraped_at = 319, serialized_metrics = '{"smaller_half":3,"larger_half":7}', next_scrape_at = 1219 WHERE capacitor_id = 'unittest4';
 		UPDATE cluster_resources SET capacity = 10, subcapacities = '[{"smaller_half":3},{"larger_half":7}]' WHERE service_id = 2 AND name = 'things';
 	`)
 
+	//move the clock forward by 300 seconds (the capacitor add step only triggers every five minutes)
+	//TODO: I hate this clock
+	for step := 1; step <= 300; step++ {
+		_ = test.TimeNow()
```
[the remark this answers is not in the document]
majewsky: Same as above.

internal/collector/capacity_scrape.go
majewsky: done

PR #358, opened 2023-09-15
add full AZ awareness to plugin scraping interfaces

majewsky:
Depending on the resource, scraping can now return capacity and usage data either for the entire region, or broken down by AZs. This distinction is represented by the enum type `core.Topological`. Quota stays affixed to the regional level for now (i.e. one number that applies across all AZs), since there is no committed requirement to support AZ quota in the backend.

The frontend does not change in any way at this point. This change is specifically about enhancing the interface between Limes internals and plugins to the point that whatever changes end up being done in the frontend can be supported.

Some more details are in the commit messages.

**Update:** I have tested that the new plugin implementations yield the same results. There are some exceptions for `sharev2/share_capacity` and `compute/instances`, where, because of rounding errors, computing the total capacity from raw values (as we did before) is different from computing the capacity per AZ from raw values and then summing over all AZs. These are literal rounding errors, though.

### DIFF
--- a/internal/core/data.go
+++ b/internal/core/data.go
@@ -68,11 +68,43 @@ func (t Topological[D]) Sum() D {
 type TopologicalData[Self any] interface {
 	// List of permitted types. This is required for type inference, as explained here:
 	// <https://stackoverflow.com/a/73851453>
-	CapacityData
+	UsageData | CapacityData
 
 	// Computes the sum of this structure and `other`.
 	// This is used to implement Topological.Sum().
 	add(other Self) Self
+}
+
+// ResourceData contains quota and usage data for a single project resource.
+type ResourceData struct {
+	Quota     int64 //negative values indicate infinite quota
+	UsageData Topological[UsageData]
+}
+
+// UsageData contains usage data for a single project resource.
+// It appears in type ResourceData.
+type UsageData struct {
+	Usage         uint64
+	PhysicalUsage *uint64 //only supported by some plugins
+	Subresources  []any   //only if supported by plugin and enabled in config
+}
+
+// add implements the TopologicalData interface.
+//
+//nolint:unused // looks like a linter bug
+func (d UsageData) add(other UsageData) UsageData {
+	result := UsageData{
+		Usage:        d.Usage + other.Usage,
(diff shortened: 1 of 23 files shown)


internal/core/data.go
```diff
@@ -0,0 +1,95 @@
...
+}
+
+// PerAZ is a shorthand to construct a Topological instance with the PerAZ member filled.
+func PerAZ[D TopologicalData[D]](data map[string]*D) Topological[D] {
+	return Topological[D]{PerAZ: data}
+}
+
+// Sum returns a sum of all data in this container. If the Regional field is
+// filled, that data is returned directly. Otherwise, all entries in the PerAZ
+// field are summed together.
```
SuperSandro2000, line 47:
```suggestion
// Sum returns a sum of all data in this container.
// If the Regional field is filled, that data is returned directly.
// Otherwise, all entries in the PerAZ field are summed together.
```

internal/core/data.go
```diff
@@ -0,0 +1,95 @@
...
+	// <https://stackoverflow.com/a/73851453>
+	CapacityData
+
+	// Computes the sum of this structure and `other`.
+	// This is used to implement Topological.Sum().
+	add(other Self) Self
+}
+
+// CapacityData contains the total and per-availability-zone capacity data for a
+// single resource.
```
SuperSandro2000, line 111:
```suggestion
// CapacityData contains the total and per-availability-zone capacity data for a single resource.
```

internal/plugins/capacity_nova.go
```diff
@@ -252,9 +250,7 @@ func (p *capacityNovaPlugin) Scrape() (result map[string]map[string]core.Capacit
 		}
 		hvCapacity.MatchingAggregates = map[string]bool{matchingAggregateName: true}
 
-		//count this hypervisor's capacity towards the totals for the whole cloud...
-		totalCapacity.Add(hvCapacity)
-		//...and the AZ level
+		//count this hypervisor's capacity towards the totals for  the AZ level
```
SuperSandro2000, line 253:
```suggestion
		//count this hypervisor's capacity towards the totals for the AZ level
```

PR #368, opened 2023-10-16
add minimum viable API for commitments

majewsky:
@VoigtS Please take a look. This should be enough API to allow your work to progress.

~~This PR is currently in draft status because the test suite is missing.~~ Test suite is done as far as I can tell.


majewsky:
I reviewed this with @VoigtS in person and got verbal approval.

PR #369, opened 2023-10-17
add ID columns to *_resources tables

majewsky:
For various computations esp. relating to commitment confirmation, it will be necessary to model AZ-aware capacity, quota and usage data in the database. We currently have `project_services` below `projects` and `project_resources` below `project_services`. I want to attach a new table `project_az_resources` below `project_resources` in the same way (same for cluster level and potentially also domain level).

To avoid duplicating the current `(service_id, name)` key from the resources table into the new table, this PR adds a numeric `id` column as primary key to the resources tables, same as for most other tables.

- [x] verify that the migrations work correctly (tested on a local database using a backup of qa-de-1 as base)
- [x] API tests pass
- [x] collector tests pass

### DIFF
--- a/internal/api/fixtures/start-data-commitments.sql
+++ b/internal/api/fixtures/start-data-commitments.sql
@@ -32,25 +32,25 @@ INSERT INTO project_services (id, project_id, type, scraped_at, checked_at) VALU
 
 -- project_resources contains only boring placeholder values
 -- berlin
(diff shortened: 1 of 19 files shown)


VoigtS:
Reading the PR comment after the commits confirms my thought that service_id was a compsite key before, so I removed that comment frome mine. Looks like a consequent change with a migration on top.

PR #372, opened 2023-10-23
finish conversion to AZ-aware DB records

majewsky:
As a followup to #370, this changeset moves all uses of AZ-aware data from the sum fields on the resources to the AZ resources tables, and then removes the sum fields that are now unused.

The one exception is the `project_resources.quota` field. Since the old quota distribution model is not AZ-aware, we continue storing quota on the resource level for now. The new quota distribution model will use the prepared `project_az_resources.quota` field once it arrives.

### DIFF
--- a/internal/collector/metrics.go
+++ b/internal/collector/metrics.go
@@ -549,7 +549,7 @@ func (c *DataMetricsCollector) Collect(ch chan<- prometheus.Metric) {
 
 		var (
 			capacityPerAZ map[limes.AvailabilityZone]uint64
-			usagePerAZ    map[limes.AvailabilityZone]uint64
+			usagePerAZ    map[limes.AvailabilityZone]*uint64
 		)
 		err = json.Unmarshal([]byte(capacityPerAZJSON), &capacityPerAZ)
 		if err != nil {
@@ -577,10 +577,10 @@ func (c *DataMetricsCollector) Collect(ch chan<- prometheus.Metric) {
 				)
 
 				azUsage := usagePerAZ[az]
-				if azUsage != 0 {
+				if azUsage != nil && *azUsage != 0 {
 					ch <- prometheus.MustNewConstMetric(
 						clusterUsagePerAZDesc,
-						prometheus.GaugeValue, float64(azUsage),
+						prometheus.GaugeValue, float64(*azUsage),
 						string(az), serviceType, resourceName,
 					)
 				}
--- a/internal/collector/metrics.go
+++ b/internal/collector/metrics.go
@@ -27,7 +27,7 @@ import (
 
 	"github.com/go-gorp/gorp/v3"
 	"github.com/prometheus/client_golang/prometheus"
-	limesresources "github.com/sapcc/go-api-declarations/limes/resources"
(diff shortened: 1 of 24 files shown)


internal/collector/metrics.go
```diff
@@ -455,9 +455,11 @@ func (c *DataMetricsCollector) Describe(ch chan<- *prometheus.Desc) {
 }
 
 var clusterMetricsQuery = sqlext.SimplifyWhitespace(`
-	SELECT cs.type, cr.name, cr.capacity, cr.capacity_per_az
+	SELECT cs.type, cr.name, JSON_OBJECT_AGG(car.az, car.raw_capacity), JSON_OBJECT_AGG(car.az, car.usage)
 	  FROM cluster_services cs
 	  JOIN cluster_resources cr ON cr.service_id = cs.id
+	  JOIN cluster_az_resources car ON car.resource_id = cr.id
+	 GROUP BY cs.type, cr.name
```
SuperSandro2000, line 462: I am confused. Are we using first tabs and then one, two or aligned spaces?

internal/reports/utils.go
```diff
@@ -46,3 +60,25 @@ func mergeMaxTime(lhs *limes.UnixEncodedTime, rhs *time.Time) *limes.UnixEncoded
...
+	}
+	if string(*target) == "" || string(*target) == "[]" {
+		*target = json.RawMessage(input)
+		return
+	}
+	*target = json.RawMessage(fmt.Sprintf("%s,%s",
+		strings.TrimSuffix(string(*target), "]"),
+		strings.TrimPrefix(input, "["),
+	))
+}
```
SuperSandro2000, line 84: Do we have explizit tests for this?

internal/db/migrations.go
```diff
@@ -368,4 +368,32 @@ var sqlMigrations = map[string]string{
...
+	"030_drop_redundant_resources_columns.up.sql": `
+		ALTER TABLE cluster_resources
+			DROP COLUMN capacity,
+			DROP COLUMN subcapacities,
+			DROP COLUMN capacity_per_az;
+		ALTER TABLE project_resources
+			DROP COLUMN usage,
+			DROP COLUMN subresources,
+			DROP COLUMN physical_usage;
+	`,
```
VoigtS, line 380: Why are the subcapacities and az capacities stored as text?

VoigtS: Also: what exactly counts as a subcapacity?

internal/collector/metrics.go
majewsky: This file follows my existing pattern of right-aligning the leading keywords of each line. The tab indentation for the entire thing is following gofmt style, the spaces after ensure the right-alignment.

(I'm not saying this to argue in favor of this style, just explaining the original reasoning. I would like to avoid opening a side column to adjudicate SQL indentation styles.)

internal/reports/utils.go
majewsky: I did not consider it necessary because this is covered by the report generation, and it's also not a public interface.

PR #375, opened 2023-10-30
add new PerAZ field to ProjectResourceReport

majewsky:
Since this increases the report size substantially, this is gated behind the HTTP header `X-Limes-V2-API-Preview: per-az` for now, which will only be used by clients that actually need the AZ breakdown.

The API spec does not reflect this, since this is currently a "hidden API". It will become fully official when we bump the API to v2. At that point, I will also update the API spec to reflect these new capabilities.

### DIFF
--- a/internal/reports/project.go
+++ b/internal/reports/project.go
@@ -71,7 +71,7 @@ var (
 // reports with the highest detail levels can be several MB large, we don't just
 // return them all in a big list. Instead, the `submit` callback gets called
 // once for each project report once that report is complete.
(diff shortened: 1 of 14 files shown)


internal/reports/project.go
```diff
@@ -224,13 +257,16 @@ func GetProjectResources(cluster *core.Cluster, domain db.Domain, project *db.Pr
...
 	if projectReport != nil {
-		finalizeProjectReport(projectReport)
+		err := finalizeProjectResourceReport(projectReport, currentProjectID, now, dbi, filter)
+		if err != nil {
+			return err
+		}
 		return submit(projectReport)
 	}
 	return nil
 }
```
VoigtS, line 268: Where is "now" used in this?

PR #379, opened 2023-11-07
ensure that each project_resource has at least one project_az_resource

majewsky:
This is a useful invariant to uphold to simplify reporting queries, but it could be violated by quota plugins that calculate usage by counting or summing, where no usage in any AZ would be reported if there are no items to count or sum in an empty project.


VoigtS (approved):
Looks good to me.

PR #385, opened 2023-12-12
confirm pending commitments after capacity scrape

majewsky:
- [x] implementation
- [x] tests


VoigtS (approved):
This looks good! The unit tests are also well crafted.

PR #388, opened 2023-12-15
add project_az_resources.historical_usage and fill during scraping 

majewsky:
We need this data to evaluate the autogrow quota distribution algorithm later on.
This is split into two commits which I recommend reviewing separately.


VoigtS (approved):
This reflects the timestamp handling that we designed on paper.
Approved.

PR #410, opened 2024-03-05
Introduce TransferCommitment API

VoigtS:
productionChecklist:

This PR introduces commitment transfers from a source to a target project.
Two API's will be implemented:
- [x] start-transfer: Either marks a commitment as transferrable or splits the commitment into two parts where the provided amount will be the indicator for the transferrable commitment
- [x] transfer-commitment: Where the actual commitment transfer from a source to a target AZ_RESOURCE_ID will be executed.

### DIFF
--- a/docs/users/api-spec-resources.md
+++ b/docs/users/api-spec-resources.md
@@ -758,7 +758,7 @@ Prepares a commitment to be transferred from a source project to a target projec
 {
   "commitment": {
     "amount": 100,
-    "transfer_status": "unlisted",
+    "transfer_status": "unlisted"
   }
 }
 ```
--- a/docs/users/api-spec-resources.md
+++ b/docs/users/api-spec-resources.md
@@ -773,9 +773,11 @@ The response is a JSON of the commitment including the following fields that ide
   }
 }
 ```
-### POST /v1/domains/:id/projects/:id/transfer-commitment/:id?token=:token
+### POST /v1/domains/:id/projects/:id/transfer-commitment/:id
 Transfers the commitment from a source project to a target project.
 Requires a project-admin token.
+Requires a transfer token in the request header:
+`Transfer-Token: [value]`.
 This endpoint receives the target project ID, but the commitment ID from the source project.
 Requires a generated token from the API: `/v1/domains/:id/projects/:id/commitments/:id/start-transfer`.
 On success the API clears the `transfer_token` and `transfer_status` from the commitment.
--- a/docs/users/api-spec-resources.md
+++ b/docs/users/api-spec-resources.md
@@ -775,6 +775,7 @@ The response is a JSON of the commitment including the following fields that ide
 ```
 ### POST /v1/domains/:id/projects/:id/transfer-commitment/:id?token=:token
 Transfers the commitment from a source project to a target project.
+Requires a project-admin token.
 This endpoint receives the target project ID, but the commitment ID from the source project.
 Requires a generated token from the API: `/v1/domains/:id/projects/:id/commitments/:id/start-transfer`.
 On success the API clears the `transfer_token` and `transfer_status` from the commitment.
--- a/docs/users/api-spec-resources.md
+++ b/docs/users/api-spec-resources.md
@@ -769,7 +769,7 @@ The response is a JSON of the commitment including the following fields that ide
 {
   "commitment": {
     "transfer_token": "token",
-    "transfer_status": "unlisted",
+    "transfer_status": "unlisted"
   }
 }
 ```
--- a/docs/users/api-spec-resources.md
+++ b/docs/users/api-spec-resources.md
@@ -764,7 +764,7 @@ Prepares a commitment to be transferred from a source project to a target projec
 ```
 If the amount to transfer is equal to the commitment, the whole commitment will be marked as transferrable. If the amount is less than the commitment, the commitment will be split in two and the requested amount will be marked as transferrable.
 The transfer status indicates if the commitment stays `unlisted` (private) or `public`.
-The response is a JSON of the commitment including the following fields that identify a commitment in it's transferrable state:
+The response is a JSON of the commitment including the following fields that identify a commitment in its transferrable state:
 ```json
 {
   "commitment": {
--- a/docs/users/api-spec-resources.md
+++ b/docs/users/api-spec-resources.md
@@ -762,7 +762,9 @@ Prepares a commitment to be transferred from a source project to a target projec
   }
 }
 ```
-If the amount to transfer is equal to the commitment, the whole commitment will be marked as transferrable. If the amount is less than the commitment, the commitment will be split in two and the requested amount will be marked as transferrable.The transfer status indicates if the commitment stays `unlisted` (private) or `public`. The response is a JSON of the commitment including the following fields that identify a commitment in it's transferrable state:
+If the amount to transfer is equal to the commitment, the whole commitment will be marked as transferrable. If the amount is less than the commitment, the commitment will be split in two and the requested amount will be marked as transferrable.
+The transfer status indicates if the commitment stays `unlisted` (private) or `public`.
+The response is a JSON of the commitment including the following fields that identify a commitment in it's transferrable state:
 ```json
 {
   "commitment": {
--- a/docs/users/api-spec-resources.md
+++ b/docs/users/api-spec-resources.md
@@ -752,6 +752,28 @@ Returns 200 (OK) on success, and a JSON document like `{"result":true}` or `{"re
 
 The `result` field indicates whether this commitment can be created without a `confirm_by` attribute, that is, confirmed immediately upon creation.
 
+### POST /v1/domains/:id/projects/:id/commitments/:id/start-transfer
+Prepares a commitment to be transferred from a source project to a target project. Requires a project-admin token, and a request body that is a JSON document like:
+```json
+{
+  "commitment": {
+    "amount": 100,
+    "transfer_status": "unlisted",
+  }
+}
+```
+If the amount to transfer is equal to the commitment, the whole commitment will be marked as transferrable. If the amount is less than the commitment, the commitment will be split in two and the requested amount will be marked as transferrable.The transfer status indicates if the commitment stays `unlisted` (private) or `public`. The response is a JSON of the commitment including the following fields that identify a commitment in it's transferrable state:
+```json
+{
+  "commitment": {
+    "transfer_token": "token",
+    "transfer_status": "unlisted",
+  }
+}
+```
+### POST /v1/domains/:id/projects/:id/transfer-commitment/:id?token=:token
+Transfers the commitment from a source project to a target project. This endpoint receives the target project ID, but the commitment ID from the source project. Requires a generated token from the API: `/v1/domains/:id/projects/:id/commitments/:id/start-transfer`. On success the API clears the `transfer_token` and `transfer_status` from the commitment. After that, it returns the commitment as a JSON document.  
+
 ### DELETE /v1/domains/:domain\_id/projects/:project\_id/commitments/:id
 
 Deletes a commitment within the given project. Requires a cloud-admin token. On success, returns 204 (No Content).
--- a/docs/users/api-spec-resources.md
+++ b/docs/users/api-spec-resources.md
@@ -774,7 +774,11 @@ The response is a JSON of the commitment including the following fields that ide
 }
 ```
 ### POST /v1/domains/:id/projects/:id/transfer-commitment/:id?token=:token
-Transfers the commitment from a source project to a target project. This endpoint receives the target project ID, but the commitment ID from the source project. Requires a generated token from the API: `/v1/domains/:id/projects/:id/commitments/:id/start-transfer`. On success the API clears the `transfer_token` and `transfer_status` from the commitment. After that, it returns the commitment as a JSON document.  
+Transfers the commitment from a source project to a target project.
+This endpoint receives the target project ID, but the commitment ID from the source project.
+Requires a generated token from the API: `/v1/domains/:id/projects/:id/commitments/:id/start-transfer`.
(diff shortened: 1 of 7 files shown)


docs/users/api-spec-resources.md
```diff
@@ -752,6 +752,28 @@ Returns 200 (OK) on success, and a JSON document like `{"result":true}` or `{"re
 
 The `result` field indicates whether this commitment can be created without a `confirm_by` attribute, that is, confirmed immediately upon creation.
 
+### POST /v1/domains/:id/projects/:id/commitments/:id/start-transfer
+Prepares a commitment to be transferred from a source project to a target project. Requires a project-admin token, and a request body that is a JSON document like:
+```json
+{
+  "commitment": {
+    "amount": 100,
+    "transfer_status": "unlisted",
```
SuperSandro2000, line 761:
```suggestion
    "transfer_status": "unlisted"
```

That's not valid json 🙈

docs/users/api-spec-resources.md
```diff
@@ -752,6 +752,28 @@ Returns 200 (OK) on success, and a JSON document like `{"result":true}` or `{"re
...
+  }
+}
+```
+If the amount to transfer is equal to the commitment, the whole commitment will be marked as transferrable. If the amount is less than the commitment, the commitment will be split in two and the requested amount will be marked as transferrable.The transfer status indicates if the commitment stays `unlisted` (private) or `public`. The response is a JSON of the commitment including the following fields that identify a commitment in it's transferrable state:
+```json
+{
+  "commitment": {
+    "transfer_token": "token",
+    "transfer_status": "unlisted",
```
SuperSandro2000, line 770:
```suggestion
    "transfer_status": "unlisted"
```

docs/users/api-spec-resources.md
```diff
@@ -752,6 +752,28 @@ Returns 200 (OK) on success, and a JSON document like `{"result":true}` or `{"re
...
+  }
+}
+```
+### POST /v1/domains/:id/projects/:id/transfer-commitment/:id?token=:token
+Transfers the commitment from a source project to a target project. This endpoint receives the target project ID, but the commitment ID from the source project. Requires a generated token from the API: `/v1/domains/:id/projects/:id/commitments/:id/start-transfer`. On success the API clears the `transfer_token` and `transfer_status` from the commitment. After that, it returns the commitment as a JSON document.
```
SuperSandro2000, line 775:
```suggestion
Transfers the commitment from a source project to a target project.
This endpoint receives the target project ID, but the commitment ID from the source project.
Requires a generated token from the API: `/v1/domains/:id/projects/:id/commitments/:id/start-transfer`.
On success the API clears the `transfer_token` and `transfer_status` from the commitment.
After that, it returns the commitment as a JSON document.  
```

internal/api/commitment.go
```diff
@@ -434,3 +457,215 @@ func (p *v1Provider) DeleteProjectCommitment(w http.ResponseWriter, r *http.Requ
...
+	if !token.Require(w, "project:edit") {
+		return
+	}
+	dbDomain := p.FindDomainFromRequest(w, r)
+	if dbDomain == nil {
+		return
+	}
+	dbProject := p.FindProjectFromRequest(w, r, dbDomain)
+	if dbProject == nil {
+		return
```
SuperSandro2000, line 474: Shouldn't we return an error here?

internal/api/commitment_test.go
```diff
@@ -542,3 +560,211 @@ func TestDeleteCommitmentErrorCases(t *testing.T) {
...
+		Method:       "POST",
+		Path:         "/v1/domains/uuid-for-germany/projects/uuid-for-berlin/commitments/1/start-transfer",
+		ExpectStatus: http.StatusAccepted,
+		ExpectBody:   assert.JSONObject{"commitment": resp2},
+		Body:         assert.JSONObject{"commitment": assert.JSONObject{"amount": 9, "transfer_status": "public"}},
+	}.Check(t, s.Handler)
+
+	// Test on unconfirmed commitment should fail.
+	// ID is 4, because 2 additional commitments were created previously.
+	var confirmBy = time.Now().Unix()
```
SuperSandro2000, line 648: Shouldn't we use the mock clock here?

internal/api/utils.go
```diff
@@ -44,3 +46,12 @@ func unwrapOrDefault[T any](value *T, defaultValue T) T {
 	}
 	return *value
 }
+
+func GenerateToken() string {
+	tokenBytes := make([]byte, 12)
```
SuperSandro2000, line 51:
```suggestion
	tokenBytes := make([]byte, 24)
```
maybe 12 is a bit short

internal/api/commitment_test.go
VoigtS: That slipped through. Thank you.

internal/api/commitment.go
```diff
@@ -434,3 +457,222 @@ func (p *v1Provider) DeleteProjectCommitment(w http.ResponseWriter, r *http.Requ
...
+	}
+
+	// Mark whole commitment or a newly created, splitted one as transferrable.
+	tx, err := p.DB.Begin()
+	if respondwith.ErrorText(w, err) {
+		return
+	}
+	defer sqlext.RollbackUnlessCommitted(tx)
+	transferToken := p.generateTransferToken()
+	if req.Amount >= dbCommitment.Amount {
```
majewsky, line 521: A requested amount greater than the commitment amount should be treated as an error and caught during validation.

internal/api/commitment.go
```diff
@@ -434,3 +457,222 @@ func (p *v1Provider) DeleteProjectCommitment(w http.ResponseWriter, r *http.Requ
...
+		DomainID:    dbDomain.UUID,
+		DomainName:  dbDomain.Name,
+		ProjectID:   dbProject.UUID,
+		ProjectName: dbProject.Name,
+		Commitment:  c,
+	})
+	respondwith.JSON(w, http.StatusAccepted, map[string]any{"commitment": c})
+}
+
+func (p *v1Provider) splitCommitment(dbCommitment db.ProjectCommitment, amount uint64) db.ProjectCommitment {
```
majewsky, line 575: This does not actually "do" anything to the database, as the name implies. I suggest the following name instead:
```suggestion
func (p *v1Provider) buildSplitCommitment(dbCommitment db.ProjectCommitment, amount uint64) db.ProjectCommitment {
```
I'm also open to other suggestions if you have any.

internal/api/commitment.go
```diff
@@ -434,3 +457,222 @@ func (p *v1Provider) DeleteProjectCommitment(w http.ResponseWriter, r *http.Requ
...
+func (p *v1Provider) TransferCommitment(w http.ResponseWriter, r *http.Request) {
+	httpapi.IdentifyEndpoint(r, "/v1/domains/:id/projects/:id/transfer-commitment/:id?token=:token")
+	token := p.CheckToken(r)
+	if !token.Require(w, "project:edit") {
+		http.Error(w, "insufficient access rights.", http.StatusForbidden)
+		return
+	}
+	transferToken := r.URL.Query().Get("token")
+	if transferToken == "" {
+		respondwith.ErrorText(w, errors.New("no transfer token provided"))
```
majewsky, line 601: `respondwith.ErrorText` sends a 500 response (Internal Server Error) and thus should not be used for client errors. `http.Error` is the correct choice here.

internal/api/commitment_test.go
```diff
@@ -542,3 +560,211 @@ func TestDeleteCommitmentErrorCases(t *testing.T) {
...
+		test.WithDBFixtureFile("fixtures/start-data-commitments.sql"),
+		test.WithConfig(testCommitmentsYAMLWithoutMinConfirmDate),
+		test.WithAPIHandler(NewV1API),
+	)
+
+	var transferToken = test.GenerateDummyToken()
+
+	// Test on confirmed commitment should succeed.
+	// TransferAmount >= CommitmentAmount
+	req1 := func(transfer_status string) assert.JSONObject {
```
majewsky, line 575: Go does not usually use snake_case:
```suggestion
	req1 := func(transferStatus string) assert.JSONObject {
```

internal/api/commitment_test.go
```diff
@@ -542,3 +560,211 @@ func TestDeleteCommitmentErrorCases(t *testing.T) {
...
+		Path:         "/v1/domains/uuid-for-germany/projects/uuid-for-berlin/commitments/new",
+		Body:         assert.JSONObject{"commitment": req1("")},
+		ExpectStatus: http.StatusCreated,
+	}.Check(t, s.Handler)
+
+	assert.HTTPRequest{
+		Method:       "POST",
+		Path:         "/v1/domains/uuid-for-germany/projects/uuid-for-berlin/commitments/1/start-transfer",
+		ExpectStatus: http.StatusAccepted,
+		ExpectBody:   assert.JSONObject{"commitment": resp1},
```
majewsky, line 616: Continuing from my previous remark earlier in the request body parsing, this should only send the required attributes. (And when creating the commitment, `transfer_status` and `transfer_token` should not be sent because those are not applicable for the `/new` endpoint.) The test should be clear about what is actually required to induce the expected behavior.

internal/api/commitment_test.go
```diff
@@ -542,3 +560,211 @@ func TestDeleteCommitmentErrorCases(t *testing.T) {
...
+		Body:         assert.JSONObject{"commitment": req2},
+		ExpectStatus: http.StatusCreated,
+	}.Check(t, s.Handler)
+
+	assert.HTTPRequest{
+		Method:       "POST",
+		Path:         "/v1/domains/uuid-for-germany/projects/uuid-for-berlin/commitments/4/start-transfer",
+		ExpectStatus: http.StatusUnprocessableEntity,
+		Body:         assert.JSONObject{"commitment": assert.JSONObject{"amount": 10, "transfer_status": "unlisted"}},
+	}.Check(t, s.Handler)
```
majewsky, line 673: When checking error cases, please also check that the error message matches the expected error condition:
```go
ExpectBody: assert.StringData("something something\n"),
```

internal/api/core.go
```diff
@@ -110,6 +113,11 @@ func NewTokenValidator(provider *gophercloud.ProviderClient, eo gophercloud.Endp
 	return &tv, err
 }
 
+func (p *v1Provider) OverrideGenerateToken(generateTransferToken func() string) *v1Provider {
```
majewsky, line 116: I would like this naming to be more consistent:
```suggestion
func (p *v1Provider) OverrideGenerateTransferToken(generateTransferToken func() string) *v1Provider {
```

internal/api/utils.go
```diff
@@ -44,3 +46,12 @@ func unwrapOrDefault[T any](value *T, defaultValue T) T {
 	}
 	return *value
 }
+
+func GenerateToken() string {
```
majewsky, line 50: Picking up from the naming remark from before:
```suggestion
func GenerateTransferToken() string {
```
Also, please add a docstring to explain the usage context.

majewsky (changes requested):
Please do not be disheartened by the number of remarks. You did a good job getting accustomed to the existing code structure and following house conventions. But there is some work left to do, both in the stylistic and functional departments:

internal/api/commitment_test.go
VoigtS: Fixed: this was a misunderstanding. The resource body should contain the commitment.

internal/api/utils.go
VoigtS: Please check the docstring that I added to the function.

VoigtS:
Reminder: Check on Monday - Previous OLTP implementation had a bug in the transferred commitment.
Only the return value contained the Transferstatus and Token of the splitted commitment `transferCommitment`.
However, the values were not mirrored in the database.

The ORM implementation now takes this into account and saves the values in the database. A additional test case was added that tests the transfer of a splitted commitment.

internal/api/commitment.go
```diff
@@ -434,3 +448,218 @@ func (p *v1Provider) DeleteProjectCommitment(w http.ResponseWriter, r *http.Requ
...
+	}
+	targetProject := p.FindProjectFromRequest(w, r, dbDomain)
+	if targetProject == nil {
+		http.Error(w, "project not found.", http.StatusNotFound)
+		return
+	}
+
+	// find commitment by transfer_token
+	var dbCommitment db.ProjectCommitment
+	err := p.DB.SelectOne(&dbCommitment, getCommitmentWithMatchingTransferTokenQuery, transferToken)
```
majewsky, line 621: Should also match on ID, as defense in depth.

internal/api/commitment.go
```diff
@@ -434,3 +448,218 @@ func (p *v1Provider) DeleteProjectCommitment(w http.ResponseWriter, r *http.Requ
...
+	if errors.Is(err, sql.ErrNoRows) {
+		http.Error(w, "no matching commitment found", http.StatusNotFound)
+		return
+	} else if respondwith.ErrorText(w, err) {
+		return
+	}
+
+	// get target AZ_RESOURCE_ID
+	var targetResourceID db.ProjectAZResourceID
+	err = p.DB.QueryRow(findTargetAZResourceIDBySourceIDQuery, targetProject.ID, dbCommitment.AZResourceID).Scan(&targetResourceID)
```
majewsky, line 631: This looks like it does not match on service type and resource name. How is the commitment matched to the correct service and resource on the receiving side?

The tests can plausibly work "by accident" because they all use `first/capacity` which is the alphabetically first service/resource.

majewsky (changes requested):

internal/api/commitment.go
VoigtS: The query now respects the commitment ID (read out of the URL)

VoigtS:
Fixed: Movement of the transfer token from queryString into the request header.

majewsky (approved):

PR #419, opened 2024-03-27
allow project admins to delete commitments up to 24h after creation

majewsky:
This was more complicated than I anticipated, so I would appreciate a second set of eyes.


VoigtS (approved):

VoigtS:
Well done, this looks good to me.

PR #422, opened 2024-04-04
confirm available capacity when moving commitments

majewsky:
This fixes a class of bugs where, in the worst case, a malicious user could create arbitrarily many commitments, even though commitments are supposed to be bound by the available capacity. The infinite-commitment glitch would work as follows:

1. Have admin access to two projects (A and B), where project A has some non-zero usage of a resource. Project A starts out without commitments for that resource.
2. Create an immediately confirmed commitment in project A covering the full amount of used resource.
3. Move this commitment to project B.
4. Go back to step 2 and repeat ad nauseam.

The issue here is that the invariant is not just `sum_over_projects(committed) <= capacity`, but `sum_over_projects(max(committed, usage)) <= capacity`. Moving commitments between projects with different usage levels does not always uphold this invariant. Instead, it's sometimes necessary to force commitments to "stick" to the usage that they cover in their existing project.

### DIFF
--- a/internal/datamodel/allocation_stats.go
+++ b/internal/datamodel/allocation_stats.go
@@ -39,15 +39,17 @@ import (
 // - CanMoveExistingCommitment
 // - ConfirmPendingCommitments
 type clusterAZAllocationStats struct {
-	Capacity     uint64
-	ProjectStats map[db.ProjectServiceID]projectAZAllocationStats
+	Capacity uint64
+	// Using db.ProjectResourceID as a key here is only somewhat arbitrary:
+	// ProjectServiceID and ProjectID could also be used, but then we would have to use more JOINs in some queries.
+	ProjectStats map[db.ProjectResourceID]projectAZAllocationStats
 }
 
-func (c clusterAZAllocationStats) FitsAfterCommitmentChanges(additions, subtractions map[db.ProjectServiceID]uint64) bool {
+func (c clusterAZAllocationStats) FitsAfterCommitmentChanges(additions, subtractions map[db.ProjectResourceID]uint64) bool {
 	// calculate `sum_over_projects(max(committed, usage))` including the requested changes
 	usedCapacity := uint64(0)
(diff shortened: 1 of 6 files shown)


internal/datamodel/allocation_stats.go
```diff
@@ -35,27 +35,33 @@ import (
...
+func (c clusterAZAllocationStats) FitsAfterCommitmentChanges(additions, subtractions map[db.ProjectServiceID]uint64) bool {
+	// calculate `sum_over_projects(max(committed, usage))` including the requested changes
 	usedCapacity := uint64(0)
 	for projectServiceID, stats := range c.ProjectStats {
-		if projectServiceID == serviceID {
-			usedCapacity += max(stats.Committed+amount, stats.Usage)
-		} else {
-			usedCapacity += max(stats.Committed, stats.Usage)
-		}
+		committed := saturatingSub(stats.Committed+additions[projectServiceID], subtractions[projectServiceID])
```
VoigtS, line 49: A question I had reading the previous code: Why do we use `projectServiceID` as a key for the project stats? Reading the commitments and usage I would assume this happens at Resource level. Expecting the resource ID.
But we are working on Service level here.

VoigtS (dismissed):

internal/datamodel/allocation_stats.go
majewsky: I believe I chose ProjectServiceID as key because I had `project_services` joined into all relevant queries anyway, in order to match on `project_services.type`. ProjectResourceID is also a reasonable choice. I had a look at the queries and found that one query can be simplified by keying on ProjectResourceID instead, so let's go with that.

VoigtS (approved):

internal/datamodel/allocation_stats.go
VoigtS: Thank you very much for the explanation and the effort put into this!

PR #426, opened 2024-04-11
add PUT /v1/domains/:id/projects/:id/max-quota

majewsky:
This is a very temporary API to fill the new `project_resources.max_quota_from_admin` field. I did not add this to the API spec because it's only a hidden feature for now. It will become properly documented in the v2 API that we will for sure do this year, no way that's gonna get delayed, I promise.

### DIFF
--- a/internal/api/api_test.go
+++ b/internal/api/api_test.go
@@ -34,6 +34,7 @@ import (
 	limesrates "github.com/sapcc/go-api-declarations/limes/rates"
 	limesresources "github.com/sapcc/go-api-declarations/limes/resources"
 	"github.com/sapcc/go-bits/assert"
+	"github.com/sapcc/go-bits/easypg"
 	"github.com/sapcc/go-bits/sqlext"
 
 	"github.com/sapcc/limes/internal/core"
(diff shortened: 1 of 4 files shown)


internal/api/api_test.go
```diff
@@ -2081,3 +2082,108 @@ func Test_StrictDomainQuotaLimit(t *testing.T) {
...
+			Method:       "PUT",
+			Path:         "/v1/domains/uuid-for-germany/projects/uuid-for-berlin/max-quota",
+			Body:         makeRequest("shared", assert.JSONObject{"name": "things", "max_quota": value}),
+			ExpectStatus: http.StatusAccepted,
+		}.Check(t, s.Handler)
+		tr.DBChanges().AssertEqualf(`
+			UPDATE project_resources SET max_quota_from_admin = %d WHERE id = 4 AND service_id = 2 AND name = 'things';
+		`, value)
+	}
+
```
VoigtS, line 2115: I understand the main portion of the API now, what's unclear to me is what happens if the user sets the quota below the current value of the commitments or usage?

VoigtS (changes requested):

internal/api/api_test.go
majewsky: It won't work. ACPQ grants the hard-minimum quota without checking constraints: https://github.com/sapcc/limes/blob/e025b56d67c24b54825c843390c2e860e839a0f4/internal/datamodel/apply_computed_project_quota.go#L280-L285

Constraints are applied to `Desired` before applying it to `Allocated`. The hard minimum starts out in `Allocated`. This is also covered by unit tests as far as I remember.

VoigtS (approved):

PR #464, opened 2024-05-21
capacity-manila: consider ResourceDemand when splitting capacity/usage between shares and snapshots

majewsky:
Checklist:


SuperSandro2000 (approved):
doesn't look horrible

PR #472, opened 2024-05-28
commitments: allow self-service deletion of confirmed commitments that are still fresh

majewsky:


frr2018:
Allowing the deletion even if the CR is confirmed makes total sense. The deletion has to work in both cases because such a CR could be created by accident as well.

PR #478, opened 2024-06-17
plugins/nova: remove usage of GET os-quota-class-sets/flavors

majewsky:
This used to be where our Ironic team indicates to Nova which flavors use separate instance quotas. This is duplicate with the existing `quota:separate` extra spec on the flavor and caused frequent inconsistencies. The Nova implementation has now been migrated to check the `quota:separate` extra spec directly.

I tested this in QA; `test-get-quota` output did not change from before to after.


VoigtS (changes requested):
Where is the `quota:separate` spec coming from? I'm a bit confused by it.

majewsky:
It's put on the flavor by seeds, e.g. https://github.com/sapcc/helm-charts/blob/bf8a15a0d95ad75d4719cfe4a0fc4840bb8b059c/openstack/ironic/templates/seed.yaml#L177

VoigtS (approved):

PR #489, opened 2024-06-27
Add max and minUsage to project report

VoigtS:
This PR aims to include the values from the historical_usage into the `GET` projects API.
Only if historical_usage exists, the minUsage, maxUsage and the configured retention duration will be included into the AZ report.


internal/reports/project.go
```diff
@@ -118,14 +119,15 @@ func GetProjectResources(cluster *core.Cluster, domain db.Domain, project *db.Pr
 			azQuota           *uint64
 			azUsage           *uint64
 			azPhysicalUsage   *uint64
+			azHisoricalUsage  *string
```
majewsky, line 122:
```suggestion
			azHistoricalUsage  *string
```

internal/reports/project.go
```diff
@@ -236,6 +238,23 @@ func GetProjectResources(cluster *core.Cluster, domain db.Domain, project *db.Pr
 				PhysicalUsage: azPhysicalUsage,
 				Subresources:  json.RawMessage(*azSubresources),
 			}
+
+			if *azHisoricalUsage != "" {
+				config := cluster.QuotaDistributionConfigForResource(*serviceType, *resourceName)
+				retentionPeriod := config.Autogrow.UsageDataRetentionPeriod
+				duration := limesresources.CommitmentDuration{
+					Short: retentionPeriod.Into(),
+				}
```
majewsky, line 247: If alternative quota distribution models are added again in the future, this might fail. Please add a check for `config.Autogrow != nil`, and otherwise keep the duration field zero-valued as a fallback.

majewsky (changes requested):

internal/reports/project.go
VoigtS: I updated it, please have another look.

majewsky (approved):

PR #502, opened 2024-07-25
Add GET commitment by token endpoint

VoigtS:
The transfer commitment endpoint is fine in scenarios when it get called after the start-transfer endpoint, because it returns the commitment first.
Because new requirments arise that require to transfer a commitment between projects, the start-transfer endpoint and the transfer endpoint are now decoupled from each other. To mitigate this, I would like to introduce an endpoint that gets a commitment by its token.

This approach also ensures that a commitment can be checked for its details (resource, AZ) before it actually gets transferred to the target project. This is an important detail for the UI.

### DIFF
--- a/docs/users/api-spec-resources.md
+++ b/docs/users/api-spec-resources.md
@@ -592,6 +592,9 @@ Requires a generated token from the API: `/v1/domains/:id/projects/:id/commitmen
 On success the API clears the `transfer_token` and `transfer_status` from the commitment.
 After that, it returns the commitment as a JSON document.  
 
+### GET "/v1/domains/:id/projects/:id/commitments/:token"
+To ensure that a commitment can be checked for its `resource` type or `availability zone` before it gets transferred to a target project, this endpoint fetches the target commitment by its respective token.
+
 ### DELETE /v1/domains/:domain\_id/projects/:project\_id/commitments/:id
 
 Deletes a commitment within the given project. Requires a cloud-admin token. On success, returns 204 (No Content).
--- a/docs/users/api-spec-resources.md
+++ b/docs/users/api-spec-resources.md
@@ -592,7 +592,7 @@ Requires a generated token from the API: `/v1/domains/:id/projects/:id/commitmen
 On success the API clears the `transfer_token` and `transfer_status` from the commitment.
 After that, it returns the commitment as a JSON document.  
 
-### GET "/v1/domains/:id/projects/:id/commitments/:token"
+### GET /v1/domains/:id/projects/:id/commitments/:token
 To ensure that a commitment can be checked for its `resource` type or `availability zone` before it gets transferred to a target project, this endpoint fetches the target commitment by its respective token.
 
 ### DELETE /v1/domains/:domain\_id/projects/:project\_id/commitments/:id
--- a/docs/users/api-spec-resources.md
+++ b/docs/users/api-spec-resources.md
@@ -592,7 +592,7 @@ Requires a generated token from the API: `/v1/domains/:id/projects/:id/commitmen
 On success the API clears the `transfer_token` and `transfer_status` from the commitment.
 After that, it returns the commitment as a JSON document.  
 
-### GET /v1/domains/:id/projects/:id/commitments/:token
+### GET /v1/commitments/:token
 To ensure that a commitment can be checked for its `resource` type or `availability zone` before it gets transferred to a target project, this endpoint fetches the target commitment by its respective token.
 
 ### DELETE /v1/domains/:domain\_id/projects/:project\_id/commitments/:id
--- a/internal/api/commitment.go
+++ b/internal/api/commitment.go
@@ -641,32 +641,11 @@ func (p *v1Provider) GetCommitmentByTransferToken(w http.ResponseWriter, r *http
 		return
 	}
 
-	// A token might not be unique in very rare edge cases. We handle this here.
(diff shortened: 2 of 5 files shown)


docs/users/api-spec-resources.md
```diff
@@ -592,6 +592,9 @@ Requires a generated token from the API: `/v1/domains/:id/projects/:id/commitmen
 On success the API clears the `transfer_token` and `transfer_status` from the commitment.
 After that, it returns the commitment as a JSON document.  
 
+### GET "/v1/domains/:id/projects/:id/commitments/:token"
```
majewsky, line 595: Style nitpick:
```suggestion
### GET /v1/domains/:id/projects/:id/commitments/:token
```

internal/api/commitment.go
```diff
@@ -87,6 +87,9 @@ var (
 	getCommitmentWithMatchingTransferTokenQuery = sqlext.SimplifyWhitespace(`
 		SELECT * FROM project_commitments WHERE id = $1 AND transfer_token = $2
 	`)
+	findCommitmentByTransferToken = sqlext.SimplifyWhitespace(`
+		SELECT * from project_commitments where transfer_token = $1
```
majewsky, line 91: Style nitpick:
```suggestion
		SELECT * FROM project_commitments WHERE transfer_token = $1
```

internal/api/commitment.go
```diff
@@ -624,6 +627,83 @@ func (p *v1Provider) buildSplitCommitment(dbCommitment db.ProjectCommitment, amo
...
+	}
+	if rowCount > 1 {
+		http.Error(w, "detected multiple commitments with the same token.", http.StatusConflict)
+		return
+	}
+
+	var dbCommitment db.ProjectCommitment
+	err = p.DB.SelectOne(&dbCommitment, findCommitmentByTransferToken, transferToken)
+	if errors.Is(err, sql.ErrNoRows) {
+		http.Error(w, "commitment unexpectingly not found.", http.StatusNotFound)
```
majewsky, line 679:
```suggestion
		http.Error(w, "commitment unexpectedly not found.", http.StatusNotFound)
```

majewsky (changes requested):
Some nitpicks, some larger comments.

internal/api/commitment.go
```diff
@@ -624,6 +627,83 @@ func (p *v1Provider) buildSplitCommitment(dbCommitment db.ProjectCommitment, amo
...
+		rowCount += 1
+	}
+	if rowCount == 0 {
+		http.Error(w, "no matching commitment found.", http.StatusNotFound)
+		return
+	}
+	if rowCount > 1 {
+		http.Error(w, "detected multiple commitments with the same token.", http.StatusConflict)
+		return
+	}
```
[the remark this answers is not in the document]
VoigtS: For now I solved it with a partial index on the` transfer_token` column. It turned out the default value for this column is an empty string wich collides with a unique constraint. Because updating all the type definitions and implementations doesn't really feel worth it to implement a NULL column, I settled with this for now. What do you think about it?

internal/api/commitment.go
```diff
@@ -624,6 +627,45 @@ func (p *v1Provider) buildSplitCommitment(dbCommitment db.ProjectCommitment, amo
 	}
 }
 
+// GetCommitmentByTransferToken handles GET /v1/commitments/{token}
+func (p *v1Provider) GetCommitmentByTransferToken(w http.ResponseWriter, r *http.Request) {
+	httpapi.IdentifyEndpoint(r, "/v1/commitments/:token")
+	token := p.CheckToken(r)
+	if !token.Require(w, "project:show") {
+		http.Error(w, "insufficient access rights.", http.StatusForbidden)
```
majewsky, line 635: Require() already writes an error message, so this line is superfluous.

internal/api/commitment.go
```diff
@@ -624,6 +627,45 @@ func (p *v1Provider) buildSplitCommitment(dbCommitment db.ProjectCommitment, amo
...
+		return
+	}
+	transferToken := mux.Vars(r)["token"]
+	if transferToken == "" {
+		http.Error(w, "no transfer token provided", http.StatusBadRequest)
+		return
+	}
```
majewsky, line 642: This check is unreachable and can be removed. gorilla/mux already requires at least 1 character on each variable match.

internal/db/migrations.go
```diff
@@ -260,4 +260,8 @@ var sqlMigrations = map[string]string{
 		ALTER TABLE project_services
 			ADD COLUMN quota_sync_duration_secs REAL NOT NULL DEFAULT 0;
 	`,
+	"044_add_unique_key_to_transfer_token.up.sql": `
+		CREATE UNIQUE INDEX on project_commitments (transfer_token)
+            WHERE transfer_token <> '';
+	`,
```
majewsky, line 266: Please also add a down migration, even if only for completeness sake.

majewsky (changes requested):

internal/api/commitment.go
VoigtS: You are right, I had a similar case on the TransferCommitment endpoint which I cleaned up too.

internal/db/migrations.go
VoigtS: I added an empty down migration now.

VoigtS:
@majewsky Please check out #508 as an alternative solution proposition.

VoigtS:
Cancelling this PR, due to #508

PR #506, opened 2024-07-26
plugins/manila: fix scraping in regions without any snapmirror deployments

majewsky:


majewsky:
I need this to fix ongoing scraping errors in Limes, but none of the other approvers is on deck, so I'm merging this as a hotfix.

PR #514, opened 2024-08-07
add support for API-level resource renaming

majewsky:
I consider the core part of this (in `doc/` and `internal/core/`) to be fairly straightforward. In `internal/reports/`, all ServiceType and ResourceName variables are renamed to clarify their role:

- `serviceType` -> either `dbServiceType` or `apiIdentity.ServiceType`
- `resourceName` -> either `dbResourceName` or `apiIdentity.ResourceName`

The main complication arises because most of our report queries do outer joins when traversing from cluster/domains/projects into their respective services and resources. This allows us to select e.g. only project metadata when there is no service, or only service metadata when there is no resource in the service.

Specifically the last part, selecting a service without a resource, is not feasible anymore, because we can only evaluate the renaming rules on a full pair of service type or resource name, not for a service type alone. (In fact, in the general case, a DB-level service type might map into multiple API-level service types.)

After a lot of back and forth, I decided that the cleanest option is to convert these joins from outer to inner, to ensure that we always have a non-null service type and resource name on all result rows. This leads to a behavior change where service entries are never initialized if they have no resources, and domain/project entries are never initialized if they have no service.

The former is an API change that I'm willing to accept, because the API spec does not forbid it and there appear to be no users relying on it.

However, there are indeed users that query projects with `?service=none&resource=none` to only get project-level metadata. This is something that I'm likely to drop support for in the v2 API because they can just use the Keystone API if they only want project metadata, but for now I made sure that this usecase is still supported (this is covered by the existing tests using GET queries with `?service=unknown`).

- [x] add renaming config
- [x] make it work in API with the existing tests (that do not rename anything)
- [x] add renaming support to limes-data-metrics
- [x] add tests for reporting with renamed resources


VoigtS (approved):
This one was more difficult to get through, but I get the main gist of it, so we can give it a go. A confusion I had was with the unknown (none) service queries, but they return nil within the query subfunction so that they are able to process through.

PR #516, opened 2024-08-08
refactor metadata types on the QuotaPlugin API

majewsky:
The reasoning is explained in the individual commit messages. Since all three commits are completely unrelated to each other, I recommend reviewing each separately.


internal/reports/domain.go
```diff
@@ -355,9 +353,8 @@ func findInDomainReport(domain *limesresources.DomainReport, cluster *core.Clust
 	resource, exists := service.Resources[apiIdentity.ResourceName]
 	if !exists {
 		resInfo := cluster.InfoForResource(dbServiceType, dbResourceName)
-		resInfo.Name = apiIdentity.ResourceName
 		resource = &limesresources.DomainResourceReport{
-			ResourceInfo:     resInfo,
+			ResourceInfo:     behavior.BuildAPIResourceInfo(apiIdentity.ResourceName, resInfo),
 			CommitmentConfig: behavior.ToCommitmentConfig(now),
 		}
```
VoigtS, line 359: I'm not sure. Is it okay for us to have a naming duplication between the liquid structure (HasQuota) and the internal structure (NoQuota)?

VoigtS (changes requested):

internal/reports/domain.go
majewsky: The "internal structure" is specifically a data structure associated with the v1 API. In v2, this will not be a concern at all, because that one will be built around `liquid.ResourceInfo` from the get-go. So NoQuota will fall out of use at latest when the v1 API is eventually removed.

Core business logic already only sees `liquid.ResourceInfo` and thus uses HasQuota, whereas report generating code for the v1 API sometimes only has a `limesresources.ResourceInfo` because of how it is structured, and therefore it uses NoQuota. The latter is generated from the former and we use whatever we have available.

For what it's worth, since NoQuota is not reflected in the API, we could replace this field with a HasQuota field to match the liquid format. If that is desired, I can make a PR in go-api-declarations.

VoigtS: No, I think it is fine regarding the change from v1 to v2. Thanks for the clarification.

VoigtS (approved):

PR #525, opened 2024-08-12
Add GET endpoint for commitment conversion

VoigtS:
This draft aims to introduce the API endpoints needed to convert a commitment from a source to a target resource.
Doings:

- [x] introduce the GET endpoint to retreive the convertibles from the underlying configuration.
- [ ] Introduce the POST endpoint to transform a commitment to a target type
- [ ] Add documentation to the endpoints.


docs/users/api-spec-resources.md
```diff
@@ -588,6 +588,25 @@ After that, it returns the commitment as a JSON document.
 ### GET /v1/commitments/:token
 To ensure that a commitment can be checked for its `resource` type or `availability zone` before it gets transferred to a target project, this endpoint fetches the target commitment by its respective token.
 
+
+### POST /v1/commitments/can-convert
+Checks if a commitment could be converted to a resource of a different type.
```
majewsky, line 593: This expects a request body which is not documented, but should be.

But in the bigger picture, I don't like how this takes an entire CommitmentRequest and then uses only service type and resource name. I'd rather have only these in the request to make it clear what the scope of the answer is, e.g.

```
### GET /v1/commitment-conversions/:service_type/:resource_name
```

docs/users/api-spec-resources.md
```diff
@@ -588,6 +588,25 @@ After that, it returns the commitment as a JSON document.
...
+{
+  "conversions": [{
+    "from":            1,
+    "to":              2,
+    "target_resource": "serviceType/targetResource"
+  }]
+}
+```
+
+The `result` field indicates whether this commitment can be created without a `confirm_by` attribute, that is, confirmed immediately upon creation.
```
majewsky, line 607: This appears to be a copy-paste leftover. We should use this space to explain the example, e.g.

```
In this example, a commitment 1 unit of the original resource can be converted into a commitment for 2 units of the target resource.
```

internal/api/commitment.go
```diff
@@ -762,3 +763,52 @@ func (p *v1Provider) TransferCommitment(w http.ResponseWriter, r *http.Request)
...
+
+	var conversions []limesresources.CommitmentConversion
+	for _, targetBehavior := range p.Cluster.Config.ResourceBehaviors {
+		if targetBehavior.CommitmentConversion == (core.CommitmentConversion{}) {
+			continue
+		}
+		sourceName := string(req.ServiceType) + "/" + string(req.ResourceName)
+		if targetBehavior.FullResourceNameRx.MatchString(sourceName) {
+			continue
+		}
```
majewsky, line 788: I don't like that this bypasses the standard behavior-matching logic from core.Cluster. Please use the standard pattern to enumerate resources and have core.Cluster compute their behaviors, like in https://github.com/sapcc/limes/blob/b9096572776e3e81a535ec1eb615ea6e6cafa3a4/internal/reports/filter.go#L73-L79

internal/api/commitment.go
```diff
@@ -762,3 +763,52 @@ func (p *v1Provider) TransferCommitment(w http.ResponseWriter, r *http.Request)
...
+			To:             to,
+			TargetResource: liquid.ResourceName(targetBehavior.FullResourceNameRx),
+		}
+		conversions = append(conversions, result)
+	}
+
+	if len(conversions) == 0 {
+		http.Error(w, "no convertibles found", http.StatusUnprocessableEntity)
+		return
+	}
```
majewsky, line 804: Does that need to be a failure case? Returning an empty list sounds fine to me.

internal/api/commitment.go
```diff
@@ -762,3 +763,52 @@ func (p *v1Provider) TransferCommitment(w http.ResponseWriter, r *http.Request)
...
+		if targetBehavior.CommitmentConversion == (core.CommitmentConversion{}) {
+			continue
+		}
+		sourceName := string(req.ServiceType) + "/" + string(req.ResourceName)
+		if targetBehavior.FullResourceNameRx.MatchString(sourceName) {
+			continue
+		}
+		if sourceBehavior.CommitmentConversion.Identifier != targetBehavior.CommitmentConversion.Identifier {
+			continue
+		}
```
majewsky, line 791: Open question: I wonder if we should also check whether the units match between both resources here.

majewsky (changes requested):

docs/users/api-spec-resources.md
VoigtS: I was unsure about this too. I now included a own parse function that gets the behavior by the serviceType and resourceName directly.

docs/users/api-spec-resources.md
VoigtS: Yes, I did not notice the break within the paragraph. I wanted to keep the grammar similar to the `can-confirm` endpoint.

internal/api/commitment.go
VoigtS: It returns a empty list now.

internal/api/commitment.go
VoigtS: I implemented a check for it. Have a look.

internal/api/commitment.go
VoigtS: I agree. It felt a bit weird, handling those this way. I now implemented the way to properly retrieve the behavior from the method instead of directly looping over the behaviors.

internal/api/commitment.go
```diff
@@ -762,3 +763,77 @@ func (p *v1Provider) TransferCommitment(w http.ResponseWriter, r *http.Request)
...
+	token := p.CheckToken(r)
+	if !token.Require(w, "cluster:show_basic") {
+		return
+	}
+
+	sourceBehavior := p.parseAndValidateConversionRequest(w, r)
+	if sourceBehavior == nil {
+		return
+	}
+	sourceAPIIdentity := sourceBehavior.IdentityInV1API
```
majewsky, line 779: This is using `IdentityInV1API` in the wrong direction. Quota plugins need to be queried with the real names, not with the API identities.

Given that none of the other commitment endpoints support API-level resource renaming yet, please keep it like this for this PR. I will handle this with the next round of refactorings.

internal/api/commitment.go
```diff
@@ -762,3 +763,77 @@ func (p *v1Provider) TransferCommitment(w http.ResponseWriter, r *http.Request)
...
+
+	conversions := make([]limesresources.CommitmentConversion, 0)
+	for targetService, quotaPlugin := range p.Cluster.QuotaPlugins {
+		sourceResource := quotaPlugin.Resources()[sourceAPIIdentity.ResourceName]
```
majewsky, line 783: This uses the QuotaPlugin of the target resource to query for ResourceInfo of the source resource, which is not safe except when both resources are in the same service. This should be moved out of the loop and replaced with:

```go
sourceResInfo := p.Cluster.InfoForResource(sourceServiceType, sourceResourceName)
```

internal/api/commitment.go
```diff
@@ -762,3 +763,77 @@ func (p *v1Provider) TransferCommitment(w http.ResponseWriter, r *http.Request)
...
+				continue
+			}
+
+			from, to := p.getCommitmentConversionRate(sourceBehavior, &targetBehavior)
+			targetResourceName := limesresources.ResourceName(string(targetAPIIdentity.ServiceType) + "/" + string(targetAPIIdentity.ResourceName))
+			result := limesresources.CommitmentConversion{
+				From:           from,
+				To:             to,
+				TargetResource: targetResourceName,
+			}
```
majewsky, line 807: Squeezing service types into a `ResourceName` value is absolutely unacceptable. I missed this in the go-api-declarations PR. Followup to fix the API type: https://github.com/sapcc/go-api-declarations/pull/31

internal/api/commitment.go
```diff
@@ -762,3 +763,77 @@ func (p *v1Provider) TransferCommitment(w http.ResponseWriter, r *http.Request)
...
+			conversions = append(conversions, result)
+		}
+	}
+
+	respondwith.JSON(w, http.StatusOK, map[string]any{"conversions": conversions})
+}
+
+func (p *v1Provider) parseAndValidateConversionRequest(w http.ResponseWriter, r *http.Request) *core.ResourceBehavior {
+	serviceType := limes.ServiceType(mux.Vars(r)["service_type"])
+	resourceName := liquid.ResourceName(mux.Vars(r)["resource_name"])
```
majewsky, line 817: Although I just said that I'm going to fix this more broadly in the next refactoring, this should use the API-level type just to avoid future confusion:
```suggestion
	resourceName := limesresources.ResourceName(mux.Vars(r)["resource_name"])
```

internal/api/commitment_test.go
```diff
@@ -69,6 +69,33 @@ const testCommitmentsYAMLWithoutMinConfirmDate = `
 			commitment_is_az_aware: false
 		- resource: second/capacity
 			commitment_is_az_aware: true
+		- resource: second/capacity_portion
+			commitment_is_az_aware: true
+`
+
+const testConvertCommitmentsYAMl = `
```
majewsky, line 76: Weird casing.
```suggestion
const testConvertCommitmentsYAML = `
```

majewsky (changes requested):

docs/users/api-spec-resources.md
majewsky: I think this is an okay amount of code duplication for now.

internal/api/commitment.go
VoigtS: Thank you. Fixed it.

internal/api/commitment.go
VoigtS: I implemented the updated API type to the method

internal/api/commitment.go
VoigtS: Noted.

majewsky (approved):
To shortcut the rest of the review, I added some commits to address my remaining concerns. Please counter-review.

PR #547, opened 2024-09-05
add liquid-cinder and liquid-manila

majewsky:
For the most part, this is a straight port of the respective QuotaPlugin and CapacityPlugin implementations (so I recommend reading these side-by-side with the new code as necessary to follow along), but I used the opportunity to streamline some parts of the implementation. Most notably:

- The Cinder liquid discovers available volume types via the API instead of taking them as configuration.
- In the Manila liquid, I introduced new terminology to more clearly distinguish `VirtualShareType` (the share type names that appear in Limes resources) from `RealShareType` (the share type names that are used on the Manila API).


VoigtS (changes requested):
I looked at the liquids by comparing them to their plugin countertypes, this implementation looks good. Please have a look at the comment above.

VoigtS (approved):

PR #554, opened 2024-09-11
Add POST API to convert a commitment to a target Resource

VoigtS:
This PR aims to introduce the POST API to convert a supported commitment from a source to a target resource.

### DIFF
--- a/internal/api/commitment.go
+++ b/internal/api/commitment.go
@@ -904,7 +904,7 @@ func (p *v1Provider) ConvertCommitment(w http.ResponseWriter, r *http.Request) {
 		return
 	}
 	if dbCommitment.ConfirmBy != nil {
-		http.Error(w, "unable to convert unconfirmed commitments", http.StatusNotFound)
+		http.Error(w, "unable to convert unconfirmed commitments", http.StatusConflict)
 		return
 	}
 
--- a/internal/api/commitment.go
+++ b/internal/api/commitment.go
@@ -79,13 +79,25 @@ var (
 	`)
 
 	findCommitmentLocationByIDQuery = sqlext.SimplifyWhitespace(`
-	SELECT ps.type, pr.name
-	  FROM project_commitments pc
-	  JOIN project_az_resources par ON pc.az_resource_id = par.id
-	  JOIN project_resources pr ON par.resource_id = pr.id
-	  JOIN project_services ps ON pr.service_id = ps.id
-	 WHERE pc.id = $1
-`)
+		SELECT ps.type, pr.name
+		  FROM project_commitments pc
+		  JOIN project_az_resources par ON pc.az_resource_id = par.id
+		  JOIN project_resources pr ON par.resource_id = pr.id
+		  JOIN project_services ps ON pr.service_id = ps.id
+		 WHERE pc.id = $1
+	`)
+
+	findTargetAZResourceIDByCommitmentQuery = sqlext.SimplifyWhitespace(`
+		SELECT par.id
+		  FROM project_az_resources par
+		  JOIN project_resources pr ON par.resource_id = pr.id
+		  JOIN project_services ps ON pr.service_id = ps.id
+		 WHERE ps.type = $1 AND pr.name = $2 AND par.az = (
+		 	SELECT par.az
+			FROM project_commitments pc
+			JOIN project_az_resources par ON pc.az_resource_id = par.id
+			WHERE pc.id = $3)
+	`)
 
 	// NOTE: The third output column is `resourceAllowsCommitments`.
 	// We should be checking for `ResourceUsageReport.Forbidden == true`, but
@@ -933,11 +945,22 @@ func (p *v1Provider) ConvertCommitment(w http.ResponseWriter, r *http.Request) {
 		return
 	}
 
-	// TODO: finish the conversion part.
 	// conversion
+	if req.SourceAmount <= 0 || req.SourceAmount > dbCommitment.Amount {
+		http.Error(w, "unprocessable source_amount provided", http.StatusConflict)
+		return
+	}
 	fromAmount, toAmount := p.getCommitmentConversionRate(sourceBehavior, targetBehavior)
-	conversionAmount := (dbCommitment.Amount / fromAmount) * toAmount
-	remainderAmount := dbCommitment.Amount % fromAmount
+	conversionAmount := (req.SourceAmount / fromAmount) * toAmount
+	remainderAmount := req.SourceAmount % fromAmount
+
+	// TODO: add a check to dismiss requests with conversionAmount = 0
+
+	if conversionAmount != req.TargetAmount {
+		msg := fmt.Sprintf("conversion mismatch. calculated: %v, provided: %v", conversionAmount, req.TargetAmount)
+		http.Error(w, msg, http.StatusConflict)
+		return
+	}
 
 	tx, err := p.DB.Begin()
 	if respondwith.ErrorText(w, err) {
@@ -951,9 +974,19 @@ func (p *v1Provider) ConvertCommitment(w http.ResponseWriter, r *http.Request) {
 	// TODO: Required is the translated target AZResourceID.
 	// This would require the use of "findProjectAZResourceIDByLocationQuery" and therefore the domaind and project ID.
 	// The URL will become quite messy with this. Think about a solution.
+
+	var azResourceID db.ProjectAZResourceID
+	err = p.DB.SelectOne(&azResourceID, findTargetAZResourceIDByCommitmentQuery, targetServiceType, targetResourceName, dbCommitment.ID)
+	if errors.Is(err, sql.ErrNoRows) {
+		http.Error(w, "target resource not found", http.StatusNotFound)
+		return
+	} else if respondwith.ErrorText(w, err) {
+		return
+	}
+
 	now := p.timeNow()
 	convertedCommitment := db.ProjectCommitment{
-		AZResourceID: 0,
+		AZResourceID: azResourceID,
 		Amount:       conversionAmount,
 		Duration:     dbCommitment.Duration,
 		CreatedAt:    now,
@@ -979,7 +1012,7 @@ func (p *v1Provider) ConvertCommitment(w http.ResponseWriter, r *http.Request) {
 	}
 
 	if remainderAmount > 0 {
-		dbCommitment.Amount = remainderAmount
+		dbCommitment.Amount = dbCommitment.Amount - req.SourceAmount + remainderAmount
 		_, err = tx.Update(&dbCommitment)
 		if respondwith.ErrorText(w, err) {
 			return
--- a/internal/api/commitment.go
+++ b/internal/api/commitment.go
@@ -927,7 +927,7 @@ func (p *v1Provider) ConvertCommitment(w http.ResponseWriter, r *http.Request) {
 		return
 	}
 	if sourceBehavior.CommitmentConversion.Identifier == "" || sourceBehavior.CommitmentConversion.Identifier != targetBehavior.CommitmentConversion.Identifier {
-		msg := fmt.Sprintf("commitment is not convertible into target resource: %s", req.TargetResource)
+		msg := fmt.Sprintf("commitment is not convertible into resource %s/%s", req.TargetService, req.TargetResource)
 		http.Error(w, msg, http.StatusUnprocessableEntity)
 		return
 	}
--- a/internal/api/commitment.go
+++ b/internal/api/commitment.go
@@ -31,6 +31,7 @@ import (
 	"github.com/gorilla/mux"
 	"github.com/sapcc/go-api-declarations/limes"
 	limesresources "github.com/sapcc/go-api-declarations/limes/resources"
+	"github.com/sapcc/go-api-declarations/liquid"
 	"github.com/sapcc/go-bits/gopherpolicy"
 	"github.com/sapcc/go-bits/httpapi"
 	"github.com/sapcc/go-bits/must"
@@ -70,14 +71,21 @@ var (
 		  JOIN project_services ps ON pr.service_id = ps.id
 		 WHERE pc.id = $1 AND ps.project_id = $2
 	`)
+
 	findCommitmentByIDQuery = sqlext.SimplifyWhitespace(`
 		SELECT pc.*
 		  FROM project_commitments pc
-		  JOIN project_az_resources par ON pc.az_resource_id = par.id
-		  JOIN project_resources pr ON par.resource_id = pr.id
-		  JOIN project_services ps ON pr.service_id = ps.id
 		 WHERE pc.id = $1
 	`)
+
+	findCommitmentLocationByIDQuery = sqlext.SimplifyWhitespace(`
+	SELECT ps.type, pr.name
+	  FROM project_commitments pc
+	  JOIN project_az_resources par ON pc.az_resource_id = par.id
+	  JOIN project_resources pr ON par.resource_id = pr.id
+	  JOIN project_services ps ON pr.service_id = ps.id
+	 WHERE pc.id = $1
+`)
 
 	// NOTE: The third output column is `resourceAllowsCommitments`.
 	// We should be checking for `ResourceUsageReport.Forbidden == true`, but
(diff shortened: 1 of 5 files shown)


majewsky (changes requested):
First pass. I don't think I caught everything, but I have to leave soon, so I will leave it at that for now

internal/api/commitment.go
```diff
@@ -96,6 +96,11 @@ var (
 	findCommitmentByTransferToken = sqlext.SimplifyWhitespace(`
 		SELECT * FROM project_commitments WHERE transfer_token = $1
 	`)
+	findCommitmentByIDQuery = sqlext.SimplifyWhitespace(`
+        SELECT pc.*
+          FROM project_commitments pc
+         WHERE pc.id = $1
+    `)
```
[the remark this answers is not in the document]
VoigtS: The query is now inline.

internal/api/commitment.go
```diff
@@ -850,6 +861,193 @@ func (p *v1Provider) GetCommitmentConversions(w http.ResponseWriter, r *http.Req
 	respondwith.JSON(w, http.StatusOK, map[string]any{"conversions": conversions})
 }
 
+// ConvertCommitment handles POST /v1/domains/{domain_id}/projects/{project_id}/commitment-conversion/{commitment_id}
+func (p *v1Provider) ConvertCommitment(w http.ResponseWriter, r *http.Request) {
+	httpapi.IdentifyEndpoint(r, "/v1/domains/:domain_id/projects/:project_id/commitment-conversion/:commitment_id")
```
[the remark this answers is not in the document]
VoigtS: You are correct. The initial implementation did not take the conversion to a different project into account, leading to an unconsistent path in the end, this is now fixed.

internal/api/commitment.go
```diff
@@ -850,6 +861,193 @@ func (p *v1Provider) GetCommitmentConversions(w http.ResponseWriter, r *http.Req
...
+		msg := fmt.Sprintf("no such service and/or resource: %s/%s", req.TargetService, req.TargetResource)
+		http.Error(w, msg, http.StatusUnprocessableEntity)
+		return
+	}
+	targetBehavior := p.Cluster.BehaviorForResource(targetServiceType, targetResourceName)
+	if len(targetBehavior.CommitmentDurations) == 0 {
+		http.Error(w, "commitments are not enabled for this resource", http.StatusUnprocessableEntity)
+		return
+	}
+	if (sourceBehavior.CommitmentConversion == core.CommitmentConversion{} || targetBehavior.CommitmentConversion == core.CommitmentConversion{}) {
```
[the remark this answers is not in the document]
VoigtS: This one was a basd oversight, I also included a test that checks for a conversion to a invalid different identifier.

internal/api/commitment.go
```diff
@@ -850,6 +861,193 @@ func (p *v1Provider) GetCommitmentConversions(w http.ResponseWriter, r *http.Req
...
+		return
+	}
+	commitmentID := mux.Vars(r)["commitment_id"]
+	if commitmentID == "" {
+		http.Error(w, "no transfer token provided", http.StatusBadRequest)
+		return
+	}
+	dbDomain := p.FindDomainFromRequest(w, r)
+	if dbDomain == nil {
+		http.Error(w, "domain not found.", http.StatusNotFound)
```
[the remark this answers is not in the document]
VoigtS: I removed them also on the findProjectRequests. We are clean on this part now. Thanks for the info.

internal/api/commitment.go
```diff
@@ -850,6 +861,193 @@ func (p *v1Provider) GetCommitmentConversions(w http.ResponseWriter, r *http.Req
...
+		http.Error(w, msg, http.StatusConflict)
+		return
+	}
+	conversionAmount := (req.SourceAmount / fromAmount) * toAmount
+	remainderAmount := req.SourceAmount % fromAmount
+	if conversionAmount != req.TargetAmount {
+		msg := fmt.Sprintf("conversion mismatch. provided: %v, calculated: %v", req.TargetAmount, conversionAmount)
+		http.Error(w, msg, http.StatusConflict)
+		return
+	}
```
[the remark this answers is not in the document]
VoigtS: You are correct, the conversion rounding was later corrected with the addition of the remainder.
The implementation now rejects all requests where `req.SourceAmount % fromAmount > 0`.

internal/api/commitment.go
```diff
@@ -850,6 +861,193 @@ func (p *v1Provider) GetCommitmentConversions(w http.ResponseWriter, r *http.Req
...
+		if respondwith.ErrorText(w, err) {
+			return
+		}
+	}
+	if remainderAmount > 0 {
+		dbCommitment.Amount = dbCommitment.Amount - req.SourceAmount + remainderAmount
+		_, err = tx.Update(&dbCommitment)
+		if respondwith.ErrorText(w, err) {
+			return
+		}
```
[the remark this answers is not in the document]
VoigtS: The updated implementation now supersedes the old commitment.
It creates a `remainderCommitment` with Amount: `dbCommitment.Amount - req.SourceAmount`
and a `convertedCommitment` with Amount: `(req.SourceAmount / fromAmount) * toAmount`

internal/api/commitment.go
```diff
@@ -850,6 +858,180 @@ func (p *v1Provider) GetCommitmentConversions(w http.ResponseWriter, r *http.Req
...
+		http.Error(w, msg, http.StatusUnprocessableEntity)
+		return
+	}
+	if sourceBehavior.CommitmentConversion.Identifier == "" || sourceBehavior.CommitmentConversion.Identifier != targetBehavior.CommitmentConversion.Identifier {
+		http.Error(w, "commitment is not convertible", http.StatusUnprocessableEntity)
+		return
+	}
+
+	// conversion
+	if req.SourceAmount <= 0 || req.SourceAmount > dbCommitment.Amount {
```
majewsky, line 943: SourceAmount is unsigned:
```suggestion
	if req.SourceAmount > dbCommitment.Amount {
```

majewsky (commented):
I once again did not have enough time to complete the review today... Sorry.

internal/api/commitment.go
```diff
@@ -850,6 +858,180 @@ func (p *v1Provider) GetCommitmentConversions(w http.ResponseWriter, r *http.Req
...
+	} else if respondwith.ErrorText(w, err) {
+		return
+	}
+	sourceBehavior := p.Cluster.BehaviorForResource(sourceLoc.ServiceType, sourceLoc.ResourceName)
+	if len(sourceBehavior.CommitmentDurations) == 0 {
+		http.Error(w, "commitments are not enabled for this resource", http.StatusUnprocessableEntity)
+		return
+	}
+	if dbCommitment.ConfirmBy != nil {
+		http.Error(w, "unable to convert unconfirmed commitments", http.StatusConflict)
```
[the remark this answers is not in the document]
VoigtS: You are right, it is possible to do that. I implemented this now.
* I removed the early confirm_By check.
* Instead, the canConvertCommitment part only gets executed if the commitment has no confirm_by field.
* I added a unit test.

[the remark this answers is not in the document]
VoigtS: This also fixed a bug. Previously a commitment with the exact amount that it should be transferred as (in this case `amount: 3` would create a commitment with remainingAmount=0.
That's because the previous check was removed, but not reintigrated to check this case.

internal/api/commitment.go
```diff
@@ -850,6 +858,181 @@ func (p *v1Provider) GetCommitmentConversions(w http.ResponseWriter, r *http.Req
...
+		http.Error(w, "no route to this commitment", http.StatusNotFound)
+		return
+	} else if respondwith.ErrorText(w, err) {
+		return
+	}
+	sourceBehavior := p.Cluster.BehaviorForResource(sourceLoc.ServiceType, sourceLoc.ResourceName)
+	if len(sourceBehavior.CommitmentDurations) == 0 {
+		http.Error(w, "commitments are not enabled for this resource", http.StatusUnprocessableEntity)
+		return
+	}
```
majewsky, line 905: I think we can do without this. There should be a route to vacating a resource of commitments while still allowing maintenance on existing commitments.

internal/api/commitment.go
```diff
@@ -850,6 +858,181 @@ func (p *v1Provider) GetCommitmentConversions(w http.ResponseWriter, r *http.Req
...
+	targetBehavior := p.Cluster.BehaviorForResource(targetServiceType, targetResourceName)
+	if len(targetBehavior.CommitmentDurations) == 0 {
+		msg := fmt.Sprintf("commitments are not enabled for resource %s/%s", req.TargetService, req.TargetResource)
+		http.Error(w, msg, http.StatusUnprocessableEntity)
+		return
+	}
+	if sourceBehavior.CommitmentConversion.Identifier == "" || sourceBehavior.CommitmentConversion.Identifier != targetBehavior.CommitmentConversion.Identifier {
+		http.Error(w, "commitment is not convertible", http.StatusUnprocessableEntity)
```
majewsky, line 934:
```suggestion
		http.Error(w, "commitment is not convertible into $TARGET_RESOURCE", http.StatusUnprocessableEntity)
```

internal/api/commitment.go
```diff
@@ -850,6 +858,181 @@ func (p *v1Provider) GetCommitmentConversions(w http.ResponseWriter, r *http.Req
...
+	)
+	err = p.DB.QueryRow(findTargetAZResourceByTargetProjectQuery, targetProject.ID, targetServiceType, targetResourceName, sourceLoc.AvailabilityZone).
+		Scan(&targetResourceID, &targetAZResourceID)
+	if respondwith.ErrorText(w, err) {
+		return
+	}
+	if dbCommitment.AZResourceID == targetAZResourceID {
+		http.Error(w, "conversion attempt to the same resource.", http.StatusConflict)
+		return
+	}
```
majewsky, line 976: This is fine as defense in depth, but we ought to check `targetServiceType == sourceServiceType && targetResourceName == sourceResourceName` much earlier as a sanity check.

internal/api/commitment.go
```diff
@@ -850,6 +858,181 @@ func (p *v1Provider) GetCommitmentConversions(w http.ResponseWriter, r *http.Req
...
+	targetLoc := datamodel.AZResourceLocation{
+		ServiceType:      targetServiceType,
+		ResourceName:     targetResourceName,
+		AvailabilityZone: sourceLoc.AvailabilityZone,
+	}
+	// The commitment at the source resource was already confirmed and checked.
+	// Therefore only the addition to the target resource has to be checked against.
+	if dbCommitment.ConfirmBy == nil {
```
majewsky, line 984:
```suggestion
	if dbCommitment.ConfirmedAt != nil {
```

internal/api/commitment_test.go
```diff
@@ -1123,3 +1150,206 @@ func Test_GetCommitmentConversion(t *testing.T) {
...
+		},
+	}
+	commitmentWithConfirmBy := assert.JSONObject{
+		"commitment": assert.JSONObject{
+			"service_type":      "second",
+			"resource_name":     "capacity",
+			"availability_zone": "az-one",
+			"amount":            3,
+			"duration":          "1 hour",
+			"confirm_by":        s.Clock.Now().Add(14 * day).Unix(),
```
majewsky, line 1177: Needs to be updated to `confirmed_at`, see previous remark.

internal/api/commitment_test.go
```diff
@@ -1123,3 +1150,206 @@ func Test_GetCommitmentConversion(t *testing.T) {
...
+			"expires_at":        s.Clock.Now().Add(14*day + 1*time.Hour).Unix(),
+		}
+	}
+
+	// conversion rate is (second: 3 to first: 2)
+	assert.HTTPRequest{
+		Method:       http.MethodPost,
+		Path:         "/v1/domains/uuid-for-germany/projects/uuid-for-berlin/commitments/new",
+		Body:         commitment,
```
majewsky, line 1231: If `commitment` is only used once, I'd rather have it inline to reduce the need for scrolling while reading.

internal/api/commitment_test.go
```diff
@@ -1123,3 +1150,206 @@ func Test_GetCommitmentConversion(t *testing.T) {
...
+	if originalCommitment.State != db.CommitmentStateSuperseded {
+		t.Fatalf("commitment state should be %v. Received %v instead.", db.CommitmentStateSuperseded, originalCommitment.State)
+	}
+	err = s.DB.SelectOne(&originalCommitment, `SELECT * FROM project_commitments where ID = 2`)
+	if err != nil {
+		t.Fatal(err)
+	}
+	if originalCommitment.Amount != 18 {
+		t.Fatalf("commitment amount should be %v. Received %v instead.", 18, originalCommitment.Amount)
+	}
```
majewsky, line 1276: Can be written more compactly with `assert.DeepEqual(t, "commitment amount", originalCommitment.Amount, 18)`. For the other stuff analogously.

internal/api/commitment.go
```diff
@@ -850,6 +858,181 @@ func (p *v1Provider) GetCommitmentConversions(w http.ResponseWriter, r *http.Req
...
+		return
+	}
+	targetProject := p.FindProjectFromRequest(w, r, dbDomain)
+	if targetProject == nil {
+		return
+	}
+
+	// sourceBehavior
+	var dbCommitment db.ProjectCommitment
+	err := p.DB.SelectOne(&dbCommitment, `SELECT pc.* FROM project_commitments pc WHERE pc.id = $1`, commitmentID)
```
majewsky, line 884: We should not allow grabbing commitments from foreign projects:

```suggestion
	err := p.DB.SelectOne(&dbCommitment, findProjectCommitmentByIDQuery, commitmentID, targetProject.ID)
```

Then `targetProject` can be just `project`. The testcases consequently need to be adjusted to not convert between projects (and indeed, there should be a testcase that you cannot convert a foreign commitment without transferring it first).

internal/api/commitment_test.go
```diff
@@ -1123,3 +1150,206 @@ func Test_GetCommitmentConversion(t *testing.T) {
...
+	// Mismatching target amount
+	assert.HTTPRequest{
+		Method:       http.MethodPost,
+		Path:         "/v1/domains/uuid-for-germany/projects/uuid-for-berlin/commitments/4/convert",
+		Body:         req("first", "capacity", 6, 3),
+		ExpectBody:   assert.StringData("conversion mismatch. provided: 3, calculated: 4\n"),
+		ExpectStatus: http.StatusConflict,
+	}.Check(t, s.Handler)
+
+	// Check resulting empty commitment from conversion calculation (remainderAmount > 0)
```
majewsky, line 1322: Comment needs clarification.

majewsky (changes requested):
We're getting there.

internal/api/commitment.go
VoigtS: I implemented the query and adjusted the unit tests accordingly.

internal/api/commitment.go
VoigtS: I included the target resource into this error message.

internal/api/commitment_test.go
VoigtS: I changed it to the following, more meaningful comment:
`Reject an amount that doesn't fit into the conversion rate (remainder > 0).`

internal/api/commitment.go
VoigtS: I impelemented an early check now.
However, I keep this section in tact as a defense in depth as you said, because here we are able to explictly check against the `azResourceID`.

internal/api/commitment.go
VoigtS: You are correct, the commitments still exist and should be able to be operated on regardless if the resource itself accepts commitments. I noticed the `GET` method also checked against this case, I removed it there too.
That's because the conversion of a commitment requires a GET of the convertibles first if we are operating on UI level.

internal/api/commitment_test.go
VoigtS: I implemented this version now. However this is the only remaining part to adjust, because the conversion to a different project is now prohibited.

internal/api/commitment_test.go
VoigtS: The body's are now inline as requested.

majewsky (approved):
I added two commits. One for nitpicks, and one to extend the audit event format to include all participating commitments. Please merge if this is to your liking.

VoigtS:
Thank you, the nitpicks I missed, because I was focused on the ID checks wich made me miss some parts. The rework of the Auditevent seems fine to me as well.

PR #563, opened 2024-09-16
add core.RateBehavior

majewsky:
This is the same as core.ResourceBehavior, but for rates. I'm adding this now to support rate renaming in the same way as resource renaming (as a prerequisite for transitioning the Cronus plugin to LIQUID in the same way as for the resource-bearing plugins that were already converted).

There is one clearly ugly part where one specific error message in the PUT endpoint for rate limits now generates a 500 error in a specific case that should be a 4xx error. I don't care about fixing this in the v1 API because nobody is using this at the moment anyway. In the v2 API, we will transition PUT on rate limits to only work on one rate at the time in order to simplify the entire RateUpdater thing.

### DIFF
--- a/internal/core/name_mapping.go
+++ b/internal/core/name_mapping.go
@@ -31,9 +31,11 @@ import (
 // NameMapping contains an efficient pre-computed mapping between API-level and
 // DB-level service/resource/rate identifiers.
 type NameMapping struct {
-	cluster        *Cluster
-	resFromAPIToDB map[ResourceRef]dbResourceRef
-	resFromDBToAPI map[dbResourceRef]ResourceRef
+	cluster         *Cluster
+	resFromAPIToDB  map[ResourceRef]dbResourceRef
+	resFromDBToAPI  map[dbResourceRef]ResourceRef
+	rateFromAPIToDB map[RateRef]dbRateRef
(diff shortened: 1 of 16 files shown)


internal/core/name_mapping.go
```diff
@@ -56,6 +65,25 @@ func BuildNameMapping(cluster *Cluster) NameMapping {
 			nm.resFromDBToAPI[dbRef] = apiRef
 		}
 	}
+	for dbServiceType, quotaPlugin := range cluster.QuotaPlugins {
```
VoigtS, line 68: I'm not sure if I like this. We now have a combined mapping of `resources` and `rates` combined. If we only care about one of them, we carry the other automatically with us in the heap. This looks like a case to introduce an interface.

VoigtS (changes requested):

internal/core/name_mapping.go
majewsky: I will split this into `type ResourceNameMapping` and `type RateNameMapping`.

VoigtS (approved):

internal/core/name_mapping.go
VoigtS: Thank you, I like the split

PR #566, opened 2024-09-18
add support for rates in plugins/liquid, add liquid-cronus

majewsky:
I will not have time to test this until next week, but this ought to work, so please take a look.

For once, we have a new liquid that is an absolutely straight port of the old QuotaPlugin. No interesting changes at all.


VoigtS (approved):

internal/liquids/cronus/client.go
```diff
@@ -0,0 +1,78 @@
...
+package cronus
+
+import (
+	"context"
+	"net/http"
+
+	"github.com/gophercloud/gophercloud/v2"
+)
+
+// Client is a gophercloud.ServiceClient for the Cronus v1 API.
```
SuperSandro2000, line 29: Shouldn't this really go into https://github.com/sapcc/gophercloud-sapcc/ ?

majewsky: This only covers a tiny speck of API that only Limes is allowed to use, so there is no benefit to having it in a shared library.

PR #576, opened 2024-10-08
Add Duration Extension API

VoigtS:
This PR introduces an API to extend the duration of a commitment to a valid one provided in its configuration.
This is implemented as a draft, because I'm sure wheter to create a new commitment or overwrite the current one and add a new DB column with  an "Extended_at" field. I would like to hear your oppinion and will implement as desired.

TODO:

### DIFF
--- a/internal/api/commitment.go
+++ b/internal/api/commitment.go
@@ -1120,9 +1120,8 @@ func (p *v1Provider) ExtendCommitmentDuration(w http.ResponseWriter, r *http.Req
 	}
 	defer sqlext.RollbackUnlessCommitted(tx)
 
-	now := p.timeNow()
 	dbCommitment.Duration = req.Duration
-	dbCommitment.ExpiresAt = req.Duration.AddTo(unwrapOrDefault(dbCommitment.ConfirmBy, now))
+	dbCommitment.ExpiresAt = req.Duration.AddTo(unwrapOrDefault(dbCommitment.ConfirmBy, dbCommitment.CreatedAt))
 	_, err = tx.Update(&dbCommitment)
 	if respondwith.ErrorText(w, err) {
 		return
--- a/internal/api/commitment.go
+++ b/internal/api/commitment.go
@@ -1045,3 +1045,74 @@ func (p *v1Provider) getCommitmentConversionRate(source, target core.ResourceBeh
 	toAmount = source.CommitmentConversion.Weight / divisor
 	return fromAmount, toAmount
 }
+
+// ExtendCommitmentDuration handles POST /v1/domains/{domain_id}/projects/{project_id}/commitments/{commitment_id}/extend-duration
+func (p *v1Provider) ExtendCommitmentDuration(w http.ResponseWriter, r *http.Request) {
+	httpapi.IdentifyEndpoint(r, "/v1/domains/:domain_id/projects/:project_id/commitments/:commitment_id/extend-duration")
+	token := p.CheckToken(r)
+	if !token.Require(w, "project:edit") {
+		return
+	}
+	commitmentID := mux.Vars(r)["commitment_id"]
+	if commitmentID == "" {
+		http.Error(w, "no transfer token provided", http.StatusBadRequest)
+		return
+	}
+	dbDomain := p.FindDomainFromRequest(w, r)
+	if dbDomain == nil {
+		return
+	}
+	dbProject := p.FindProjectFromRequest(w, r, dbDomain)
+	if dbProject == nil {
+		return
+	}
+	var Request struct {
+		Duration limesresources.CommitmentDuration `json:"duration"`
+	}
+	req := Request
+	if !RequireJSON(w, r, &req) {
+		return
+	}
+
+	var dbCommitment db.ProjectCommitment
+	err := p.DB.SelectOne(&dbCommitment, findProjectCommitmentByIDQuery, commitmentID, dbProject.ID)
+	if errors.Is(err, sql.ErrNoRows) {
+		http.Error(w, "no such commitment", http.StatusNotFound)
+		return
+	} else if respondwith.ErrorText(w, err) {
+		return
+	}
+	var loc datamodel.AZResourceLocation
+	err = p.DB.QueryRow(findProjectAZResourceLocationByIDQuery, dbCommitment.AZResourceID).
+		Scan(&loc.ServiceType, &loc.ResourceName, &loc.AvailabilityZone)
+	if errors.Is(err, sql.ErrNoRows) {
+		// defense in depth: this should not happen because all the relevant tables are connected by FK constraints
+		http.Error(w, "no route to this commitment", http.StatusNotFound)
+		return
+	} else if respondwith.ErrorText(w, err) {
+		return
+	}
+	behavior := p.Cluster.BehaviorForResource(loc.ServiceType, loc.ResourceName)
+	validDurations := behavior.CommitmentDurations
+
+	existsInDurations := false
+	//allowsUpConversion := false
+	for _, duration := range validDurations {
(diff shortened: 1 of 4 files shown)


internal/api/commitment.go
```diff
@@ -1045,3 +1045,100 @@ func (p *v1Provider) getCommitmentConversionRate(source, target core.ResourceBeh
...
+		msg := fmt.Sprintf("provided duration: %s does not match the config %v", req.Duration, validDurations)
+		http.Error(w, msg, http.StatusUnprocessableEntity)
+		return
+	}
+	allowsDurationExtension := req.Duration.CompareTo(dbCommitment.Duration)
+	if allowsDurationExtension != 1 {
+		msg := fmt.Sprintf("provided duration: %s cannot be extended with config %v", req.Duration, validDurations)
+		http.Error(w, msg, http.StatusUnprocessableEntity)
+		return
+	}
```
majewsky, line 1115: I would rather have this check on the real expiry time, i.e. (pseudo-code)

```go
newExpiresAt := formulaAsBelow()
if dbCommitment.ExpiresAt.Before(newExpiresAt) {
  fail()
}
```

Right now, this works out to be functionally the same, but in case we have to add more exceptions to how expiry is calculated, this feels more robust against loopholes.

internal/api/commitment.go
```diff
@@ -1045,3 +1045,100 @@ func (p *v1Provider) getCommitmentConversionRate(source, target core.ResourceBeh
 	toAmount = source.CommitmentConversion.Weight / divisor
 	return fromAmount, toAmount
 }
+
+// ExtendCommitmentDuration handles POST /v1/domains/{domain_id}/projects/{project_id}/commitments/{commitment_id}/extend-duration
+func (p *v1Provider) ExtendCommitmentDuration(w http.ResponseWriter, r *http.Request) {
+	httpapi.IdentifyEndpoint(r, "/v1/domains/:domain_id/projects/:project_id/commitments/:commitment_id/extend-duration")
```
majewsky, line 1051: I would rather like the API endpoint to be `update-duration` instead of `extend-duration`. This would leave the possibility to adapt this endpoint to allow duration shortening by cloud admins later.

majewsky (changes requested):

internal/api/commitment.go
VoigtS: I agree with this. The logic needs to be vice versa. `newExpiresAt.Before(dbCommitment.ExpiresAt)`
Otherwise we would only allow down conversions. 

This implementation now also accepts a conversion to the same Duration, because we don't check for equal. This is probably fine, because why would we error this case out.

internal/api/commitment.go
```diff
@@ -1045,3 +1045,100 @@ func (p *v1Provider) getCommitmentConversionRate(source, target core.ResourceBeh
 	toAmount = source.CommitmentConversion.Weight / divisor
 	return fromAmount, toAmount
 }
+
+// ExtendCommitmentDuration handles POST /v1/domains/{domain_id}/projects/{project_id}/commitments/{commitment_id}/extend-duration
+func (p *v1Provider) UpdateCommitmentDuration(w http.ResponseWriter, r *http.Request) {
+	httpapi.IdentifyEndpoint(r, "/v1/domains/:domain_id/projects/:project_id/commitments/:commitment_id/extend-duration")
```
majewsky, line 1051: Some more leftover mentions of `extend-duration` instead of `update-duration`.

internal/api/commitment.go
```diff
@@ -1045,3 +1045,100 @@ func (p *v1Provider) getCommitmentConversionRate(source, target core.ResourceBeh
...
+	behavior := p.Cluster.BehaviorForResource(loc.ServiceType, loc.ResourceName)
+	validDurations := behavior.CommitmentDurations
+
+	existsInDurations := false
+	for _, duration := range validDurations {
+		if duration.String() == req.Duration.String() {
+			existsInDurations = true
+		}
+	}
+	if !existsInDurations {
```
majewsky, line 1104: There is no relevant ambiguity in how commitment durations are parsed and serialized, so we can use the same check as parseAndValidateCommitmentRequest here:

```suggestion
	if !slices.Contains(behavior.CommitmentDurations, req.Duration) {
```

internal/api/commitment.go
```diff
@@ -1045,3 +1045,100 @@ func (p *v1Provider) getCommitmentConversionRate(source, target core.ResourceBeh
...
+	dbCommitment.Duration = req.Duration
+	dbCommitment.ExpiresAt = newExpiresAt
+	_, err = tx.Update(&dbCommitment)
+	if respondwith.ErrorText(w, err) {
+		return
+	}
+	err = tx.Commit()
+	if respondwith.ErrorText(w, err) {
+		return
+	}
```
majewsky, line 1132: Since only one update statement is issued, the explicit transaction is superfluous and can be omitted.

internal/api/commitment_test.go
```diff
@@ -1325,3 +1325,109 @@ func Test_ConvertCommitments(t *testing.T) {
...
+				"availability_zone": "az-one",
+				"amount":            10,
+				"duration":          "1 hours",
+			},
+		},
+		ExpectStatus: http.StatusCreated,
+	}.Check(t, s.Handler)
+
+	// Fast forward by 1 hour. Creation_time = 0; Now = 1; (Expire = Creation_time + 2 hours)
+	s.Clock.StepBy(1 * time.Hour)
```
majewsky, line 1353: The assertion in the comment (`expire = creation_time + 2 hours`) does not hold because the duration has not been extended yet.

In fact, this is a testcase that should fail in my opinion, since `expires_at <= NOW()` in the next request. The implementation should add a check to reject extensions after `expires_at <= NOW()`, and also for `state != 'active'` (esp. to reject editing superseded commitments).

internal/api/commitment_test.go
```diff
@@ -1325,3 +1325,109 @@ func Test_ConvertCommitments(t *testing.T) {
...
+		ExpectBody:   assert.StringData("provided duration: 3 hours does not match the config [1 hour 2 hours]\n"),
+		ExpectStatus: http.StatusUnprocessableEntity,
+	}.Check(t, s.Handler)
+
+	// Negative: Provided Date < Commitment Duration
+	assert.HTTPRequest{
+		Method:       http.MethodPost,
+		Path:         "/v1/domains/uuid-for-germany/projects/uuid-for-berlin/commitments/1/update-duration",
+		Body:         assert.JSONObject{"duration": "1 hour"},
+		ExpectBody:   assert.StringData("provided duration: 1 hour cannot be extended with config [1 hour 2 hours]\n"),
```
majewsky, line 1430: This error message sounds odd, because it does not explain that we are trying to shorten the duration. I would like to see something like:
```suggestion
		ExpectBody:   assert.StringData("duration change from 2 hours to 1 hour forbidden\n"),
```

majewsky (changes requested):

internal/api/commitment_test.go
VoigtS: You are correct. I did not properly thought this case through.
We check the `expired_date` against `now` first.

However, I explicitly check against the state `superseded` now instead of active, because doing otherwise prohibits the operation on planned/pending commitments, wich was a requirment (the calculation also takes `confirm_by` into account)

If a different handling is desired, please voice it.

internal/api/commitment.go
VoigtS: The update statement is now performed without an explicit transaction.

internal/api/commitment.go
VoigtS: The leftovers are cleared up now.

internal/api/commitment_test.go
VoigtS: The response now matches your suggestion.

PR #579, opened 2024-10-09
plugins: simplify Nova flavor listing

majewsky:
In Nova microversion 2.61, flavors.ListDetail() includes the extra specs, so we can save a ton of work querying them one flavor at a time. Gophercloud already has support for this, too.

**Draft:** I have not tested this yet. Feel free to have a look at the implementation; I will test this in QA after my leave.


VoigtS (approved):
I like beeing able to reduce on additional API calls. This looks good.

PR #585, opened 2024-10-21
fix enumeration of more than 1000 Nova hypervisors

majewsky:
Pagination was introduced in microversion 2.33, so our microversion upgrade in 2fbae7df400a9c5c09d5db0c07d5d7b99eabf0e5 (PR #579) broke this accidentally. This was not caught in QA because the QA regions are not big enough to reach a 4-digit hypervisor count.


internal/plugins/nova/gophercloud_fixes.go
```diff
@@ -0,0 +1,72 @@
...
+)
+
+// Like hypervisors.HypervisorPage, but fixes bug <https://github.com/gophercloud/gophercloud/issues/3222> by being a LinkedPageBase instead of a SinglePageBase.
+type hypervisorPage struct {
+	pagination.LinkedPageBase
+}
+
+// IsEmpty is required for LinkedPageBase to work.
+func (page hypervisorPage) IsEmpty() (bool, error) {
+	if page.StatusCode == 204 {
```
SuperSandro2000, line 35:
```suggestion
	if page.StatusCode == http.StatusNoContent {
```

VoigtS (approved):

PR #593, opened 2024-11-04
add liquid-ironic

majewsky:
This introduces a differentiation between Nova and Ironic as services (from the perspective of Limes). Even though virtualized and baremetal instances are both managed by Nova, the capacity measurement strategies are entirely different, and the quotas are separate from each other (even though they're in the same quota set), so this split is warranted in my opinion.

The capacity measurement uses a new strategy for associating baremetal nodes with availability zones. My research into the OpenStack documentation and codebase suggests that this approach works in the general case, so it's a nice opportunity to get rid of the weird and brittle string matching that the old CapacityPlugin is doing.

Finally, this also gets rid of the weird flavor aliasing feature (represented by type nova.FlavorTranslationTable) because it never ended up being used. If flavor renaming is to happen in the future, we will probably do it through API-level resource renaming instead.

## Testing done

Because of the substantial implementation changes, I have tested this more than usual. I ran the liquid against all prod clusters and scraped reports for all projects with baremetal instances from both Limes and the liquid. I then verified that both sides agree on the usage numbers as well as on the reported AZ associations in the subresources in all projects.


VoigtS (approved):
I compared it to the previous plugin implementation, the AZ assignment would be a lot easier with this. Let's try it out.

PR #594, opened 2024-11-05
Add api endpoint to render request payloads for querying the LIQUID api

Varsius:
Checklist:

### DIFF
--- a/internal/api/liquid.go
+++ b/internal/api/liquid.go
@@ -1,0 +1,121 @@
+/******************************************************************************
+*
+*  Copyright 2024 SAP SE
+*
+*  Licensed under the Apache License, Version 2.0 (the "License");
+*  you may not use this file except in compliance with the License.
+*  You may obtain a copy of the License at
+*
+*      http://www.apache.org/licenses/LICENSE-2.0
+*
+*  Unless required by applicable law or agreed to in writing, software
+*  distributed under the License is distributed on an "AS IS" BASIS,
+*  WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+*  See the License for the specific language governing permissions and
+*  limitations under the License.
+*
+******************************************************************************/
+
+package api
+
+import (
+	"database/sql"
+	"errors"
+	"net/http"
+
+	"github.com/sapcc/go-bits/httpapi"
+	"github.com/sapcc/go-bits/respondwith"
+
+	"github.com/sapcc/limes/internal/core"
+	"github.com/sapcc/limes/internal/datamodel"
+	"github.com/sapcc/limes/internal/db"
+)
+
+// GetServiceCapacityRequest handles GET /admin/liquid/service-capacity-request?service_type=:type.
+func (p *v1Provider) GetServiceCapacityRequest(w http.ResponseWriter, r *http.Request) {
+	httpapi.IdentifyEndpoint(r, "/admin/liquid/service-capacity-request")
+	token := p.CheckToken(r)
+	if !token.Require(w, "cluster:show") {
+		return
+	}
+
+	serviceType := r.URL.Query().Get("service_type")
+	if serviceType == "" {
+		http.Error(w, "missing required parameter: service_type", http.StatusBadRequest)
+		return
+	}
+
+	plugin, ok := p.Cluster.CapacityPlugins[serviceType]
+	if !ok {
+		http.Error(w, "invalid service type", http.StatusBadRequest)
+		return
+	}
+
+	backchannel := datamodel.NewCapacityPluginBackchannel(p.Cluster, p.DB)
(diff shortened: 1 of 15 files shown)


internal/test/plugins/capacity_static.go
```diff
@@ -111,6 +111,24 @@ func (p *StaticCapacityPlugin) Scrape(ctx context.Context, _ core.CapacityPlugin
 	return result, serializedMetrics, nil
 }
 
+func (p *StaticCapacityPlugin) BuildServiceCapacityRequest(backchannel core.CapacityPluginBackchannel, allAZs []limes.AvailabilityZone) (*liquid.ServiceCapacityRequest, error) {
+	return &liquid.ServiceCapacityRequest{
+		AllAZs: []liquid.AvailabilityZone{"az-one", "az-two"},
+		DemandByResource: map[liquid.ResourceName]liquid.ResourceDemand{
+			"first/capacity": {
```
majewsky, line 118: `first/capacity` is a pair of service type and resource name, but only a resource name should be used as a key here.

This is running up against a difference between how the old CapacityPlugins operate vs. how LIQUID operates. Capacity plugins are not restricted to a single service and can report resources from arbitrary services (hence their `Scrape` returns a data structure that is keyed on both service type and resource name).

There is no good way to solve this right now, which gives us the moral justification to just use a kludge and announce some stuff that is unrelated to the capacity plugin's actual configuration:

```suggestion
			"capacity": {
```

internal/test/plugins/quota_generic.go
```diff
@@ -136,6 +136,20 @@ func (p *GenericQuotaPlugin) ScrapeRates(ctx context.Context, project core.Keyst
...
+	return &liquid.ServiceUsageRequest{
+		AllAZs: []liquid.AvailabilityZone{"az-one", "az-two"},
+		ProjectMetadata: &liquid.ProjectMetadata{
+			UUID: "uuid-for-berlin",
+			Name: "berlin",
+			Domain: liquid.DomainMetadata{
+				UUID: "uuid-for-germany",
+				Name: "germany",
+			},
+		},
```
majewsky, line 149: This is kind of the opposite case. We don't need to make up any data if we get good data in the arguments:
```suggestion
		ProjectMetadata: project.ForLiquid(),
```

internal/core/plugin.go
```diff
@@ -132,6 +132,11 @@ type QuotaPlugin interface {
 	// The `serializedMetrics` return value is persisted in the Limes DB and
 	// supplied to all subsequent RenderMetrics calls.
 	Scrape(ctx context.Context, project KeystoneProject, allAZs []limes.AvailabilityZone) (result map[liquid.ResourceName]ResourceData, serializedMetrics []byte, err error)
+
+	// BuildServiceUsageRequest generates the request body payload for querying
+	// the LIQUID API endpoint /v1/projects/:uuid/report-usage
+	BuildServiceUsageRequest(project KeystoneProject, allAZs []limes.AvailabilityZone) (*liquid.ServiceUsageRequest, error)
```
majewsky, line 138: This is going to be a bit nitpicky. But since it is not a valid response to return a nil result, I'd like to see the pointer type removed:

```suggestion
	BuildServiceUsageRequest(project KeystoneProject, allAZs []limes.AvailabilityZone) (liquid.ServiceUsageRequest, error)
```

Same for BuildServiceCapacityRequest.

Historically, I'm guilty of overusing pointer types in return signatures, mostly because it's more convenient to write `return nil, err` instead of `return SomeReallyLongTypeName{}, err`. But I've been trying to cut back on this over time. Pointer types ought to be reserved for their two major uses that they express intent of: modifying a passed argument or `self`, and indicating that a value can be unpopulated (similar to `Option<>` in other languages).

majewsky (changes requested):
This is looking very good already, but I have some concerns about the interface declaration and the mocks.

internal/api/liquid.go
```diff
@@ -0,0 +1,123 @@
...
+
+	plugin, ok := p.Cluster.CapacityPlugins[serviceType]
+	if !ok {
+		http.Error(w, "invalid service type", http.StatusBadRequest)
+		return
+	}
+
+	backchannel := datamodel.NewCapacityPluginBackchannel(p.Cluster, p.DB)
+	serviceCapacityRequest, err := plugin.BuildServiceCapacityRequest(backchannel, p.Cluster.Config.AvailabilityZones)
+	if errors.Is(err, errNotALiquid) {
```
majewsky, line 58: This will not work as you expect because `errors.Is()` checks for the specific identity of `errNotALiquid`, not just its error message (think memory address instead of memory contents). For this to work, `errNotALiquid` needs to be visible to the plugin implementations and also used by them.

majewsky (approved):
Thanks!

PR #598, opened 2024-11-11
replace gophercloud/utils with go-bits/gophercloudext

majewsky:
As expected, this removes several transitive dependencies, including `gopkg.in/yaml.v3`.


internal/collector/capacity_scrape_test.go
```diff
@@ -199,7 +199,7 @@ func Test_ScanCapacity(t *testing.T) {
 
 	// next scan should throw out the crap records and recreate the deleted ones;
 	// also change the reported Capacity to see if updates are getting through
-	s.Cluster.CapacityPlugins["unittest"].(*plugins.StaticCapacityPlugin).Capacity = 23
+	s.Cluster.CapacityPlugins["unittest"].(*plugins.StaticCapacityPlugin).Capacity = 23 //nolint:errcheck
```
SuperSandro2000, line 202:
```suggestion
	s.Cluster.CapacityPlugins["unittest"].(*plugins.StaticCapacityPlugin).Capacity = 23 //nolint:errcheck // in context, this type cast cannot fail
```
probably?

majewsky: As discussed yesterday, this one (like the vast majority of lints introduced in the new errcheck version) occur in test files. I don't really care to document each one specifically, since it's always the same pattern of casting from an interface to the concrete mock that is instantiated further up in the test.

majewsky: Resolved by https://github.com/sapcc/go-makefile-maker/pull/211

PR #600, opened 2024-11-13
Add max quota configuration for project level

VoigtS:
We would like to introduce the `max quota` configuration not only on domain level, but on project level as well. Because this requirment is different from the domain level one, a new field gets introduced to the same API endpoint which triggers if a user accesses it with project level privileges.

~~I am unsure about the naming of the db columns. I kept the orignal domain column name for now. I'm open to rename it as well if it is required.~~ The columns are now renamed as suggested.

The API provides a `max_quota` field which now returns the minimum value of the `outside_admin` or `local_admin` column. If a  speperate field is required, I can adjust this implementation.

Before this gets published, the Oslo policies need to be updated, because the `project:max_quota_edit` role was originally used for the domain level. The `domain:max_quota_edit` role needs to be added as well.


internal/db/models.go
```diff
@@ -123,6 +123,7 @@ type ProjectResource struct {
 	MinQuotaFromBackend     *uint64             `db:"min_quota_from_backend"`
 	MaxQuotaFromBackend     *uint64             `db:"max_quota_from_backend"`
 	MaxQuotaFromAdmin       *uint64             `db:"max_quota_from_admin"`
+	MaxQuotaFromProject     *uint64             `db:"max_quota_from_project"`
```
majewsky, line 126: It's becoming evident to me that the original `max_quota_from_admin` naming was flawed and should be fixed (Postgres has `ALTER TABLE ... RENAME` for this purpose). "From admin" vs. "from project" is not a good contrast because there also exists admins on the project level (but "from admin" refers to admins on the domain level). I suggest:
```suggestion
	MaxQuotaFromOutsideAdmin *uint64             `db:"max_quota_from_outside_admin"`
	MaxQuotaFromLocalAdmin   *uint64             `db:"max_quota_from_local_admin"`
```

VoigtS: I updated the naming scheme accordingly

internal/api/projects.go
```diff
@@ -184,7 +184,10 @@ func (p *v1Provider) PutProjectMaxQuota(w http.ResponseWriter, r *http.Request)
 	httpapi.IdentifyEndpoint(r, "/v1/domains/:id/projects/:id/max-quota")
 	requestTime := p.timeNow()
 	token := p.CheckToken(r)
-	if !token.Require(w, "project:edit_max_quota") {
+	projectAccess := token.Check("project:edit_max_quota")
+	domainAccess := token.Check("domain:edit_max_quota")
+	if !projectAccess && !domainAccess {
+		http.Error(w, "Forbidden", http.StatusForbidden)
```
majewsky, line 190: I have two notes here.

1. The pattern for policy rule names is that the first part identifies the object whose attributes are being changed. Since these endpoints update project quota constraints, all policies shall be `project:...` instead of `domain:...`. (Compare to the state before, where `project:edit_max_quota` was only accessible to domain admins.)
2. For editing `max_quota_from_local_admin`, I would like to avoid policy rule sprawl and reuse `project:edit` (which used to refer to editing of quotas, and currently refers to editing of commitments).

All in all, while also noting that domain admins are also project admins by inheritance:

```suggestion
	if !token.Require("project:edit") {
		return
	}
	domainAccess := token.Check("project:edit_as_outside_admin")
```

majewsky (changes requested):

majewsky (approved):

PR #604, opened 2024-11-19
remove support for Ironic in QuotaPlugins/CapacityPlugins

majewsky:
It is now only supported to use liquid-ironic for Ironic resources.

- The FlavorAliases/FlavorTranslationTable feature is removed as it was a) never used and b) only intended for Ironic flavors.

- The HypervisorTypeRules feature is removed because we only used this in practice to distinguish between hypervisor type `none` (for baremetal) or `vmware`. With Ironic removed, this can be replaced by a fixed string.

- The FlavorNameSelection feature is reduced to a single regex which works out the same in practice. It is not necessary to have multiple regexes here anymore: This was used for category assignment of resources, but this has moved into ResourceBehavior now.


majewsky:
@Varsius FYI - Please use this as a guide for what needs to be carried over into liquid-nova.

VoigtS (approved):

PR #606, opened 2024-11-19
generalize DistributeDemandFairly()

majewsky:
I need to use this function in liquid-cinder as well.

The new implementation is the same as the old one, except that the last paragraph is rewritten to remove references to Manila-specific identifiers.


VoigtS (approved):
Check the lints, otherwise this is fine.

PR #611, opened 2024-11-25
Add liquid nova report usage endpoint

Varsius:
Checklist:

### DIFF
--- a/internal/liquids/nova/liquid.go
+++ b/internal/liquids/nova/liquid.go
@@ -22,16 +22,32 @@ package nova
 import (
 	"context"
 	"errors"
+	"fmt"
+	"regexp"
+	"slices"
+	"strings"
 	"time"
 
 	"github.com/gophercloud/gophercloud/v2"
 	"github.com/gophercloud/gophercloud/v2/openstack"
 	"github.com/gophercloud/gophercloud/v2/openstack/compute/v2/flavors"
+	"github.com/gophercloud/gophercloud/v2/openstack/compute/v2/quotasets"
 	"github.com/sapcc/go-api-declarations/liquid"
+
+	"github.com/sapcc/limes/internal/liquids"
 )
 
 type Logic struct {
-	NovaV2 *gophercloud.ServiceClient `json:"-"`
+	// configuration
+	WithSubresources bool `json:"with_subresources"`
+	// connections
+	NovaV2            *gophercloud.ServiceClient `json:"-"`
+	OSTypeProber      *OSTypeProber              `json:"-"`
+	ServerGroupProber *ServerGroupProber         `json:"-"`
+	// computed state
+	ignoredFlavorNames liquids.State[[]string]                                `json:"-"`
+	hasPooledResource  liquids.State[map[string]map[liquid.ResourceName]bool] `json:"-"`
+	hwVersionResources liquids.State[[]liquid.ResourceName]                   `json:"-"`
 }
 
 // Init implements the liquidapi.Logic interface.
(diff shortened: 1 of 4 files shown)


internal/liquids/nova/liquid.go
```diff
@@ -41,8 +53,64 @@ func (l *Logic) Init(ctx context.Context, provider *gophercloud.ProviderClient,
...
+		return fmt.Errorf("while enumerating default quotas: %w", err)
+	}
+	l.hasPooledResource = make(map[string]map[liquid.ResourceName]bool)
+	hwVersionResourceRx := regexp.MustCompile(`^hw_version_(\S+)_(cores|instances|ram)$`)
+	for resourceName := range defaultQuotaClassSet {
+		match := hwVersionResourceRx.FindStringSubmatch(resourceName)
+		if match == nil {
+			continue
+		}
+		hwVersion, baseResourceName := match[1], liquid.ResourceName(match[2])
```
majewsky, line 83: These folks also need to appear in `BuildServiceInfo` with the same naming pattern, see existing implementation in the Nova quota plugin.

There is unfortunately not a good way to test this currently. The Nova folks have not deployed support for this anywhere yet.

internal/liquids/nova/liquid.go
```diff
@@ -93,6 +161,18 @@ func (l *Logic) BuildServiceInfo(ctx context.Context) (liquid.ServiceInfo, error
 	return liquid.ServiceInfo{
 		Version:   time.Now().Unix(),
 		Resources: resources,
+		UsageMetricFamilies: map[liquid.MetricName]liquid.MetricFamilyInfo{
+			"liquid_nova_instance_counts_by_hypervisor": {
+				Type:      liquid.MetricTypeCounter,                                 // TODO: Counter or Gauge?
+				Help:      "Total number of instances, grouped by hypervisor type.", // TODO: Is this correct? Liquid nova only has one hypervisor type if I understand correctly
+				LabelKeys: []string{"hypervisor_type"},
+			},
+			"liquid_nova_instance_counts_bvy_hypervisor_and_az": {
```
majewsky, line 170:
```suggestion
			"liquid_nova_instance_counts_by_hypervisor_and_az": {
```

internal/liquids/nova/liquid.go
```diff
@@ -93,6 +161,18 @@ func (l *Logic) BuildServiceInfo(ctx context.Context) (liquid.ServiceInfo, error
 	return liquid.ServiceInfo{
 		Version:   time.Now().Unix(),
 		Resources: resources,
+		UsageMetricFamilies: map[liquid.MetricName]liquid.MetricFamilyInfo{
+			"liquid_nova_instance_counts_by_hypervisor": {
+				Type:      liquid.MetricTypeCounter,                                 // TODO: Counter or Gauge?
```
majewsky, line 166: Gauge. Counter is for monotonically increasing values.

internal/liquids/nova/liquid.go
```diff
@@ -22,16 +22,30 @@ package nova
...
+	// configuration
+	WithSubresources bool `json:"with_subresources"`
+	// connections
+	NovaV2            *gophercloud.ServiceClient `json:"-"`
+	OSTypeProber      *OSTypeProber              `json:"-"`
+	ServerGroupProber *ServerGroupProber         `json:"-"`
+	// computed state
+	ignoredFlavorNames []string                                `json:"-"`
+	hasPooledResource  map[string]map[liquid.ResourceName]bool `json:"-"`
+	hwVersionResources []liquid.ResourceName                   `json:"-"`
```
majewsky, line 48: The part in `Init()` that computes this should rather be in `BuildServiceInfo()`, in order to allow liquid-nova to adapt to changing configuration (esp. flavors) on the fly. Because `BuildServiceInfo()` can run in parallel with other logic methods, this requires wrapping the data in `liquids.State[...]` to avoid data races: https://pkg.go.dev/github.com/sapcc/limes/internal/liquids#State - Other liquids already use this, so grep for `liquids.State` to see usage examples.

internal/liquids/nova/subresources.go
```diff
@@ -0,0 +1,135 @@
...
+
+type novaServerListOpts struct {
+	AllTenants bool   `q:"all_tenants"`
+	TenantID   string `q:"tenant_id"`
+}
+
+func (opts novaServerListOpts) ToServerListQuery() (string, error) {
+	q, err := gophercloud.BuildQueryString(opts)
+	return q.String(), err
+}
```
majewsky, line 135: This was added back in the old plugin because the upstream type did not contain all the attributes that we need. But now, https://pkg.go.dev/github.com/gophercloud/gophercloud/v2/openstack/compute/v2/servers#ListOpts has both `all_tenants` and `tenant_id`, so this type can be removed and replaced by `servers.ListOpts`.

internal/liquids/nova/usage.go
```diff
@@ -0,0 +1,203 @@
...
+		if _, exists := resources[instanceResourceName]; !exists {
+			// otherwise used the appropriate pooled instance resource
+			isPooled = true
+			instanceResourceName = l.pooledResourceName(attrs.HWVersion, "instances")
+		}
+
+		// count subresource towards "instances" (or separate instance resource)
+		resources[instanceResourceName].AddLocalizedUsage(az, 1)
+		if l.WithSubresources {
+			azData := UsageInAZ(resources[instanceResourceName], az)
```
majewsky, line 185: It demonstrates good instinct that you're being defensive about whether the AZResourceReport exists or not. But in this case, `AddLocalizedUsage()` already guarantees it, so this can be greatly simplified to:
```suggestion
			azData := resources[instanceResourceName].PerAZ[az] // and drop the UsageInAZ function
```

internal/liquids/nova/usage.go
```diff
@@ -0,0 +1,203 @@
...
+			Quota: &absoluteLimits.MaxTotalInstances,
+			PerAZ: liquid.AZResourceUsageReport{Usage: absoluteLimits.TotalInstancesUsed}.PrepareForBreakdownInto(req.AllAZs),
+		},
+		"ram": {
+			Quota: &absoluteLimits.MaxTotalRAMSize,
+			PerAZ: liquid.AZResourceUsageReport{Usage: absoluteLimits.TotalRAMUsed}.PrepareForBreakdownInto(req.AllAZs),
+		},
+		"server_groups": {
+			Quota: &absoluteLimits.MaxServerGroups,
+			PerAZ: map[liquid.AvailabilityZone]*liquid.AZResourceUsageReport{liquid.AvailabilityZoneAny: {Usage: absoluteLimits.TotalServerGroupsUsed}},
```
majewsky, line 115: The liquid library has a convenience constructor for this:
```suggestion
			PerAZ: liquid.InAnyAZ(liquid.AZResourceUsageReport{Usage: absoluteLimits.TotalServerGroupsUsed}),
```
Same below for `server_group_members`.

internal/liquids/nova/liquid.go
```diff
@@ -101,12 +196,24 @@ func (l *Logic) ScanCapacity(ctx context.Context, req liquid.ServiceCapacityRequ
...
 func (l *Logic) SetQuota(ctx context.Context, projectUUID string, req liquid.ServiceQuotaRequest, serviceInfo liquid.ServiceInfo) error {
-	return errors.New("TODO")
+	opts := make(novaQuotaUpdateOpts, len(serviceInfo.Resources))
+	for resName := range serviceInfo.Resources {
+		opts[string(resName)] = req.Resources[resName].Quota
+	}
+	return quotasets.Update(ctx, l.NovaV2, projectUUID, opts).Err
+}
+
+func (l *Logic) IgnoreFlavor(flavorName string) bool {
```
majewsky, line 208: Present tense in a method name implies an activity, but this method does not cause anyone to ignore anything; it just checks whether something is already ignored. In this case, a past tense feels more appropriate to me:
```suggestion
func (l *Logic) IsIgnoredFlavor(flavorName string) bool {
```

internal/liquids/nova/usage.go
```diff
@@ -0,0 +1,203 @@
...
+		return liquid.ServiceUsageReport{}, fmt.Errorf("while collecting instance data: %w", err)
+	}
+
+	for _, subres := range allSubresources {
+		var attrs SubresourceAttributes
+		if err = json.Unmarshal(subres.Attributes, &attrs); err != nil { // TODO: Unmarshalling here + marshalling when building the subresources seems inefficient
+			return liquid.ServiceUsageReport{}, err
+		}
+
+		az := attrs.AZ
```
majewsky, line 167: This value should be checked against `allAZs` and otherwise be coerced into `liquid.AvailabilityZoneUnknown`, either here or in `buildInstanceSubresource`. Otherwise, Limes will complain if the report references AZs not known to it.

In fact, this pattern is a thing that keeps reappearing in most liquids with AZ awareness, so I opened a PR to add it to the library: https://github.com/sapcc/go-api-declarations/pull/40

majewsky (approved):

PR #614, opened 2024-11-29
Add support for AZ-separated quotas

VoigtS:
This implementation handles the introduction of resource topologies. This allows the handling of synchronizing AZ-aware and AZ-separated quotas with the backend. 
The plugins will check if a liquid provides a valid toplogy. Invalid AZ's that are detected during scrapes will be rejected. Scraped AZ-separated quota will be included into the `backend_quota` column that resides in the `project_az_resources` table. From limes quota will then be synced with the backend if necessary. The `backend_quota` values will then be updated to the calculated `quota` values.

Things to note:
~* The AZ key sorting in the uitlity functions might be able to be improved.~
~* `MatchLiquidReportToTopology` might collect the errors for all resources instead of failing at the first mismatch.~
~* SetQuota required a new type to support passing through the AZ-aware qutoas to the backend. We might have to adjust the naming of the structure and its contents.~

TODOS:
- [x] Adjust the algorithm to not provide base quota in any for az-separated resources. (Requires additional agreement on how to approach this topic.) -> The base quota value will now be provided to all AZs if the topology is AZSeparated.
- [x] Cleanup steps involving `ResourceBehavior.CommitmentIsAZAware`

### DIFF
--- a/internal/api/commitment.go
+++ b/internal/api/commitment.go
@@ -240,14 +240,14 @@ func (p *v1Provider) parseAndValidateCommitmentRequest(w http.ResponseWriter, r 
 		http.Error(w, "commitments are not enabled for this resource", http.StatusUnprocessableEntity)
 		return nil, nil, nil
 	}
-	if resInfo.Topology != liquid.FlatResourceTopology {
-		if !slices.Contains(p.Cluster.Config.AvailabilityZones, req.AvailabilityZone) {
-			http.Error(w, "no such availability zone", http.StatusUnprocessableEntity)
+	if resInfo.Topology == liquid.FlatResourceTopology {
+		if req.AvailabilityZone != limes.AvailabilityZoneAny {
+			http.Error(w, `resource does not accept AZ-aware commitments, so the AZ must be set to "any"`, http.StatusUnprocessableEntity)
 			return nil, nil, nil
 		}
 	} else {
-		if req.AvailabilityZone != limes.AvailabilityZoneAny {
-			http.Error(w, `resource does not accept AZ-aware commitments, so the AZ must be set to "any"`, http.StatusUnprocessableEntity)
+		if !slices.Contains(p.Cluster.Config.AvailabilityZones, req.AvailabilityZone) {
+			http.Error(w, "no such availability zone", http.StatusUnprocessableEntity)
 			return nil, nil, nil
 		}
 	}
--- a/internal/api/commitment.go
+++ b/internal/api/commitment.go
@@ -32,6 +32,7 @@ import (
 	"github.com/sapcc/go-api-declarations/cadf"
 	"github.com/sapcc/go-api-declarations/limes"
 	limesresources "github.com/sapcc/go-api-declarations/limes/resources"
+	"github.com/sapcc/go-api-declarations/liquid"
 	"github.com/sapcc/go-bits/audittools"
 	"github.com/sapcc/go-bits/gopherpolicy"
 	"github.com/sapcc/go-bits/httpapi"
@@ -234,11 +235,12 @@ func (p *v1Provider) parseAndValidateCommitmentRequest(w http.ResponseWriter, r 
 		return nil, nil, nil
 	}
 	behavior := p.Cluster.BehaviorForResource(dbServiceType, dbResourceName)
+	resInfo := p.Cluster.InfoForResource(dbServiceType, dbResourceName)
 	if len(behavior.CommitmentDurations) == 0 {
 		http.Error(w, "commitments are not enabled for this resource", http.StatusUnprocessableEntity)
 		return nil, nil, nil
 	}
-	if behavior.CommitmentIsAZAware {
+	if resInfo.Topology != liquid.FlatResourceTopology {
 		if !slices.Contains(p.Cluster.Config.AvailabilityZones, req.AvailabilityZone) {
 			http.Error(w, "no such availability zone", http.StatusUnprocessableEntity)
 			return nil, nil, nil
--- a/internal/api/commitment_test.go
+++ b/internal/api/commitment_test.go
@@ -49,8 +49,6 @@ const testCommitmentsYAML = `
 		- resource: first/.*
 			commitment_durations: ["1 hour", "2 hours"]
 			commitment_min_confirm_date: '1970-01-08T00:00:00Z' # one week after start of mock.Clock
-		- resource: first/things
-		- resource: first/capacity
 `
 const testCommitmentsYAMLWithoutMinConfirmDate = `
 	availability_zones: [ az-one, az-two ]
--- a/internal/api/commitment_test.go
+++ b/internal/api/commitment_test.go
@@ -25,10 +25,12 @@ import (
 	"testing"
 	"time"
 
+	"github.com/sapcc/go-api-declarations/liquid"
 	"github.com/sapcc/go-bits/assert"
 
 	"github.com/sapcc/limes/internal/db"
 	"github.com/sapcc/limes/internal/test"
+	"github.com/sapcc/limes/internal/test/plugins"
 )
 
 const day = 24 * time.Hour
@@ -48,9 +50,7 @@ const testCommitmentsYAML = `
 			commitment_durations: ["1 hour", "2 hours"]
 			commitment_min_confirm_date: '1970-01-08T00:00:00Z' # one week after start of mock.Clock
 		- resource: first/things
-			commitment_is_az_aware: false
 		- resource: first/capacity
-			commitment_is_az_aware: true
 `
 const testCommitmentsYAMLWithoutMinConfirmDate = `
 	availability_zones: [ az-one, az-two ]
@@ -65,12 +65,6 @@ const testCommitmentsYAMLWithoutMinConfirmDate = `
 		# the resources in "first" have commitments, the ones in "second" do not
 		- resource: second/.*
 			commitment_durations: ["1 hour", "2 hours", "3 hours"]
-		- resource: second/things
-			commitment_is_az_aware: false
-		- resource: second/capacity
-			commitment_is_az_aware: true
-		- resource: second/capacity_portion
-			commitment_is_az_aware: true
 `
 
 const testConvertCommitmentsYAML = `
@@ -95,10 +89,8 @@ const testConvertCommitmentsYAML = `
 		- resource: third/.*
 			commitment_durations: ["1 hour", "2 hours"]
 		- resource: first/capacity
-			commitment_is_az_aware: true
 			commitment_conversion: {identifier: flavor1, weight: 48}
 		- resource: second/capacity
-			commitment_is_az_aware: true
 			commitment_conversion: {identifier: flavor1, weight: 32}
 		- resource: third/capacity_c32
 			commitment_conversion: {identifier: flavor1, weight: 32}
@@ -118,6 +110,10 @@ func TestCommitmentLifecycleWithDelayedConfirmation(t *testing.T) {
 		test.WithConfig(testCommitmentsYAML),
 		test.WithAPIHandler(NewV1API),
 	)
+	plugin := s.Cluster.QuotaPlugins["first"].(*plugins.GenericQuotaPlugin)
+	plugin2 := s.Cluster.QuotaPlugins["second"].(*plugins.GenericQuotaPlugin)
+	plugin.LiquidServiceInfo.Resources = map[liquid.ResourceName]liquid.ResourceInfo{"capacity": {Topology: liquid.AZAwareResourceTopology}, "things": {Topology: liquid.FlatResourceTopology}}
+	plugin2.LiquidServiceInfo.Resources = map[liquid.ResourceName]liquid.ResourceInfo{"capacity": {Topology: liquid.AZAwareResourceTopology}, "things": {Topology: liquid.FlatResourceTopology}}
 
 	// GET returns an empty list if there are no commitments
 	assert.HTTPRequest{
@@ -476,6 +472,9 @@ func TestPutCommitmentErrorCases(t *testing.T) {
 		test.WithConfig(testCommitmentsYAML),
 		test.WithAPIHandler(NewV1API),
 	)
+
+	plugin := s.Cluster.QuotaPlugins["first"].(*plugins.GenericQuotaPlugin)
+	plugin.LiquidServiceInfo.Resources = map[liquid.ResourceName]liquid.ResourceInfo{"things": {Topology: liquid.FlatResourceTopology}}
 
 	request := assert.JSONObject{
 		"service_type":      "first",
--- a/internal/collector/scrape.go
+++ b/internal/collector/scrape.go
@@ -303,6 +303,12 @@ func (c *Collector) writeResourceScrapeResult(dbDomain db.Domain, dbProject db.P
 				azRes.Usage = data.Usage
 				azRes.PhysicalUsage = data.PhysicalUsage
 
+				// set AZ backend quota
+				resInfo := c.Cluster.InfoForResource(srv.Type, res.Name)
+				if resInfo.HasQuota && resInfo.Topology == liquid.AZSeparatedResourceTopology {
+					azRes.BackendQuota = &data.Quota
+				}
+
 				// warn when the backend is inconsistent with itself
 				if data.Subresources != nil && uint64(len(data.Subresources)) != data.Usage {
 					logg.Info("resource quantity mismatch in project %s, resource %s/%s, AZ %s: usage = %d, but found %d subresources",
--- a/internal/collector/scrape.go
+++ b/internal/collector/scrape.go
@@ -312,9 +312,7 @@ func (c *Collector) writeResourceScrapeResult(dbDomain db.Domain, dbProject db.P
 				resInfo := c.Cluster.InfoForResource(srv.Type, res.Name)
 				if resInfo.Topology == liquid.AZSeparatedResourceTopology && resInfo.HasQuota {
 					azRes.BackendQuota = data.Quota
-				}
-				// reset backendQuota entries for topology changes
-				if resInfo.Topology != liquid.AZSeparatedResourceTopology {
+				} else {
 					azRes.BackendQuota = nil
 				}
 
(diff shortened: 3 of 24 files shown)


internal/collector/sync_quota_to_backend.go
```diff
@@ -93,15 +93,30 @@ func (c *Collector) processQuotaSyncTask(ctx context.Context, srv db.ProjectServ
...
+	azQuotaSyncMarkResourcesAsAppliedQuery = sqlext.SimplifyWhitespace(`
+		WITH resourceIDs as (
+		  SELECT id
+			FROM project_resources
+		 WHERE service_id = $1
+		)
+		UPDATE project_az_resources
+		   SET backend_quota = quota
+	 	 WHERE resource_id in (SELECT id from resourceIDs)
+	`)
```
majewsky, line 119: This will also fill `backend_quota` for resources with AZAwareResourceTopology, which is not what we want. You're not seeing this behavior in your test because you fill `quota` manually for AZSeparatedResourceTopology resources only. But in real life, ACPQ will fill `quota` for AZAwareResourceTopology resources as well.

internal/datamodel/apply_computed_project_quota.go
```diff
@@ -374,7 +375,14 @@ func acpqComputeQuotas(stats map[limes.AvailabilityZone]clusterAZAllocationStats
 				}
 			}
 			if sumOfLocalizedQuotas < cfg.ProjectBaseQuota {
-				target[limes.AvailabilityZoneAny][resourceID].Desired = cfg.ProjectBaseQuota - sumOfLocalizedQuotas
+				// AZ separated toplogy receives the basequota to all available AZs
+				if resInfo.Topology == liquid.AZSeparatedResourceTopology {
+					for az := range isRelevantAZ {
+						target[az][resourceID].Desired = cfg.ProjectBaseQuota - target[az][resourceID].Allocated
```
majewsky, line 381: `Desired` is not computed on top of `Allocated`:
```suggestion
						target[az][resourceID].Desired = cfg.ProjectBaseQuota
```

majewsky (changes requested):

internal/api/commitment.go
```diff
@@ -234,11 +235,12 @@ func (p *v1Provider) parseAndValidateCommitmentRequest(w http.ResponseWriter, r
 		return nil, nil, nil
 	}
 	behavior := p.Cluster.BehaviorForResource(dbServiceType, dbResourceName)
+	resInfo := p.Cluster.InfoForResource(dbServiceType, dbResourceName)
 	if len(behavior.CommitmentDurations) == 0 {
 		http.Error(w, "commitments are not enabled for this resource", http.StatusUnprocessableEntity)
 		return nil, nil, nil
 	}
-	if behavior.CommitmentIsAZAware {
+	if resInfo.Topology != liquid.FlatResourceTopology {
```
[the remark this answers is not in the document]
VoigtS: I reversed the statements as described.

internal/api/commitment_test.go
```diff
@@ -48,9 +50,7 @@ const testCommitmentsYAML = `
 			commitment_durations: ["1 hour", "2 hours"]
 			commitment_min_confirm_date: '1970-01-08T00:00:00Z' # one week after start of mock.Clock
 		- resource: first/things
-			commitment_is_az_aware: false
 		- resource: first/capacity
-			commitment_is_az_aware: true
```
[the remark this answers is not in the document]
VoigtS: You are right. I removed them at the other parts, those were left over.

internal/plugins/utils.go
```diff
@@ -19,6 +19,68 @@
...
+	for key := range mapToSort {
+		sortedKeys = append(sortedKeys, string(key))
+	}
+	sort.Strings(sortedKeys)
+
+	return sortedKeys
+}
+
+func CheckResourceTopologies(serviceInfo liquid.ServiceInfo) (err error) {
+	errs := []error{}
```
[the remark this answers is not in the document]
VoigtS: I included this fix while fixing the error collecting in the same codeblock.

internal/plugins/capacity_liquid.go
```diff
@@ -90,6 +94,19 @@ func (p *liquidCapacityPlugin) Scrape(ctx context.Context, backchannel core.Capa
...
+		errs := []error{}
+		perAZ := resp.Resources[liquid.ResourceName(resourceName)].PerAZ
+		toplogy := p.LiquidServiceInfo.Resources[liquid.ResourceName(resourceName)].Topology
+		err := MatchLiquidReportToTopology(perAZ, toplogy)
+		if err != nil {
+			errs = append(errs, fmt.Errorf("service: %s, resource: %s: %w", p.ServiceType, resourceName, err))
+		}
+		if len(errs) > 0 {
+			return nil, nil, errors.Join(errs...)
+		}
```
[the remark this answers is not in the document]
VoigtS: This is now fixed. It was implemented correctly in the test file. The transfer to the liquid clients was messed up.

internal/plugins/liquid.go
```diff
@@ -119,9 +123,22 @@ func (p *liquidQuotaPlugin) Scrape(ctx context.Context, project core.KeystonePro
...
+		perAZ := resp.Resources[liquid.ResourceName(resourceName)].PerAZ
+		toplogy := p.LiquidServiceInfo.Resources[liquid.ResourceName(resourceName)].Topology
+		err := MatchLiquidReportToTopology(perAZ, toplogy)
+		if err != nil {
+			errs = append(errs, fmt.Errorf("service: %s, resource: %s: %w", p.ServiceType, resourceName, err))
+		}
+		if len(errs) > 0 {
+			return nil, nil, errors.Join(errs...)
+		}
+	}
```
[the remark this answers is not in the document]
VoigtS: This is now fixed. It was implemented correctly in the test file. The transfer to the liquid clients was messed up.

internal/plugins/utils.go
```diff
@@ -19,6 +19,68 @@
...
+
+func SortMapKeys[M map[K]V, K ~string, V any](mapToSort M) []string {
+	var sortedKeys []string
+	for key := range mapToSort {
+		sortedKeys = append(sortedKeys, string(key))
+	}
+	sort.Strings(sortedKeys)
+
+	return sortedKeys
+}
```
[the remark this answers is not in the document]
VoigtS: This is now fixed. It required a slice.Collect in order to sort, because maps.Keys returns an iterator. The type casts are now removed as well.

internal/plugins/utils.go
```diff
@@ -19,6 +19,68 @@
...
+
+	resourceNames := SortMapKeys(resources)
+	for _, resourceName := range resourceNames {
+		topology := resources[liquid.ResourceName(resourceName)].Topology
+		if topology == "" {
+			continue
+		}
+		if !topology.IsValid() {
+			errs = append(errs, fmt.Errorf("invalid toplogy: %s on resource: %s", topology, resourceName))
+		}
```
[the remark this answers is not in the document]
VoigtS: I fixed all occurences that showed up.

internal/core/plugin.go
```diff
@@ -181,6 +181,11 @@ type QuotaPlugin interface {
 	CollectMetrics(ch chan<- prometheus.Metric, project KeystoneProject, serializedMetrics []byte) error
 }
 
+type Quotas struct {
+	QuotaForResource uint64
+	QuotasForAZs     map[liquid.AvailabilityZone]liquid.AZResourceQuotaRequest
+}
+
```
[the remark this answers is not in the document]
VoigtS: I changed the datatype accordingly.

internal/collector/scrape.go
```diff
@@ -208,11 +208,16 @@ func (c *Collector) writeResourceScrapeResult(dbDomain db.Domain, dbProject db.P
...
 			resData.UsageData = core.InAnyAZ(core.UsageData{Usage: 0})
 			resourceData[resName] = resData
 		} else {
+			// AZ separated resources will not include "any" AZ. The basequota will be distributed towards the existing AZs.
+			if resInfo.Topology == liquid.AZSeparatedResourceTopology {
+				continue
+			}
 			// for AZ-aware resources, ensure that we also have a ProjectAZResource in
 			// "any", because ApplyComputedProjectQuota needs somewhere to write base
 			// quotas into if enabled
```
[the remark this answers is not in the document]
VoigtS: This is now implemented. I also added a testcase for it by deleting `az-two` from the resource `things`.

internal/collector/sync_quota_to_backend.go
VoigtS: You are right. By using the service ID I was disregarding the toplogy of the services underlying toplogy. I changed it so the quota snyc collects the affected resource IDs and only operates on them.

internal/collector/sync_quota_to_backend.go
```diff
@@ -123,36 +134,70 @@ func (c *Collector) performQuotaSync(ctx context.Context, srv db.ProjectService,
...
+			err := rows.Scan(&availabilityZone, &currentAZQuota, &targetAZQuota)
+			if err != nil {
+				return err
+			}
+			// defense in depth: configured backend_quota for AZ any or unknown are not valid for the azSeparatedQuota topology.
+			if (availabilityZone == liquid.AvailabilityZoneAny || availabilityZone == liquid.AvailabilityZoneUnknown) && currentAZQuota != nil {
+				return fmt.Errorf("detected invalid AZ: %s for resource: %s with topology: %s has backend_quota: %v", availabilityZone, resourceName, resInfo.Topology, currentAZQuota)
+			}
+			azSeparatedResouceIDs = append(azSeparatedResouceIDs, resourceID)
+			targetAZQuotasInDB[resourceName] = make(map[liquid.AvailabilityZone]liquid.AZResourceQuotaRequest)
```
majewsky, line 176: This should not overwrite existing values from other AZs:
```suggestion
			if targetAZQuotasInDB[resourceName] == nil {
				targetAZQuotasInDB[resourceName] = make(map[liquid.AvailabilityZone]liquid.AZResourceQuotaRequest)
			}
```

internal/datamodel/apply_computed_project_quota_test.go
```diff
@@ -547,9 +610,19 @@ func withCommitted(committed uint64, stats projectAZAllocationStats) projectAZAl
...
+func expectACPQResultForAZSeparated(t *testing.T, input map[limes.AvailabilityZone]clusterAZAllocationStats, cfg core.AutogrowQuotaDistributionConfiguration, constraints map[db.ProjectResourceID]projectLocalQuotaConstraints, expected acpqGlobalTarget) {
+	t.Helper()
+	actual, _ := acpqComputeQuotas(input, cfg, constraints, liquid.ResourceInfo{Topology: liquid.AZSeparatedResourceTopology})
+	createACPQResult(t, input, cfg, actual, expected)
+}
+
 func expectACPQResult(t *testing.T, input map[limes.AvailabilityZone]clusterAZAllocationStats, cfg core.AutogrowQuotaDistributionConfiguration, constraints map[db.ProjectResourceID]projectLocalQuotaConstraints, expected acpqGlobalTarget) {
 	t.Helper()
-	actual, _ := acpqComputeQuotas(input, cfg, constraints)
+	actual, _ := acpqComputeQuotas(input, cfg, constraints, liquid.ResourceInfo{})
```
majewsky, line 621: We're testing with empty-string as Topology here, which is not supposed to be a supported value. `expectACPQResult` should instead take a `liquid.ResourceInfo` argument, and all existing calls should be updated to take either Flat or AZAware as appropriate.

internal/plugins/capacity_liquid.go
```diff
@@ -90,6 +94,19 @@ func (p *liquidCapacityPlugin) Scrape(ctx context.Context, backchannel core.Capa
...
 			p.LiquidServiceType, p.LiquidServiceInfo.Version, resp.InfoVersion)
 	}
+	resourceNames := SortMapKeys(p.LiquidServiceInfo.Resources)
+	var errs []error
+	for _, resourceName := range resourceNames {
+		perAZ := resp.Resources[resourceName].PerAZ
+		topology := p.LiquidServiceInfo.Resources[resourceName].Topology
+		err := MatchLiquidReportToTopology(perAZ, topology)
+		if err != nil {
+			errs = append(errs, fmt.Errorf("service: %s, resource: %s: %w", p.ServiceType, resourceName, err))
```
majewsky, line 104: This does not need to specify the service type, since the caller already adds that to the error message:
```suggestion
			errs = append(errs, fmt.Errorf("resource: %s: %w", resourceName, err))
```
Same in the QuotaPlugin below.

internal/plugins/liquid.go
```diff
@@ -174,12 +194,12 @@ func castSliceToAny[T any](input []T) (output []any) {
...
+func (p *liquidQuotaPlugin) SetQuota(ctx context.Context, project core.KeystoneProject, quotaReq map[liquid.ResourceName]liquid.ResourceQuotaRequest) error {
 	req := liquid.ServiceQuotaRequest{
-		Resources: make(map[liquid.ResourceName]liquid.ResourceQuotaRequest, len(quotas)),
+		Resources: make(map[liquid.ResourceName]liquid.ResourceQuotaRequest, len(quotaReq)),
 	}
-	for resName, quota := range quotas {
-		req.Resources[resName] = liquid.ResourceQuotaRequest{Quota: quota}
+	for resName, request := range quotaReq {
+		req.Resources[resName] = liquid.ResourceQuotaRequest{Quota: request.Quota, PerAZ: request.PerAZ}
 	}
```
majewsky, line 203: It doesn't let me do a suggestion here for some reason, but the type already nearly matches, so this can be shortened to `req := liquid.ServiceQuotaRequest{Resources: quotaReq}`.

internal/plugins/utils.go
```diff
@@ -19,6 +19,71 @@
...
+
+	"github.com/sapcc/go-api-declarations/liquid"
+	"github.com/sapcc/go-bits/logg"
+)
+
 func p2u64(val uint64) *uint64 {
 	return &val
 }
+
+func SortMapKeys[M map[K]V, K ~string, V any](mapToSort M) []K {
```
majewsky, line 36: I forgot to nitpick this before because of the larger remark on this function, but `Sort` implies that this function operates in-place, which it does not:
```suggestion
func SortedMapKeys[M map[K]V, K ~string, V any](mapToSort M) []K {
```

majewsky (changes requested):
We're getting there. A few more nitpicks, and here's a big thing on top: I don't know how to flag this on a specific line, so I'm going to write it here. If the code works the same as in my head, we will have a neverending series of useless quota syncs on resources with AZSeparatedResourceTopology. As per the spec, these resources should have `ResourceUsageReport.Quota == nil` (i.e. no total quota should be reported), which means that `project_resources.backend_quota` will be empty. ACPQ currently always fills `project_resources.quota` for the resources that it operates on, so on the resource level, we always have `quota != backend_quota` and therefore a quota sync is forced during each ACPQ run.

My gut feeling is that ACPQ should just not fill `project_resources.quota` for AZSeparateResourceTopology resources. (The absence of the overall quota could then also be a trigger for the UI to hide the region-wide quota/usage bar.) But I have not finished thinking about whether that will induce corner cases elsewhere.

internal/plugins/liquid.go
VoigtS: You are correct. I changed the value assignment as suggested.

internal/plugins/utils.go
VoigtS: I included this change in a separate commit, because I changed all referenced to the function with it.

internal/datamodel/apply_computed_project_quota_test.go
VoigtS: I included this suggestion. Each test case now provides a specific topology.

internal/plugins/capacity_liquid.go
VoigtS: I made a separate commit for it where I adjusted both.

VoigtS:
> We're getting there. A few more nitpicks, and here's a big thing on top: I don't know how to flag this on a specific line, so I'm going to write it here. If the code works the same as in my head, we will have a neverending series of useless quota syncs on resources with AZSeparatedResourceTopology. As per the spec, these resources should have `ResourceUsageReport.Quota == nil` (i.e. no total quota should be reported), which means that `project_resources.backend_quota` will be empty. ACPQ currently always fills `project_resources.quota` for the resources that it operates on, so on the resource level, we always have `quota != backend_quota` and therefore a quota sync is forced during each ACPQ run.
> 
> My gut feeling is that ACPQ should just not fill `project_resources.quota` for AZSeparateResourceTopology resources. (The absence of the overall quota could then also be a trigger for the UI to hide the region-wide quota/usage bar.) But I have not finished thinking about whether that will induce corner cases elsewhere.

I did not know that it is intended for the liquid to return a nil value if it has the AZ separated toplogy. I included a skip for setting the ACPQ quota and provided an explanation for it.

internal/datamodel/apply_computed_project_quota.go
```diff
@@ -188,6 +189,11 @@ func ApplyComputedProjectQuota(serviceType db.ServiceType, resourceName liquid.R
 	}
 	servicesWithUpdatedQuota := make(map[db.ProjectServiceID]struct{})
 	err = sqlext.WithPreparedStatement(tx, acpqUpdateProjectQuotaQuery, func(stmt *sql.Stmt) error {
+		// Skip resources with AZSeparated toplogy. The quota scrape would receive a resource nil value, while ACPQ calculates qouta.
+		// This would lead to unnecessary quota syncs with the backend, because backendQuota != quota.
+		if resInfo.Topology == liquid.AZSeparatedResourceTopology {
+			return nil
+		}
```
majewsky, line 196: Because of this, `servicesWithUpdatedQuota` will never be filled for AZ-separated resources, which means that quota syncs will never be triggered from ACPQ for those resources. For AZ-separated resources, `servicesWithUpdatedQuota` needs to be filled during `acpqUpdateAZQuotaQuery` instead.

majewsky (changes requested):
I have pushed a few nitpick fixes. Also, there is one last bigger thing:

internal/datamodel/apply_computed_project_quota.go
VoigtS: I included a fix in PR: [717a028](https://github.com/sapcc/limes/pull/614/commits/717a02832ecaa78cc2e5616a460c3be27b84c13b) `servicesWithUpdatedQuota` will be declared before `acpqUpdateAZQuotaQuery` takes place and will populate the fields with the required service IDs.

majewsky (approved):
Awesome, thanks!

PR #628, opened 2024-12-13
Liquid nova report capacity

Varsius:
Checklist:


internal/liquids/nova/hypervisor_subcapacity.go
```diff
@@ -105,3 +163,42 @@ func (b *SplitFlavorSubcapacityBuilder) AddHypervisor(h MatchingHypervisor) {
...
+	attrs := SubcapacityAttributes{
+		AggregateName: h.AggregateName,
+		Traits:        h.Traits,
+	}
+	buf, err := json.Marshal(attrs)
+	if err != nil {
+		return fmt.Errorf("while serializing Subcapacity Attributes: %w", err)
+	}
+	b.Subcapacities = append(b.Subcapacities, liquid.Subcapacity{
+		// TODO: What should be the naming here? We list subcapacities for cores, ram and instances on the same split flavor capacity
```
majewsky, line 184: I don't like there being multiple subcapacities for different aspects of the same thing. Let's have this as a single subcapacity with `Capacity: 0,` and all the actual data in attributes shaped like the old subcapacities.

internal/liquids/nova/usage.go
```diff
@@ -0,0 +1,203 @@
...
+	// so we will obtain AZ-aware usage stats by counting up all subresources,
+	// even if we don't end up showing them in the API
+	allSubresources, err := l.buildInstanceSubresources(ctx, projectUUID)
+	if err != nil {
+		return liquid.ServiceUsageReport{}, fmt.Errorf("while collecting instance data: %w", err)
+	}
+
+	for _, subres := range allSubresources {
+		var attrs SubresourceAttributes
+		if err = json.Unmarshal(subres.Attributes, &attrs); err != nil { // TODO: Unmarshalling here + marshalling when building the subresources seems inefficient
```
majewsky, line 163: Consider passing the subresources as type `liquid.SubresourceBuilder[A]` instead.

internal/liquids/nova/capacity.go
```diff
@@ -0,0 +1,572 @@
...
+	return &value
+}
+
+// Like `lhs - rhs`, but never underflows below 0.
+func saturatingSub(lhs, rhs uint64) uint64 {
+	if lhs < rhs {
+		return 0
+	}
+	return lhs - rhs
+}
```
majewsky, line 572: Please use the equivalent functions from package liquids:

https://github.com/sapcc/limes/blob/2dc280056b951470d394a219397eb72cfe70ac89/internal/liquids/utils.go#L69-L80

internal/liquids/nova/hypervisor_subcapacity.go
```diff
@@ -39,16 +43,32 @@ type Subcapacity struct {
...
 	InstancesSubcapacities []any
 	RAMSubcapacities       []any
 }
 
-func (b *PooledSubcapacityBuilder) AddHypervisor(h MatchingHypervisor, maxRootDiskSize float64) {
+// PooledSubcapacityBuilder is used to build subcapacity lists for pooled resources.
+type PooledSubcapacityBuilder struct {
+	// These are actually []Subcapacity, but we store them as []any because
+	// that's what goes into type core.CapacityData in the end.
+	CoresSubcapacities     []liquid.Subcapacity
```
majewsky, line 60: Comment is not relevant to this new type and can be removed.

internal/liquids/nova/hypervisor_subcapacity.go
```diff
@@ -105,3 +161,45 @@ func (b *SplitFlavorSubcapacityBuilder) AddHypervisor(h MatchingHypervisor) {
...
+	}
+	buf, err := json.Marshal(attrs)
+	if err != nil {
+		return fmt.Errorf("while serializing Subcapacity Attributes: %w", err)
+	}
+	b.Subcapacities = append(b.Subcapacities, liquid.Subcapacity{
+		Name:       h.Hypervisor.Service.Host,
+		Capacity:   0,
+		Attributes: json.RawMessage(buf),
+	})
```
majewsky, line 202: It's more idiomatic to use SubcapacityBuilder + Finalize. I did not note this up ahead because there is a case there to reuse the single serialization across three Subresources.

internal/liquids/nova/capacity.go
```diff
@@ -98,6 +109,39 @@ func (c PartialCapacity) IntoCapacityData(resourceName string, maxRootDiskSize f
...
+			Usage:         &c.MemoryMB.Usage,
+			Subcapacities: subcapacities,
+		}
+	case "instances":
+		amount := 10000 * uint64(len(c.MatchingAggregates))
+		if maxRootDiskSize != 0 {
+			maxAmount := uint64(float64(c.LocalGB.Capacity) / maxRootDiskSize)
+			if amount > maxAmount {
+				amount = maxAmount
+			}
```
majewsky, line 133:
```suggestion
			amount = min(amount, maxAmount)
```

internal/liquids/nova/capacity.go
```diff
@@ -110,3 +154,407 @@ func (m PartialCapacityMetric) CappedToUsage() PartialCapacityMetric {
...
+		err = servers.ExtractServersInto(allPages, &instances)
+		if err != nil {
+			return liquid.ServiceCapacityReport{}, fmt.Errorf("while listing active instances for flavor %s: %w", flavor.Name, err)
+		}
+
+		for _, instance := range instances {
+			az := instance.AZ
+			if !slices.Contains(req.AllAZs, az) {
+				az = liquid.AvailabilityZoneUnknown
+			}
```
majewsky, line 304: Can be simplified with NormalizeAZ.

majewsky (changes requested):

majewsky (approved):
Code's looking good now, thanks!

I will add a commit for the documentation once I get to it, hopefully tomorrow.

internal/core/translation_rule.go
```diff
@@ -341,3 +339,126 @@ func translateIronicSubresources(input string, az limes.AvailabilityZone, resNam
...
+			Status:         in.Attributes.Status,
+			Metadata:       in.Attributes.Metadata,
+			Tags:           in.Attributes.Tags,
+			AZ:             in.Attributes.AZ,
+			HypervisorType: "vmware",
+			FlavorName:     in.Attributes.Flavor.Name,
+			VCPUs:          in.Attributes.Flavor.VCPUs,
+			MemoryMiB:      limes.ValueWithUnit{Unit: limes.UnitMebibytes, Value: in.Attributes.Flavor.MemoryMiB},
+			DiskGiB:        limes.ValueWithUnit{Unit: limes.UnitGibibytes, Value: in.Attributes.Flavor.DiskGiB},
+			VideoMemoryMiB: &limes.ValueWithUnit{Unit: limes.UnitMebibytes, Value: *in.Attributes.Flavor.VideoMemoryMiB},
```
majewsky, line 457: This should not panic if the VideoMemoryMiB field is nil.

majewsky (changes requested):

internal/core/translation_rule.go
```diff
@@ -341,3 +339,128 @@ func translateIronicSubresources(input string, az limes.AvailabilityZone, resNam
...
+			Metadata map[string]string       `json:"metadata"`
+			Tags     []string                `json:"tags"`
+			AZ       liquid.AvailabilityZone `json:"availability_zone"`
+			Flavor   struct {
+				Name           string  `json:"name"`
+				VCPUs          uint64  `json:"vcpu"`
+				MemoryMiB      uint64  `json:"ram_mib"`
+				DiskGiB        uint64  `json:"disk_gib"`
+				VideoMemoryMiB *uint64 `json:"video_ram_mib"`
+				HWVersion      string  `json:"-"`
```
majewsky, line 416: Nitpick: Since HWVersion is never serialized, it is irrelevant here (and also on `type oldFormat` below) and can be removed.

internal/core/translation_rule.go
```diff
@@ -341,3 +339,128 @@ func translateIronicSubresources(input string, az limes.AvailabilityZone, resNam
...
+		Traits           []string               `json:"traits"`
+	}
+	outputs := make([]oldFormat, len(inputs))
+	for idx, in := range inputs {
+		out := oldFormat{
+			ServiceHost:      in.Name,
+			AvailabilityZone: az,
+			AggregateName:    in.Attributes.AggregateName,
+			Capacity:         &in.Capacity,
+			Usage:            in.Usage,
```
majewsky, line 388: This only covers the format for pooled resources, where only one value is given for capacity and usage. For split flavors, `CapacityVector` and `UsageVector` need to be filled. A test should be added for the split-flavor case.

majewsky (changes requested):
I have finished the documentation; please counter-review. Also, two comments on the TranslationRule implementations:

internal/core/translation_rule.go
Varsius: From what I understand, we do not report subcapacities on split-flavors anymore. This was only done in the case we could not report on the pooled resources which is know always possible.

majewsky: You're right! I got confused because `type Subcapacity` did not have "deprecated" in it. I have renamed it to `type DeprecatedSubcapacity` now.

majewsky (approved):

PR #630, opened 2024-12-16
ACPQ: fix crash on Forbidden = true resource with AZSeparatedResourceTopology

majewsky:


VoigtS (approved):
Yep, I can see how this leads to issues. Thank you!

PR #636, opened 2024-12-17
Add test cases to prevent future API breaks on AZ separated changes

VoigtS:
As promised in previous PRs. A new fixture was added to cover our existing abstraction levels with one dedicated case.

### DIFF
--- a/internal/api/fixtures/start-data-az-separated.sql
+++ b/internal/api/fixtures/start-data-az-separated.sql
@@ -1,0 +1,13 @@
+CREATE OR REPLACE FUNCTION unix(i integer) RETURNS timestamp AS $$ SELECT TO_TIMESTAMP(i) AT TIME ZONE 'Etc/UTC' $$ LANGUAGE SQL;
+
+INSERT INTO domains (id, name, uuid) VALUES (1, 'germany', 'uuid-for-germany');
+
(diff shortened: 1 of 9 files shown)


internal/api/fixtures/start-data-az-separated.sql
```diff
@@ -0,0 +1,13 @@
...
+
+INSERT INTO projects (id, domain_id, name, uuid, parent_uuid) VALUES (1, 1, 'berlin', 'uuid-for-berlin', 'uuid-for-germany');
+
+INSERT INTO project_services (id, project_id, type, scraped_at, rates_scraped_at, checked_at, rates_checked_at) VALUES (1, 1, 'shared',   UNIX(22), UNIX(23), UNIX(22), UNIX(23));
+
+INSERT INTO project_resources (id, service_id, name, quota, backend_quota) VALUES (1,  1, 'capacity_portion', NULL, NULL);
+
+-- AZ separated resource does not include any az.
+INSERT INTO project_az_resources (id, resource_id, az, quota, usage, physical_usage, subresources) VALUES (1,  1,  'az-one', 5,    1, NULL, '');
+INSERT INTO project_az_resources (id, resource_id, az, quota, usage, physical_usage, subresources) VALUES (2,  1,  'az-two', 5,    1, NULL, '');
```
majewsky, line 13: For the complete prototypical example, we would also expect `backend_quota` to be non-null and equal to `quota`.

VoigtS: Thank you for the catch. I included them

majewsky (approved):

PR #642, opened 2025-01-09
add ironic node to az assignment override

VoigtS:
The assignment of baremetal nodes to an AZ is solved via the ID matching to resource providers. Because the resource provider IDs might mismatch to the node IDs, a manual config override will be provided for the ironic liquid.

Checklist:

### DIFF
--- a/internal/liquids/ironic/liquid.go
+++ b/internal/liquids/ironic/liquid.go
@@ -39,7 +39,7 @@ type Logic struct {
 	// configuration
 	WithSubcapacities bool                               `json:"with_subcapacities"`
 	WithSubresources  bool                               `json:"with_subresources"`
-	NodeToAZOvewrites map[string]liquid.AvailabilityZone `json:"node_to_az_overwrites"`
+	NodeToAZOverrides map[string]liquid.AvailabilityZone `json:"node_to_az_overrides"`
 	// connections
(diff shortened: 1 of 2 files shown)


internal/liquids/ironic/liquid.go
```diff
@@ -37,8 +37,9 @@ import (
 
 type Logic struct {
 	// configuration
-	WithSubcapacities bool `json:"with_subcapacities"`
-	WithSubresources  bool `json:"with_subresources"`
+	WithSubcapacities bool                               `json:"with_subcapacities"`
+	WithSubresources  bool                               `json:"with_subresources"`
+	NodeToAZOvewrites map[string]liquid.AvailabilityZone `json:"node_to_az_overwrites"`
```
majewsky, line 42: Typo:
```suggestion
	NodeToAZOverwrites map[string]liquid.AvailabilityZone `json:"node_to_az_overwrites"`
```

Also, you probably mean "override" instead of "overwrite".

Also also, please add one line of comment that this is to cover legacy deployments.

majewsky (changes requested):

internal/liquids/ironic/liquid.go
VoigtS: Yes, I included the mentioned points.

majewsky (approved):

PR #644, opened 2025-01-14
adjust ironic limit

VoigtS:
New baremetal deployments in eu-de-1 cause memory allocation spikes for the ironic liquid. 
We already had an optimization for the taxing node details in [in ths commit](https://github.com/sapcc/limes/pull/608/files).
I would like to try and reduce hte paging limit for node details. Local tests are too fluctuating to make a clear difference, the memory allocations to each page seem to decrease tho. If this has no notable impact, I will revert this change.


majewsky (commented):
How about we make this a config option, with a default of 100 as before?

majewsky (approved):
@VoigtS I added docs and also fixed a few other concerns. Please take a look and merge if okay.

VoigtS:
@majewsky Thank you very much.

PR #646, opened 2025-01-17
Liquid Cinder: Add FCD support

VoigtS:
This PR aims to support FCD checks from Pools to VolumeTypes via the `StorageProtocol` and the `QualityType`.
The old way of matching against the `volumeBackendName` is still available for the current Pools.

The catch on the new approach is that `VolumeTypes` might a `Encryption` attribute attachted to them at some point. Those will have the same `StorageProtocol` and `QualityType` as the the one without encryption.
This behavior is similar how we treat `volumeBackendName` already. The collected Pool capacity will just be split into the matching volumeTypes.

I tested the implementation against the cluster and yes, we do retrieve non-matching Pools into the resultset now. 
Please have a look at this, there might be optimization that I can do. Other than that I will update the documentation for the new matching tomorrow.

Checklist:
TODO:

### DIFF
--- a/docs/liquids/cinder.md
+++ b/docs/liquids/cinder.md
@@ -47,7 +47,7 @@ The following methods to assign pools to volume types are implemented:
 | Matching | Volume Type         | Pools               |
 | -------- | ------------------- | ------------------- |
 | Type 1   | volume_backend_name | volume_backend_name |
-| Type2    | storage_protocol    | storage_protocol    |
+| Type 2   | storage_protocol    | storage_protocol    |
 |          | quality_type        | quality_type        |
 
 Pools without a matching volume type are ignored.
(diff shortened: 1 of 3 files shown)


docs/liquids/cinder.md
```diff
@@ -28,31 +28,46 @@ When writing quota, the overall quotas are set to the sum of the respective volu
...
+
+- Regular Pools are grouped into volume types by matching their `volume_backend_name` against `extra_specs.volume_backend_name` of the volume type.
+- Remaining Pools are grouped into volume types by matching their `storage_protocol` and `quality_type` against the `extra_specs.storage_protocol` and
+  `extra_secs.quality_type` of the volume type.
+
+| Matching | Volume Type         | Pools               |
+| -------- | ------------------- | ------------------- |
+| Type 1   | volume_backend_name | volume_backend_name |
+| Type 2   | storage_protocol    | storage_protocol    |
+|          | quality_type        | quality_type        |
```
majewsky, line 51: To avoid this being two separate rows:
```suggestion
| Type 2   | storage_protocol<br>quality_type    | storage_protocol<br>quality_type    |
```
Markdown does not support colspan, but I found `<br>` to look reasonably nice.

majewsky (approved):
Nice, thanks!

PR #653, opened 2025-01-31
Add nova topologies

VoigtS:
Add missing resource topologies for nova liquid.


internal/liquids/nova/liquid.go
```diff
@@ -139,32 +139,36 @@ func (l *Logic) BuildServiceInfo(ctx context.Context) (liquid.ServiceInfo, error
 	if err != nil {
 		return liquid.ServiceInfo{}, err
 	}
-
 	resources := map[liquid.ResourceName]liquid.ResourceInfo{
```
SuperSandro2000, line 142:
```suggestion

	resources := map[liquid.ResourceName]liquid.ResourceInfo{
```

PR #654, opened 2025-02-04
Acpq quota overrides bug

Varsius:
Checklist:

### DIFF
--- a/internal/datamodel/apply_computed_project_quota.go
+++ b/internal/datamodel/apply_computed_project_quota.go
@@ -364,7 +364,7 @@ func acpqComputeQuotas(stats map[limes.AvailabilityZone]clusterAZAllocationStats
 			}
 		}
 	}
-	target.EnforceConstraints(constraints, allAZsInOrder, isProjectResourceID, isAZAware)
+	target.EnforceConstraints(stats, constraints, allAZsInOrder, isProjectResourceID, isAZAware)
 	target.TryFulfillDesired(stats, cfg, allowsQuotaOvercommit)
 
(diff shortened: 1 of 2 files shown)


internal/datamodel/apply_computed_project_quota.go
```diff
@@ -430,14 +434,39 @@ func (target acpqGlobalTarget) EnforceConstraints(constraints map[db.ProjectReso
...
+			// phase 2: if not enough quota could be assigned due to capacity constraints,
+			// iniate second distribution round with distribution proportionally to the available capacity.
+			// Since min quota should be enforced, more quota than available capacity may be distributed
+			if missingQuota > 0 {
+				capacityScalePerAZ := make(map[limes.AvailabilityZone]uint64)
+				for _, az := range allAZs {
+					if az == limes.AvailabilityZoneAny && isAZAware {
+						continue
+					}
+					capacityScalePerAZ[az] = *c.MinQuota * stats[az].Capacity // TODO: Possible overflow?
```
Varsius, line 461: There might be problems if for example quota and capacity are measured in B or KB. In this case, we would multiply two very large numbers which could cause an overflow.

majewsky: Indeed, we already had problems with integer overflow in ACPQ in the past. See this testcase: https://github.com/sapcc/go-bits/blob/dd50ca8a0a5f3ed376485a79eb01a63d4f5c4d06/liquidapi/algorithms_test.go#L32

majewsky (approved):

PR #661, opened 2025-02-11
add api endpoint for commitment merging

Varsius:
Checklist:


majewsky (changes requested):

internal/db/models.go
```diff
@@ -207,6 +208,20 @@ const (
...
+	RelatedCommitmentIDs []ProjectCommitmentID `json:"related_ids,omitempty"`
+}
+
+type CommitmentReason string
+
+const (
+	CommitmenReasonCreate  CommitmentReason = "create"
+	CommitmenReasonSplit   CommitmentReason = "split"
+	CommitmenReasonConvert CommitmentReason = "convert"
+	CommitmenReasonMerge   CommitmentReason = "merge"
```
majewsky, line 222: Typo: CommitmenReason -> CommitmentReason

internal/db/migrations.go
```diff
@@ -187,4 +187,73 @@ var sqlMigrations = map[string]string{
...
+		WITH creation_context_data AS (
+			SELECT id as commitment_id, predecessor_id,
+				CASE 
+					WHEN predecessor_id IS NULL THEN 'create'
+					WHEN EXISTS (
+						SELECT 1
+						FROM project_commitments p2
+						WHERE p2.id = predecessor_id
+						AND p2.az_resource_id = project_commitments.az_resource_id
+					) THEN 'split'
```
majewsky, line 209: I like the way you think, but this does not work in the general case. When a split is done, and then afterwards the new commitment is moved into a different project, the `az_resource_id` will be different because it is still the same resource, but in a different project. You will need to join up to `project_resources` to match on the resource name.

internal/db/migrations.go
```diff
@@ -187,4 +187,73 @@ var sqlMigrations = map[string]string{
...
+			END
+		)
+		FROM creation_context_data
+		WHERE project_commitments.id = creation_context_data.commitment_id;
+
+		-- Step 3: Populate supersede context
+		WITH supersede_context_data AS (
+			SELECT p1.id AS superseded_id, p2.id AS successor_id, p2.az_resource_id AS successor_az_resource_id,
+				CASE 
+					WHEN p1.az_resource_id = p2.az_resource_id THEN 'split'
```
majewsky, line 230: Same here.

internal/db/migrations.go
```diff
@@ -187,4 +187,73 @@ var sqlMigrations = map[string]string{
 		ALTER TABLE project_az_resources
 			ADD COLUMN backend_quota BIGINT default NULL;
 	`,
+	"047_commitment_workflow_context.down.sql": `
+		-- We will probably not need this, no implementation for now
+	`,
+	"047_commitment_workflow_context.up.sql": `
+		-- Step 1: Create new fields for commitment workflow contexts
+		ALTER TABLE project_commitments
+			ADD COLUMN creation_context_json JSONB,
```
majewsky, line 196: This field must always be populated, so it should be NOT NULL. We cannot create it as such right away because it must be filled first, but you can set this after the fact with `ALTER TABLE table ALTER COLUMN column SET NOT NULL`.

majewsky (changes requested):
I will do a more detailed review later today or tomorrow, but for the time being, here are comments on the DB migration and model:

majewsky (commented):
FYI, I read a bit more (didn't get to finish again....). To make the feedback cycle shorter, I added some commits for minor stuff. Please review my commits, esp. the one touching the migrations.

internal/api/commitment.go
```diff
@@ -519,7 +680,7 @@ func (p *v1Provider) DeleteProjectCommitment(w http.ResponseWriter, r *http.Requ
 func (p *v1Provider) canDeleteCommitment(token *gopherpolicy.Token, commitment db.ProjectCommitment) bool {
 	// up to 24 hours after creation of fresh commitments, future commitments can still be deleted by their creators
 	if commitment.State == db.CommitmentStatePlanned || commitment.State == db.CommitmentStatePending || commitment.State == db.CommitmentStateActive {
-		if commitment.PredecessorID == nil && p.timeNow().Before(commitment.CreatedAt.Add(24*time.Hour)) {
+		if p.timeNow().Before(commitment.CreatedAt.Add(24 * time.Hour)) {
```
majewsky, line 683: We still want to have the equivalent of `PredecessorID == nil` here. Otherwise, merging commitments would be a loophole to allow deletion, because the merged commitment has `CreatedAt = time.Now()`. Since PredecessorID is gone, this should instead whether the creation context says `reason == "create"`.

internal/collector/commitment_cleanup_test.go
```diff
@@ -111,27 +117,35 @@ func TestCleanupOldCommitmentsJob(t *testing.T) {
...
-		PredecessorID: pointerTo(db.ProjectCommitmentID(3)),
+		ID:                  4,
+		AZResourceID:        2,
+		Amount:              10,
+		Duration:            commitmentForOneDay,
+		CreatedAt:           s.Clock.Now().Add(-oneDay).Add(5 * time.Minute),
+		ConfirmedAt:         pointerTo(s.Clock.Now().Add(-oneDay)),
+		ExpiresAt:           s.Clock.Now(),
+		State:               db.CommitmentStateActive,
+		CreationContextJSON: json.RawMessage(buf),
```
majewsky, line 148: This commitment (ID=4) says here that it was created directly, but SupersedeContext on ID=3 says that ID=4 is the result of a convert, so this CreationContext should specify the same reason.

majewsky (changes requested):
We are nearing completion here. I pushed another commit with just whitespace changes (esp. in `tr.DBChanges().AssertEqualf()`; that fellow normalizes whitespace on its input, so you can make liberal use of indentation to make the test code look neat).

Finally, since #667 was merged, you will need to do a rebase. There will be a few merge conflicts.

majewsky (approved):
When you merge this, please take a look at the qa-de-1 database to see how the DB migration goes.

PR #667, opened 2025-02-18
Add Mail support to limes

VoigtS:
This PR introduces mailing functionality to limes.
Confirmed and expired comitments should queue the commitment information for a project and send the email body to an API that is able to send those emails to their proper recipients.

Checklist:
- [x] Add functionality for pending commitments that get confirmed (opt-in for the customer)
- [x] Add functionality to transfer queued mails to an appropriate API.
- [x] Add functionality for expired commitments


internal/collector/capacity_scrape_test.go
```diff
@@ -613,3 +616,127 @@ func Test_ScanCapacityWithCommitments(t *testing.T) {
...
+					UPDATE cluster_capacitors SET scraped_at = %d, next_scrape_at = %d WHERE capacitor_id = 'scans-second';
+				`,
+			scrapedAt1.Unix(), scrapedAt1.Add(15*time.Minute).Unix(),
+			scrapedAt2.Unix(), scrapedAt2.Add(15*time.Minute).Unix(),
+		))
+	}
+
+	desyncedAt1 := s.Clock.Now().Add(-5 * time.Second)
+	desyncedAt2 := s.Clock.Now()
+	tr.DBChanges().AssertEqualf(`%s
```
majewsky, line 649: Since this change is not significant for this test, you can also `tr.DBChanges().Ignore()` them.

internal/core/cluster.go
```diff
@@ -40,6 +41,7 @@ import (
 // cluster.
 type Cluster struct {
 	Config          ClusterConfiguration
+	MailTemplate    *template.Template
```
majewsky, line 44: We will need multiple mail templates for different occasions, so this should be a struct.

internal/core/config.go
```diff
@@ -149,7 +150,7 @@ type AutogrowQuotaDistributionConfiguration struct {
 
 // NewClusterFromYAML reads and validates the configuration in the given YAML document.
 // Errors are logged and will result in program termination, causing the function to not return.
-func NewClusterFromYAML(configBytes []byte) (cluster *Cluster, errs errext.ErrorSet) {
+func NewClusterFromYAML(configBytes, mailTemplate []byte) (cluster *Cluster, errs errext.ErrorSet) {
```
majewsky, line 153: The mail template should be given as part of the config file, i.e. somewhere in `ClusterConfiguration`.

majewsky (changes requested):

internal/core/config.go
VoigtS: I moved the config over. The slight disadvantage here is that by reading the config, I lose the ability to turn the file string into a template early. This functionality is moved into the `getMailContent`.

majewsky: No, that should definitely happen on startup. Just because we take the template strings as part of the config, does not mean that we need to use them from there each time. We can have the compiled templates as part of `type Cluster`, similar to how we compile ServiceConfiguration into QuotaPlugin and store those in the Cluster.

internal/core/cluster.go
```diff
@@ -79,6 +86,18 @@ func NewCluster(config ClusterConfiguration) (c *Cluster, errs errext.ErrorSet)
...
 
+	// Create mail templates
+	confirmTPl, err := template.New("confirm").Parse(config.MailForms.ConfirmedCommitments)
+	if err != nil {
+		errs.Addf("could not parse mail template: %w", err)
+	}
+	c.MailTemplates.ConfirmedCommitments = confirmTPl
+	expireTpl, err := template.New("expire").Parse(config.MailForms.ExpiringCommitments)
+	if err != nil {
+		errs.Addf("could not parse mail template: %w", err)
```
SuperSandro2000, line 97: We should return different errors here to know which template had the error.

internal/collector/mail_delivery.go
```diff
@@ -0,0 +1,110 @@
...
+	// how long to wait after error before retrying sending mails
+	mailDeliveryErrorInterval = 24 * time.Hour
+)
+
+func (c *Collector) MailDeliveryJob(registerer prometheus.Registerer, client MailDelivery) jobloop.Job {
+	return (&jobloop.ProducerConsumerJob[MailDeliveryTask]{
+		Metadata: jobloop.JobMetadata{
+			ReadableName: "mail delivery",
+			CounterOpts: prometheus.CounterOpts{
+				Name: "limes_mail_delivery",
```
majewsky, line 50: Names for counter metric families should be plural:
```suggestion
				Name: "limes_mail_deliveries",
```

internal/collector/mail_delivery_test.go
```diff
@@ -0,0 +1,89 @@
...
+	tr, tr0 := easypg.NewTracker(t, s.DB.Db)
+	tr0.Ignore()
+	// day 1: successfully send a mail
+	mustT(t, job.ProcessOne(s.Ctx))
+	tr.DBChanges().AssertEqualf(`DELETE FROM project_mail_notifications WHERE id = 1;`)
+
+	// day 2: fail a mail delivery and increase the fail counter
+	s.Clock.StepBy(24 * time.Hour)
+	nextDelivery := s.Clock.Now().Add(24 * time.Hour)
+	mustT(t, job.ProcessOne(s.Ctx))
```
majewsky, line 82: This should fail and display the error reported by the mock MailClient.

internal/core/config.go
```diff
@@ -47,6 +47,7 @@ type ClusterConfiguration struct {
 	ResourceBehaviors        []ResourceBehavior                `yaml:"resource_behavior"`
 	RateBehaviors            []RateBehavior                    `yaml:"rate_behavior"`
 	QuotaDistributionConfigs []*QuotaDistributionConfiguration `yaml:"quota_distribution_configs"`
+	MailForms                MailForms                         `yaml:"mail_templates"`
```
majewsky, line 50: Variable names should match config file field names if possible:
```suggestion
	MailTemplates            MailTemplateConfiguration         `yaml:"mail_templates"`
```

internal/datamodel/commitment_mail_notification.go
```diff
@@ -0,0 +1,77 @@
...
+	Commitments []CommitmentInfo
+}
+
+type CommitmentInfo struct {
+	Commitment db.ProjectCommitment
+	Date       string
+	Resource   AZResourceLocation
+}
+
+func (m MailInfo) CreateMailNotification(tpl *template.Template, subject string, projectID db.ProjectID, now time.Time) (db.MailNotification, error) {
```
majewsky, line 42: I don't like that the method receiver is `MailInfo` here. The method receiver should usually be "the thing that does the action". It makes much more sense to me to have the template be the method receiver, since it acts to generate the mail payload, whereas `MailInfo` is only a passive participant as a data payload.

Second, I don't like that the mail subject ends up being a hardcoded argument at the callsite. This should be configurable just like the mail body. I suggest replacing `core.MailForms` and `core.MailTemplates` as follows:

```go
type MailTemplateConfiguration struct { // field in type Config
	ConfirmedCommitments MailTemplate `yaml:"confirmed_commitments"`
	ExpiringCommitments  MailTemplate `yaml:"expiring_commitments"`
}

type MailTemplate struct {
	Subject string `yaml:"subject"`
	Body string `yaml:"body"`
	Compiled *template.Template `yaml:"-"` // filled during Config.Validate()
}

func (t MailTemplate) Render(m MailInfo /* or whatever it's called */, projectID db.ProjectID, now time.Time) (db.MailNotification, error) { ... }
```

majewsky (changes requested):
I will stop here because a lot of these comments already have to refer to each other, and I don't want to make the confusion bigger.

majewsky (approved):
I made another round of reviews as direct code commits; please counter-review these. If these look good, we're pretty close to done here. One last thing: Please add documentation.

- `docs/operators/config.md` should explain the new required environment variable, and what kind of API we expect to have available under that endpoint.
- Each public declaration (i.e. every upper-cased `type`, `func`, `var`) should have at least one or two sentences of documentation, such that `godoc` output or <https://pkg.go.dev/github.com/sapcc/limes/internal/core> etc. can be navigated in a useful way.

majewsky (approved):

PR #672, opened 2025-02-28
[mail] handle queue for projects without metadata

VoigtS:
This implementation handles the way of how we deal with projects that don't maintain any metadata.
If no subproject/project or domain delivers any metadata, the queue will be freed of the mail request and a log will be placed.

### DIFF
--- a/internal/collector/mail_client.go
+++ b/internal/collector/mail_client.go
@@ -29,7 +29,7 @@ import (
 // MailClient is an interface that provides the methods to communicate with a mail backend service.
 type MailClient interface {
 	// Builds the request to sent the mail content to a mail API.
-	PostMail(ctx context.Context, req MailRequest) error
+	PostMail(ctx context.Context, req MailRequest) (*http.Response, error)
 }
 
 // mailClientImpl is an implmentation of MailClient.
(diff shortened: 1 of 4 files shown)


internal/collector/mail_client.go
```diff
@@ -29,7 +29,7 @@ import (
 // MailClient is an interface that provides the methods to communicate with a mail backend service.
 type MailClient interface {
 	// Builds the request to sent the mail content to a mail API.
-	PostMail(ctx context.Context, req MailRequest) error
+	PostMail(ctx context.Context, req MailRequest) (*http.Response, error)
```
majewsky, line 32: I don't like that we're leaking an entire HTTP response across this abstraction boundary. An alternative way would be to return a Result type with just the information that we need (e.g. `type PostMailResult struct { IsUndeliverable bool }`). But since undeliverability is a special type of error, I suggest instead that we make it a special error type. So this function continues to return just an error and the error might be:

```
type UndeliverableMailError struct {
  Inner error
}

func (e UndeliverableMailError) Error() string { ... } // implements https://pkg.go.dev/builtin#error
func (e UndeliverableMailError) Unwrap() error { ... } // implements the interface implied by https://pkg.go.dev/errors
```

This can be checked at the callsite using `if uerr, ok := errext.As[UndeliverableMailError](err); ok {`.

internal/collector/mail_delivery.go
```diff
@@ -90,8 +92,17 @@ func (c *Collector) processMailDeliveryTask(ctx context.Context, task db.MailNot
...
+	resp, mailErr := client.PostMail(ctx, request)
 	if mailErr != nil {
+		defer resp.Body.Close()
+		// remove the mail queue for (sub)projects with no maintained metadata
+		if resp.StatusCode == http.StatusTeapot {
+			_, err := c.DB.Delete(&task)
+			if err != nil {
+				return err
+			}
+			logg.Error("Mail delivery detected a project with no managed metadata. ProjectID: %s", task.ProjectID)
```
majewsky, line 104: Please also log the error message reported by Campfire, just so we have it in the log somewhere.

majewsky (changes requested):

majewsky (approved):
I added a commit that changes an undeliverable signal to not count as a task failure (which is going to be relevant for alerts on the mail delivery jobloop). If you agree, please merge.

PR #673, opened 2025-02-28
commitments: Add notify on confirm (mail) support

VoigtS:
If this implementation fits the desired outcome, I will create a PR for go-api-declarations before this gets merged.

### DIFF
--- a/internal/api/commitment.go
+++ b/internal/api/commitment.go
@@ -400,7 +400,7 @@ func (p *v1Provider) CreateProjectCommitment(w http.ResponseWriter, r *http.Requ
 		CreationContextJSON: json.RawMessage(buf),
 	}
 	if req.NotifyOnConfirm && req.ConfirmBy == nil {
-		http.Error(w, "notification on confirm can't be set for commitments with immediate confirmation", http.StatusConflict)
+		http.Error(w, "notification on confirm cannot be set for commitments with immediate confirmation", http.StatusConflict)
 		return
 	}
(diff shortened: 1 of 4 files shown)


internal/api/commitment.go
```diff
@@ -398,6 +399,9 @@ func (p *v1Provider) CreateProjectCommitment(w http.ResponseWriter, r *http.Requ
 		ExpiresAt:           req.Duration.AddTo(unwrapOrDefault(confirmBy, now)),
 		CreationContextJSON: json.RawMessage(buf),
 	}
+	if req.ConfirmBy != nil {
+		dbCommitment.NotifyOnConfirm = req.NotifyOnConfirm
+	}
```
majewsky, line 404: The docstring in go-api-declarations says that NotifyOnConfirm "can only be set if ConfirmBy is filled". If we're using that kind of language, we should actually make good on it and reject `req.NotifyConfirm && req.ConfirmBy == nil`.

majewsky (changes requested):
Please go ahead with the go-api-declarations PR.

VoigtS:
The PR for go-api-declarations: https://github.com/sapcc/go-api-declarations/pull/42

majewsky (approved):
I have added two commits fixing minor nitpicks:
1. Underscores are a syntax element in Markdown, so it's good practice to put snake\_case words in codeblocks.
2. It's not good style to use contractions in written English, so I replaced `can't` by `cannot` in an error message.

PR #675, opened 2025-03-06
Introduce MockLiquidClient

wagnerd3:
I figured I will open a draft PR, so that I can record some questions here directly on the code.

Checklist:


internal/collector/fixtures/scrape_data_metrics.prom
```diff
@@ -1,101 +1,101 @@
...
-limes_project_commitment_min_expires_at{domain="germany",domain_id="uuid-for-germany",project="berlin",project_id="uuid-for-berlin",resource="things",service="unittest",service_name="generic-unittest"} 0
-limes_project_commitment_min_expires_at{domain="germany",domain_id="uuid-for-germany",project="dresden",project_id="uuid-for-dresden",resource="capacity",service="unittest",service_name="generic-unittest"} 3.154507e+07
-limes_project_commitment_min_expires_at{domain="germany",domain_id="uuid-for-germany",project="dresden",project_id="uuid-for-dresden",resource="capacity_portion",service="unittest",service_name="generic-unittest"} 0
-limes_project_commitment_min_expires_at{domain="germany",domain_id="uuid-for-germany",project="dresden",project_id="uuid-for-dresden",resource="things",service="unittest",service_name="generic-unittest"} 0
+limes_project_commitment_min_expires_at{domain="germany",domain_id="uuid-for-germany",project="berlin",project_id="uuid-for-berlin",resource="capacity",service="unittest",service_name="someServiceName"} 3.154326e+07
```
wagnerd3, line 28: The values here change because one test case is commented out (not valid), so one timeslice is missing.

internal/collector/scrape_test.go
```diff
@@ -166,50 +274,59 @@ func Test_ScrapeSuccess(t *testing.T) {
...
+	serviceUsageReport.Resources["capacity"].Quota = p2i64(110)
+	serviceUsageReport.Resources["things"].PerAZ["az-two"].Usage = 3
+	serviceUsageReport.Resources["things"].PerAZ["az-two"].Subresources = append(serviceUsageReport.Resources["things"].PerAZ["az-two"].Subresources, liquid.Subresource{Name: "index", Usage: pointerTo(uint64(4))})
+	serviceUsageReport.Metrics["things_usage"] = []liquid.Metric{{Value: 3}}
 	// Scrape should pick up the changed resource data
 	// (no quota sync should be requested since there is one requested already)
 	mustT(t, job.ProcessOne(s.Ctx, withLabel))
 	mustT(t, job.ProcessOne(s.Ctx, withLabel))
 
+	// TODO: How does the ResourceScrapeJob trigger backend_quota updates? The jobs are distinct from each other. This caused quite some confusion when understanding the tests.
```
wagnerd3, line 286: This would be interesting to understand

wagnerd3 (commented):
Will probably do another round of comments when more files were changed.

internal/collector/scrape_test.go
```diff
@@ -537,50 +659,39 @@ func Test_ScrapeButNoResources(t *testing.T) {
...
+
 	c := getCollector(t, s)
 	job := c.ResourceScrapeJob(s.Registry)
 	withLabel := jobloop.WithLabel("service_type", "noop")
 
 	// check that Scrape() behaves properly when encountering a quota plugin with
 	// no Resources() (in the wild, this can happen because some quota plugins
 	// only have Rates())
-	mustT(t, job.ProcessOne(s.Ctx, withLabel))
+	mustFailT(t, job.ProcessOne(s.Ctx, withLabel), errors.New(`during resource scrape of project germany/berlin: missing report for resource "things"`))
```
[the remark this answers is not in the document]
majewsky: Generally, I'm fine with becoming more strict about this, yes.

internal/collector/scrape_test.go
```diff
@@ -176,38 +282,42 @@ func Test_ScrapeSuccess(t *testing.T) {
...
-		UPDATE project_resources SET max_quota_from_backend = 1000 WHERE id = 3 AND service_id = 1 AND name = 'things';
-		UPDATE project_resources SET min_quota_from_backend = 10 WHERE id = 4 AND service_id = 2 AND name = 'capacity';
-		UPDATE project_resources SET max_quota_from_backend = 1000 WHERE id = 6 AND service_id = 2 AND name = 'things';
-		UPDATE project_services SET scraped_at = %[1]d, checked_at = %[1]d, next_scrape_at = %[2]d WHERE id = 1 AND project_id = 1 AND type = 'unittest';
-		UPDATE project_services SET scraped_at = %[3]d, checked_at = %[3]d, next_scrape_at = %[4]d WHERE id = 2 AND project_id = 2 AND type = 'unittest';
-	`,
-		scrapedAt1.Unix(), scrapedAt1.Add(scrapeInterval).Unix(),
-		scrapedAt2.Unix(), scrapedAt2.Add(scrapeInterval).Unix(),
-	)
+	// TODO: How to handle minQuata/maxQuota with liquid?
```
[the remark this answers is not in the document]
majewsky: The old minQuota/maxQuota stuff was built in an excessively generic way. Moving forward, we should only support what LIQUID supports, i.e. `maxQuota = 0` and nothing else. The DB fields `{min,max}_quota_from_backend` should be replaced by `forbidden` at some point, but I'm fine with leaving that for a later PR.

internal/collector/quota_overrides_test.go
```diff
@@ -37,6 +40,32 @@ func TestApplyQuotaOverrides(t *testing.T) {
...
+		InfoVersion: 1,
+		Resources: map[liquid.ResourceName]*liquid.ResourceUsageReport{
+			"capacity": {
+				Quota: pointerTo(int64(100)),
+				PerAZ: map[liquid.AvailabilityZone]*liquid.AZResourceUsageReport{
+					"az-one": {},
+					"az-two": {},
+				},
+			},
+			"capacity_portion": {
```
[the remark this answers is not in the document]
majewsky: I cannot say definitively that we are not using this resource for obscure corner cases, but generally, I'm fine with getting rid of it.

internal/collector/scrape_test.go
```diff
@@ -669,7 +789,9 @@ func Test_TopologyScrapes(t *testing.T) {
 	s.Clock.StepBy(scrapeInterval)
 
 	// topology of a resource changes. Reset AZ-separated backend_quota
-	plugin.LiquidServiceInfo.Resources = map[liquid.ResourceName]liquid.ResourceInfo{"capacity": {Topology: liquid.AZSeparatedResourceTopology}, "things": {Topology: liquid.AZAwareResourceTopology}}
+	resourceEntry := serviceInfo.Resources["things"]
```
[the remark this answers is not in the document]
majewsky: Not that I'm aware of. When building reports, we keep them as `map[someString]*someStruct`, specifically because the pointer indirection in the middle enables updates deep within the data structure. But the Info structs are expected to be built once in a straight-forward manner, so we did not include pointer indirection there to save allocations.

internal/collector/scrape_test.go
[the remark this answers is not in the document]
wagnerd3: created https://github.wdf.sap.corp/cc/storage-resource-services/issues/559

internal/collector/quota_overrides_test.go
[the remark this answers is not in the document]
wagnerd3: @Varsius will take care to check whether the `capacity_portion` impacts the `capacity` in any way. When I looked through the code, I found that this should not be the case.

internal/collector/scrape_test.go
majewsky: Oh, I see. In tests, I usually only run one job at a time. To demonstrate that a quota-sync job _would_ run at a certain point of the process, it's enough in my opinion to have the `tr.DBChanges().AssertEqualf()` that shows that the respective trigger field in the DB gets set accordingly.

internal/collector/quota_overrides_test.go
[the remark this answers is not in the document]
Varsius: I removed all occurences of `capacity_portion` in 7bbf9f5bb84e1fc76831442db3de21f8a92aa953. I encountered two TODOs regarding metrics where I need feedback (see `internal/collector/capacity_scrape_test.go`).

internal/api/liquid_test.go
```diff
@@ -33,35 +36,71 @@ const (
...
+					test_mode: true
 		capacitors:
 		- id: unittest
-			type: --test-static
+			type: liquid
+			params:
+				service_type: unittest
+				test_mode: true
+		resource_behavior:
+		- { resource: unittest/capacity, overcommit_factor: 1.5, topology: flat }
```
majewsky, line 63: Why is it necessary to declare `topology: flat` here? This appears to be redundant with what the ServiceInfo override in TestGetServiceCapacityRequest, and unused in the other test. I'm guessing this is a leftover from before the ServiceInfo override was added? If possible, this should be removed (and then also the addition of the `ResourceBehavior.Topology` field should be reverted).

internal/core/plugin.go
```diff
@@ -282,5 +282,97 @@ var (
...
+// Allows for the implementation of a mock client that is used in unit tests
+type LiquidClient interface {
+	GetInfo(ctx context.Context) (result liquid.ServiceInfo, err error)
+	GetCapacityReport(ctx context.Context, req liquid.ServiceCapacityRequest) (result liquid.ServiceCapacityReport, err error)
+	GetUsageReport(ctx context.Context, projectUUID string, req liquid.ServiceUsageRequest) (result liquid.ServiceUsageReport, err error)
+	PutQuota(ctx context.Context, projectUUID string, req liquid.ServiceQuotaRequest) (err error)
+}
+
+// TODO: this should go in internal/test this does not work due to import cycles
+// MockLiquidClient implements the LiquidClient interface
```
majewsky, line 295: As far as I can see, the import cycle would be `internal/test imports internal/test/plugins imports internal/plugins imports internal/test`. If so, we can fix that. The only relevant usage in `internal/plugins` is calling NewMockClient, so this import cycle can be broken with dependency injection. While we're on it, NewMockClient is too vague of a name, so I will suggest renaming it to NewMockLiquidClient.

Leave the LiquidClient interface in package core, and also put this here:

```go
var NewMockLiquidClient func() LiquidClient
```

Because nothing is assigned, this function pointer is nil and would panic on use. But then, we do this in package test:

```go
type MockLiquidClient struct {...}
func init() {
  core.NewMockLiquidClient = func() core.LiquidClient {
    // ... function body as in existing core.NewMockClient() function ...
  }
}
```

The callsites work as before, with just `core.NewMockLiquidClient()`.

internal/core/resource_behavior.go
```diff
@@ -43,6 +43,7 @@ type ResourceBehavior struct {
 	IdentityInV1API          ResourceRef                         `yaml:"identity_in_v1_api"`
 	TranslationRuleInV1API   TranslationRule                     `yaml:"translation_rule_in_v1_api"`
 	Category                 string                              `yaml:"category"`
+	Topology                 liquid.Topology                     `yaml:"topology"`
```
majewsky, line 46: As per the previous note, this should be removed.

internal/plugins/liquid.go
```diff
@@ -68,9 +69,15 @@ func (p *liquidQuotaPlugin) Init(ctx context.Context, client *gophercloud.Provid
 		p.LiquidServiceType = "liquid-" + string(p.ServiceType)
 	}
 
+	if p.TestMode {
+		p.LiquidClient = core.NewMockClient()
+		p.LiquidServiceInfo, err = p.LiquidClient.GetInfo(ctx)
```
majewsky, line 74: Reminder: When the liquidapi validation PR has been opened and merged, this PR should be rebased on it to ensure that validation is also called for the mock service info (and mock reports).

internal/test/setup.go
```diff
@@ -154,9 +178,97 @@ func NewSetup(t *testing.T, opts ...SetupOption) Setup {
...
+					ProjectID:      dbProject.ID,
+					Type:           svcConfig.ServiceType,
+					ScrapedAt:      &t0,
+					RatesScrapedAt: &t0,
+					CheckedAt:      &t0,
+					RatesCheckedAt: &t0,
+				}
+				mustDo(t, s.DB.Insert(dbProjectService))
+				s.ProjectServices = append(s.ProjectServices, dbProjectService)
+				for _, rsBehavior := range s.Cluster.Config.ResourceBehaviors {
```
majewsky, line 220: This implies that we absolutely must have exactly one resource behavior for each resource, and also we cannot test with resource behaviors that have actual regexes in FullResourceNameRx. Is there a reason why we need to iterate through ResourceBehaviors? I would much rather enumerate resources using the usual interface (`s.Cluster.QuotaPlugins[].Resources()`, which can then be replaced with `s.Cluster.Liquids[].ServiceInfo().Resources` later).

majewsky (changes requested):
This is as much as I could review on the train ride today. I only looked at the more fundamental packages, so `internal/api` and `internal/collector` are not covered by this review yet.

internal/api/liquid_test.go
Varsius: This is needed for creating the database records with
```
test.WithProject(core.KeystoneProject{
	Name: "project-1",
	UUID: "uuid-for-project-1",
}),
test.WithEmptyRecordsAsNeeded,
```

This happens when instantiating the test setup. The service info override is applied afterwards.

internal/plugins/liquid.go
majewsky: The liquidapi validation PR is merged, so please rebase as noted here. In fact, you probably want to rebase on #692 which includes the liquidapi validation PR merge, and also changes some stuff around in test configs because the config format gets touched.

internal/collector/scrape_test.go
majewsky: The misunderstanding was about where the BackendQuota field was filled. Since we do that update in ORM, Daniel did not catch it when looking for an SQL statement.

internal/collector/scrape_test.go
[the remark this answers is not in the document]
majewsky: This test should still be updated to set Forbidden and check the impact of that, even though that is different from what the test previously did.

internal/collector/capacity_scrape_test.go
```diff
@@ -338,13 +507,14 @@ func Test_ScanCapacity(t *testing.T) {
 	registry := prometheus.NewPedanticRegistry()
 	pmc := &CapacityPluginMetricsCollector{Cluster: s.Cluster, DB: s.DB}
 	registry.MustRegister(pmc)
-	assert.HTTPRequest{
+	// TODO: How to handle the metrics?
```
majewsky, line 510: The old test had two bogus metrics ("larger/smaller half") that we could replicate in the new system.

majewsky (changes requested):
Answered the remaining TODOs:

majewsky (approved):
Thank you for your efforts! This is definitely not the final word on how we deal with LIQUID mocking in tests. A lot of it looks ugly, but seeing this diff has given me a better understanding of the problem space. I will try and whip up a proposal for how to evolve this. But for now, we will go ahead and merge this to unblock future developments.

PR #687, opened 2025-04-02
liquid-manila: add empty AZResourceCapacityReports to comply with limes requirements

Varsius:
Checklist:

### DIFF
--- a/internal/liquids/manila/capacity.go
+++ b/internal/liquids/manila/capacity.go
@@ -148,7 +148,17 @@ func (l *Logic) scanCapacityForShareType(ctx context.Context, vst VirtualShareTy
 		SnapshotCapacity:   liquid.ResourceCapacityReport{PerAZ: make(map[liquid.AvailabilityZone]*liquid.AZResourceCapacityReport)},
 		SnapmirrorCapacity: liquid.ResourceCapacityReport{PerAZ: make(map[liquid.AvailabilityZone]*liquid.AZResourceCapacityReport)},
 	}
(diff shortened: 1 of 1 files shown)


internal/liquids/manila/capacity.go
```diff
@@ -148,7 +148,17 @@ func (l *Logic) scanCapacityForShareType(ctx context.Context, vst VirtualShareTy
 		SnapshotCapacity:   liquid.ResourceCapacityReport{PerAZ: make(map[liquid.AvailabilityZone]*liquid.AZResourceCapacityReport)},
 		SnapmirrorCapacity: liquid.ResourceCapacityReport{PerAZ: make(map[liquid.AvailabilityZone]*liquid.AZResourceCapacityReport)},
 	}
-	for az, azPools := range allPoolsByAZ {
+	allAZsWithUnknown := append(allAZs, liquid.AvailabilityZoneUnknown) //nolint:gocritic
```
Varsius, line 151: Is this the correct way of iterating over all AZs and AZ unknown? Feels weird since a "nolint" is required for my solution

majewsky (changes requested):

internal/liquids/manila/capacity.go
majewsky: I have not read this correctly yesterday. Including `unknown` is to catch the `allPoolsByAZ["unknown"]` entry if it exists.

majewsky (approved):

PR #688, opened 2025-04-03
use liquidapi validation functions

Varsius:


internal/plugins/liquid.go
```diff
@@ -113,6 +113,10 @@ func (p *liquidQuotaPlugin) Scrape(ctx context.Context, project core.KeystonePro
 		logg.Fatal("ServiceInfo version for %s changed from %d to %d; restarting now to reload ServiceInfo...",
```
Varsius, line 113: `resp.InfoVersion != p.LiquidServiceInfo.Version` is also checked in `liquid.ValidateUsageReport`. But since we currently terminate here, I left this duplicate check in. The same applies in `capacity_liquid.go`

internal/plugins/utils.go
```diff
@@ -38,23 +37,6 @@ func SortedMapKeys[M map[K]V, K ~string, V any](mapToSort M) []K {
...
-			errs = append(errs, fmt.Errorf("invalid topology: %s on resource: %s", topology, resourceName))
-		}
-	}
-	if len(errs) > 0 {
-		return errors.Join(errs...)
-	}
-	return
-}
-
 func MatchLiquidReportToTopology[V any](perAZReport map[liquid.AvailabilityZone]V, topology liquid.Topology) (err error) {
```
Varsius, line 40: Due to the test structure of `Test_TopologyScrapes` in `internal/collector/scrape_test.go` there is no trivial way of removing this function for now. This will be removed when `GenericQuotaPlugin` becomes deprecated once we solely use liquid plugins in our unit tests.

majewsky: Just so we don't forget about removing it later, please move this into `type GenericQuotaPlugin` as a private helper.

majewsky (changes requested):
So I started to write a review for this last week, but then forgot to click the submit button. Whoopsie.

majewsky (approved):

PR #692, opened 2025-04-08
allow domain-specific commitment configuration

majewsky:
Specifically, commitment durations can be set differently per domain name (and thus, if an empty list of durations is configured for a domain, that will disable commitments altogether). We need this urgently because commitments are not supposed to be accessible to the new customer domains that we will start onboarding soon.

CommitmentBehavior is being split from ResourceBehavior here, in a move that I have been considering for a while: ResourceBehavior is a big and bloated mess, with entries for all kinds of different configurations and relating to different services all mixed together into a big pile (which is over 200 LOC in our production config).

I find it more natural to group behaviors under their respective services, so CommitmentBehavior is the first to move there. Also, it's the first to use the new regexpext.ConfigSet type, which is aimed at making sets like ResourceBehavior more type-safe on the code level.

I considered making the full CommitmentBehavior domain-aware for clarity, i.e. to have the outer type be one of:

```go
ConfigSet[ResourceName, ConfigSet[DomainName, CommitmentBehavior]]
ConfigSet[DomainName, ConfigSet[ResourceName, CommitmentBehavior]]
```

But unfortunately, UntilPercent is categorically a global value, and it also does not really make sense to have different values for ConversionRule per scope, so I ended up only making the Durations domain-aware.


internal/core/commitment_behavior.go
```diff
@@ -0,0 +1,170 @@
...
+
+// CommitmentBehavior describes how commitments work for a single resource.
+//
+// It appears in type ServiceConfiguration.
+type CommitmentBehavior struct {
+	// This ConfigSet is keyed on domain name, because commitment durations
+	// (and thus committability) are allowed to differ per domain.
+	//
+	// If Durations.Pick() returns an empty slice, then commitments are entirely forbidden for that resource in the given domain.
+	Durations regexpext.ConfigSet[string, []limesresources.CommitmentDuration] `yaml:"durations_per_domain"`
```
wagnerd3, line 41: Style question: I find it hard to read that `ConfigSet` has only the generic `key` property and the variable name is `Durations` - so I don't know what value of string the key is. It is the same name as for the `Durations` array of `ScopedCommitmentBehavior`, too. Does it make sense to carry over `DurationsPerDomain`? Or say: `DomainDurations`? Or use `type DomainName string` instead of string (is that compatible with `~string`?) ?

majewsky: Renamed `CommitmentBehavior.{Durations => DurationsPerDomain}`.

wagnerd3 (approved):
I have no further comments, I think. But it's a lot of code, 2nd review is definitely advisable.

PR #707, opened 2025-04-30
merge capacitors into liquid-capacitor

wagnerd3:
Checklist:

Content:

- in this PR the `manual` and `prometheus` capacity plugins move into the `liquid` type
- respectively, there is a configuration change necessary for deployments
- in subsequent PRs, we want to merge `capacitors` and `services` into `liquids` and get rid of the `capacitors` table. I initially started to do this all in one step, but this becomes too large and cluttered very fast. Also, it would involve moving a lot of configuration stuff at once.
- I wrote a test for the `manual` path now but was unsure whether the effort to mock the prometheus client was worth the effort

I tested this PR locally against qa with all the services which have manual or prometheus values and they looked exactly like in qa.


docs/operators/config.md
```diff
@@ -445,68 +447,69 @@ This is something that we plan on changing into a graceful reload in the future.
...
 For information on liquids provided by Limes itself, please refer to the [liquids documentation](../liquids/index.md).
 
-### `manual`
+### `additional data: manual`
 
 ```yaml
 capacitors:
-  - id: manual
-    type: manual-network
+  - id: nova
```
majewsky, line 454: Since the example config says "floating IPs" and "networks":
```suggestion
  - id: neutron
```

internal/plugins/capacity_liquid.go
```diff
@@ -121,6 +139,29 @@ func (p *LiquidCapacityPlugin) Scrape(ctx context.Context, backchannel core.Capa
 		resultInService[resName] = resData
 	}
 
+	// manual capacity collection
+	fixedCapaConfig, fccExist := p.FixedCapacityConfiguration.Unpack()
+	if fccExist {
```
majewsky, line 144: For shortlived success variables like this, it's idiomatic to use short, undistinct names (usually `ok` or `exists`).
```suggestion
	fixedCapaConfig, ok := p.FixedCapacityConfiguration.Unpack()
	if ok {
```
Same below for `pccExist`.

internal/plugins/capacity_liquid.go
```diff
@@ -163,3 +204,60 @@ func (p *LiquidCapacityPlugin) DescribeMetrics(ch chan<- *prometheus.Desc) {
...
+		for _, sample := range unmatchedSamples {
+			unmatchedCapacity += float64(sample.Value)
+		}
+		result[limes.AvailabilityZoneUnknown] = &core.CapacityData{
+			Capacity: uint64(unmatchedCapacity),
+		}
+	}
+
+	// validate result
+	if !p.PrometheusCapacityConfiguration.UnwrapOrPanic("Prometheus config was accessed but not set").AllowZeroCapacity {
```
majewsky, line 252: I would like to avoid UnwrapOrPanic() if the codepath has already proven earlier that the Option is not None.

In this case, since this function only uses `p` to get `p.PrometheusCapacityConfiguration`, I recommend making the whole function a method of `type prometheusCapacityConfiguration`.

majewsky (changes requested):

docs/operators/config.md
wagnerd3: adressed

internal/plugins/capacity_liquid.go
wagnerd3: adressed

internal/plugins/capacity_liquid.go
wagnerd3: adressed

majewsky (approved):
Nice, thanks!

PR #708, opened 2025-04-30
replace resource data types by liquid reports

Varsius:


internal/collector/capacity_scrape.go
```diff
@@ -187,7 +189,12 @@ func (c *Collector) processCapacityScrapeTask(ctx context.Context, task capacity
 	}
 
 	// scrape capacity data
-	capacityData, serializedMetrics, err := plugin.Scrape(ctx, datamodel.NewCapacityPluginBackchannel(c.Cluster, c.DB), c.Cluster.Config.AvailabilityZones)
+	// TODO: How to handle two errors if there is code that needs to be executed in the error case?
+	capacityData, err := plugin.Scrape(ctx, datamodel.NewCapacityPluginBackchannel(c.Cluster, c.DB), c.Cluster.Config.AvailabilityZones)
+	serializedMetrics, err2 := plugins.LiquidSerializeMetrics(plugin.(*plugins.LiquidCapacityPlugin).LiquidServiceInfo.CapacityMetricFamilies, capacityData.Metrics)
+	if err2 != nil {
+		err = err2
+	}
```
majewsky, line 197: Re the TODO comment: The first error should take precedence. So I would say `if err == nil { err = err2 }` instead here, and then we should be fine.

Alternatively, this could be made to read better by adding a small helper function that does Scrape(), returns on error, does LiquidSerializeMetrics(), returns on error, or otherwise does `return capacityData, serializedMetrics, nil`.

internal/collector/scrape_test.go
```diff
@@ -205,6 +205,8 @@ func Test_ScrapeSuccess(t *testing.T) {
 	// a desync should be noted, but we will not run syncJob until later in this test
 	s.Clock.StepBy(scrapeInterval)
 	mustT(t, job.ProcessOne(s.Ctx, withLabel))
+	// TODO: Test fails here because scraping manipulates MockLiquidClient.serviceUsageReport
+	// Is there an elegant way to return a deep copy of the usage report in MockLiquidClient.GetUsageReport()?
```
majewsky, line 209: I don't think there's an elegant way, but it should be fairly straightforward to write a deepcopy function for these reports. Plain values like `int64` or `Option[string]` already copy by value, so you only need to care about maps or slices (including `json.RawMessage`, which is a `[]byte` behind the scenes). So the code ends up being quite a bit shorter than you may expect, see for example https://github.com/sapcc/gatekeeper-addons/blob/fd5a9e3481bb69a0047817a6e16b09a3656b5197/internal/doop/violation.go#L47-L52

There are libraries for deepcopying based on reflection, but I don't want to pull in a library for something that we can write in probably 20-30 lines of code.

internal/collector/metrics.go
```diff
@@ -296,7 +297,7 @@ type QuotaPluginMetricsInstance struct {
 func (c *QuotaPluginMetricsCollector) Describe(ch chan<- *prometheus.Desc) {
 	quotaPluginMetricsOkGauge.Describe(ch)
 	for _, plugin := range c.Cluster.QuotaPlugins {
-		plugin.DescribeMetrics(ch)
+		plugins.LiquidDescribeMetrics(ch, plugin.(*plugins.LiquidQuotaPlugin).LiquidServiceInfo.UsageMetricFamilies, []string{"domain_id", "project_id"})
```
majewsky, line 300: The QuotaPlugin interface already has a method for getting the ServiceInfo:
```suggestion
		plugins.LiquidDescribeMetrics(ch, plugin.ServiceInfo().UsageMetricFamilies, []string{"domain_id", "project_id"})
```

internal/collector/metrics.go
```diff
@@ -247,11 +248,11 @@ func (c *CapacityPluginMetricsCollector) collectOneCapacitor(ch chan<- prometheu
 	if plugin == nil {
 		return
 	}
-	err := plugin.CollectMetrics(ch, []byte(instance.SerializedMetrics), instance.CapacitorID)
+	err := plugins.LiquidCollectMetrics(ch, []byte(instance.SerializedMetrics), plugin.(*plugins.LiquidCapacityPlugin).LiquidServiceInfo.CapacityMetricFamilies, nil, nil)
```
majewsky, line 251: In a similar way, we can avoid this cast into a concrete type by adding the method `ServiceInfo() liquid.ServiceInfo` to the CapacityPlugin interface. Adding this new API surface is not a big deal; it's the same as on QuotaPlugin, and those two plugins will merge together imminently anyway.

(This applies to several other places in this PR, so I'm not going to flag it on the other locations.)

internal/plugins/client_liquid.go
majewsky, line 1: I want to get rid of `package plugins` entirely soon, so please move this file to `internal/collector/serialization.go`. Then all the functions in this file can be private again.

internal/test/mock_liquid_client.go
```diff
@@ -144,3 +146,78 @@ func (l *MockLiquidClient) SetQuotaError(err error) {
...
+		rateReportClone := liquid.RateUsageReport{}
+		rateReportClone.PerAZ = maps.Clone(rateReport.PerAZ)
+		rates[rateName] = &rateReportClone
+	}
+	result.Rates = rates
+
+	metrics := maps.Clone(report.Metrics)
+	for familyName, familyMetrics := range metrics {
+		familyMetrics = slices.Clone(familyMetrics)
+		for i, familityMetric := range familyMetrics {
```
majewsky, line 180:
```suggestion
		for i, familyMetric := range familyMetrics {
```

Same below.

internal/test/mock_liquid_client.go
```diff
@@ -144,3 +146,78 @@ func (l *MockLiquidClient) SetQuotaError(err error) {
...
+			familityMetric.LabelValues = slices.Clone(familityMetric.LabelValues)
+			familyMetrics[i] = familityMetric
+		}
+		metrics[familyName] = familyMetrics
+	}
+	result.Metrics = metrics
```
majewsky, line 221: This is repeated exactly from cloneServiceUsageReport, so we can deduplicate this into a separate function.

main.go
majewsky, line 1: Your changes in this file made me realize that we have not cleaned up the old test harnesses yet. Please delete all the `taskTest...` functions and associated helper functions; they have fully been superseded by the LIQUID test tools in limesctl.

majewsky (changes requested):

majewsky (approved):
Nice, thanks!

PR #710, opened 2025-05-05
replace min/maxQuota with forbidden

Varsius:


internal/db/models.go
```diff
@@ -122,8 +122,7 @@ type ProjectResource struct {
 	Name                     liquid.ResourceName `db:"name"`
 	Quota                    Option[uint64]      `db:"quota"`
 	BackendQuota             Option[int64]       `db:"backend_quota"`
-	MinQuotaFromBackend      Option[uint64]      `db:"min_quota_from_backend"`
-	MaxQuotaFromBackend      Option[uint64]      `db:"max_quota_from_backend"`
+	Forbidden                Option[bool]        `db:"forbidden"`
```
majewsky, line 125: This does not need to be an Option. LIQUID does not allow None here.
```suggestion
	Forbidden                bool        `db:"forbidden"`
```

And then consequently, all other places looking at Forbidden should also not use Option.

internal/db/migrations.go
```diff
@@ -317,4 +317,29 @@ var sqlMigrations = map[string]string{
...
+		UPDATE project_resources
+			SET max_quota_from_backend = 0
+			WHERE forbidden = TRUE;
+		
+		ALTER TABLE project_resources
+			DROP COLUMN forbidden;
+	`,
+	"052_project_resources_forbidden.up.sql": `
+		ALTER TABLE project_resources
+			ADD COLUMN forbidden BOOLEAN DEFAULT FALSE;
```
majewsky, line 334: For the same reason:
```suggestion
			ADD COLUMN forbidden BOOLEAN NOT NULL DEFAULT FALSE;
```

internal/collector/sync_quota_to_backend.go
```diff
@@ -190,14 +196,14 @@ func (c *Collector) performQuotaSync(ctx context.Context, srv db.ProjectService,
...
+			if targetQuotaPtr.IsNone() {
 				return fmt.Errorf("found unexpected NULL value in project_resources.quota for id = %d", resourceID)
 			}
-			targetQuota = *targetQuotaPtr
+			targetQuota, _ = targetQuotaPtr.Unpack()
 		}
 
 		targetQuotasInDB[resourceName] = targetQuota
-		if currentQuota == nil || *currentQuota < 0 || uint64(*currentQuota) != targetQuota {
+		if val, isSome := currentQuotaPtr.Unpack(); !isSome || val < 0 || uint64(val) != targetQuota {
```
majewsky, line 206: Please double-check my logic. But I think this can be shortened to:
```suggestion
		if currentQuotaPtr != Some(targetQuota) {
```
because Option types can be compared to each other by value (unlike pointer types).

internal/collector/sync_quota_to_backend.go
```diff
@@ -190,14 +196,14 @@ func (c *Collector) performQuotaSync(ctx context.Context, srv db.ProjectService,
 			}
 		} else {
 			// for anything except AZSeparatedTopology, the total target quota is just what we have in project_resources.quota
-			if targetQuotaPtr == nil {
+			if targetQuotaPtr.IsNone() {
 				return fmt.Errorf("found unexpected NULL value in project_resources.quota for id = %d", resourceID)
 			}
-			targetQuota = *targetQuotaPtr
+			targetQuota, _ = targetQuotaPtr.Unpack()
```
majewsky, line 202: The Unpack here can be combined with the IsNone check above:
```go
var ok bool
targetQuota, ok = targetQuotaPtr.Unpack()
if !ok {
  ...
}
```

majewsky (changes requested):
Two nitpicks:

internal/collector/sync_quota_to_backend.go
Varsius: `if currentQuotaPtr != Some(int64(targetQuota)) {
	needsApply = true
}`
The targetQuota needs to be converted into int64 for the comparison to work. Besides from that, the logic should work. Could this cause any weird edge cases? Before we converted int64 to uint64 and now it is the other way around. What is the reason that backend_quota and quota are different types?

Varsius: >> golangci-lint
internal/collector/sync_quota_to_backend.go:207:35: G115: integer overflow conversion uint64 -> int64 (gosec)

majewsky: Argh, the conversion lint is a good point. Then we might have to keep it as-is.

majewsky (approved):

PR #711, opened 2025-05-05
remove cluster_capacitors from DB

wagnerd3:
Checklist:

This PR has the next level of changes towards a common configuration for `services[]` and `capacitors`:
* we remove the `cluster_capacitors` table and shift the columns to the `cluster_services`
* we move around the `service_type` to be on the same level as for `services[]` already
* we shift some of the wording - the capacitors only really exist on app layer/ configuration after this change
* as a result, we assume that the `service_type` of a `service` and `capacitor` are the same if they belong together

naturally, this needs a companion PR for the deployment. I tested again locally and all the results in the DB looked the same.

### DIFF
--- a/internal/api/commitment.go
+++ b/internal/api/commitment.go
@@ -126,8 +126,8 @@ var (
 		 WHERE ps.project_id = $1 AND ps.type = $2 AND pr.name = $3 AND par.az = $4
 	`)
 	forceImmediateCapacityScrapeQuery = sqlext.SimplifyWhitespace(`
-		UPDATE cluster_capacitors SET next_scrape_at = $1 WHERE capacitor_id = (
-			SELECT capacitor_id FROM cluster_services cs JOIN cluster_resources cr ON cs.id = cr.service_id
+		UPDATE cluster_services SET next_scrape_at = $1 WHERE id = (
+			SELECT cs.id FROM cluster_services cs JOIN cluster_resources cr ON cs.id = cr.service_id
 			WHERE cs.type = $2 AND cr.name = $3
 		)
 	`)
--- a/internal/api/commitment.go
+++ b/internal/api/commitment.go
@@ -125,12 +125,6 @@ var (
 		  JOIN project_services ps ON pr.service_id = ps.id
 		 WHERE ps.project_id = $1 AND ps.type = $2 AND pr.name = $3 AND par.az = $4
 	`)
-	forceImmediateCapacityScrapeQuery = sqlext.SimplifyWhitespace(`
-		UPDATE cluster_services SET next_scrape_at = $1 WHERE id = (
-			SELECT cs.id FROM cluster_services cs JOIN cluster_resources cr ON cs.id = cr.service_id
-			WHERE cs.type = $2 AND cr.name = $3
-		)
-	`)
 )
 
 // GetProjectCommitments handles GET /v1/domains/:domain_id/projects/:project_id/commitments.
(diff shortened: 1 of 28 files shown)


internal/api/commitment.go
```diff
@@ -126,8 +126,8 @@ var (
 		 WHERE ps.project_id = $1 AND ps.type = $2 AND pr.name = $3 AND par.az = $4
 	`)
 	forceImmediateCapacityScrapeQuery = sqlext.SimplifyWhitespace(`
-		UPDATE cluster_capacitors SET next_scrape_at = $1 WHERE capacitor_id = (
-			SELECT capacitor_id FROM cluster_services cs JOIN cluster_resources cr ON cs.id = cr.service_id
+		UPDATE cluster_services SET next_scrape_at = $1 WHERE id = (
+			SELECT cs.id FROM cluster_services cs JOIN cluster_resources cr ON cs.id = cr.service_id
 			WHERE cs.type = $2 AND cr.name = $3
 		)
```
majewsky, line 132: This can be simplified now using the 1:1 relation between capacitors and services:
```suggestion
		UPDATE cluster_services SET next_scrape_at = $1 WHERE type = $2
```
Also, since the query is only a short oneliner now, it can probably be inlined, too.

internal/collector/capacity_scrape.go
```diff
@@ -43,65 +43,49 @@ import (
...
 	return (&jobloop.ProducerConsumerJob[capacityScrapeTask]{
 		Metadata: jobloop.JobMetadata{
 			ReadableName: "scrape capacity",
 			CounterOpts: prometheus.CounterOpts{
 				Name: "limes_capacity_scrapes",
-				Help: "Counter for capacity scrape operations per capacitor.",
+				Help: "Counter for capacity scrape operations per service_type.",
 			},
-			CounterLabels: []string{"capacitor_id"},
+			CounterLabels: []string{"service_type"},
```
majewsky, line 63: Since the label set changes on this metric, please adjust the relevant alert definition in helm-charts as well.

internal/collector/capacity_scrape.go
```diff
@@ -228,44 +182,39 @@ func (c *Collector) processCapacityScrapeTask(ctx context.Context, task capacity
...
+		wantedResources = append(wantedResources, db.ResourceRef[db.ClusterServiceID]{
+			ServiceID: service.ID,
+			Name:      resourceName,
+		})
 	}
 	slices.SortFunc(wantedResources, db.CompareResourceRefs) // for deterministic test behavior
 
-	// create and delete cluster_resources for this capacitor as needed
+	// create and delete cluster_resources for this cluster_service as needed
 	setUpdate := db.SetUpdate[db.ClusterResource, db.ResourceRef[db.ClusterServiceID]]{
```
majewsky, line 210: We can get rid of the ResourceRef usage here. ResourceRef was necessary before because resources were identified by the pair of keys `(service_id, name)`. But now all resources within a single task have the same `service_id`, so the second type argument can be changed to just `liquid.ResourceName` here.

majewsky (changes requested):

internal/collector/capacity_scrape.go
wagnerd3: I unified this and switched label `capacitor` for `service_type` also for `limes_capacity_plugin_metrics_ok`. 
Companion PR: https://github.com/sapcc/helm-charts/pull/8629

PR #713, opened 2025-05-08
merge capacitors+ services config, plugins to LiquidConnection

wagnerd3:
Checklist:

Scope:
* merge `capacitors` and `services` to `liquids` in config (pay attention to config change for ci)
* rename and move the former `CapacityPlugin` and `QuotaPlugin` to `LiquidConnection`
  * naming discussed with @majewsky - I am still not 100% sure I like this, because now the `LiquidConnection` holds the `LiquidClient`
  * if there is better ideas, let's discuss. essentially, we need the object to hold the state (LiquidServiceInfo), we want to persist in the DB mid-term, so that we can gracefully reload when Liquids are updated
* I moved away from the "plugin" naming wherever possible (except `DiscoveryPlugin` for now), so the jobs for cluster+project scraping are now referred to as `collection` where referencing `liquid` did not make sense (e.g. the respective prom metrics)

As always, I had this run locally against qa-de-1 and checked that both cluster and project scrapes run through as before.


internal/core/config.go
```diff
@@ -39,26 +40,24 @@ import (
...
-	AvailabilityZones []limes.AvailabilityZone `yaml:"availability_zones"`
-	CatalogURL        string                   `yaml:"catalog_url"`
-	Discovery         DiscoveryConfiguration   `yaml:"discovery"`
-	Services          []ServiceConfiguration   `yaml:"services"`
-	Capacitors        []CapacitorConfiguration `yaml:"capacitors"`
-	// ^ Sorry for the stupid pun. Not.
+	AvailabilityZones        []limes.AvailabilityZone         `yaml:"availability_zones"`
+	CatalogURL               string                           `yaml:"catalog_url"`
+	Discovery                DiscoveryConfiguration           `yaml:"discovery"`
+	Liquids                  []LiquidConfiguration            `yaml:"liquids"`
```
majewsky, line 46: I messed up and forgot to mention one important thing in the ticket:

```suggestion
	Liquids                  map[db.ServiceType]LiquidConfiguration            `yaml:"liquids"`
```

I would like for this to be a map instead of a list, because that would enable maybe refactoring the Helm values to reduce the templating a bit. (With lists, everything needs to be in a single values.yaml because there is no good way to merge bits and pieces from multiple values.yaml files together. The only merge strategy for lists is to concatenate them, so a later values.yaml file cannot modify a LiquidConfiguration added by an earlier values.yaml file.)

internal/core/config.go
```diff
@@ -104,6 +111,14 @@ type ServiceConfiguration struct {
 	CommitmentBehaviorPerResource regexpext.ConfigSet[liquid.ResourceName, CommitmentBehavior] `yaml:"commitment_behavior_per_resource"`
 }
 
+type PrometheusCapacityConfiguration struct {
```
majewsky, line 114: I generally like to have a docstring on all public fields, to make documentation pages like https://pkg.go.dev/github.com/sapcc/limes/internal/core useful. For types like this, I usually go with just the minimum amount of breadcrumbs:

```suggestion
// PrometheusCapacityConfiguration appears in type LiquidConfiguration.
type PrometheusCapacityConfiguration struct {
```

internal/core/config.go
```diff
@@ -104,6 +111,14 @@ type ServiceConfiguration struct {
...
+	APIConfig         promquery.Config               `yaml:"api"`
+	Queries           map[liquid.ResourceName]string `yaml:"queries"`
+	AllowZeroCapacity bool                           `yaml:"allow_zero_capacity"`
+}
+
+type FixedCapacityConfiguration map[liquid.ResourceName]uint64
```
majewsky, line 120: I don't think this needs to be its own type.

internal/core/config.go
```diff
@@ -90,12 +89,20 @@ func (c DiscoveryConfiguration) FilterDomains(domains []KeystoneDomain) []Keysto
...
-	PluginType  string              `yaml:"type"`
-	Area        string              `yaml:"area"`
-	Parameters  util.YamlRawMessage `yaml:"params"` // will be unmarshalled into the QuotaPlugin instance
+// LiquidConfiguration describes a service that is enabled for a certain cluster by means of a corresponding running liquid.
+// It holds configurations for how to deal with the service on project level (quota, usage, commitment) as well as cluster level (capacity).
+
+type LiquidConfiguration struct {
+	ServiceType db.ServiceType `yaml:"service_type"`
+	Area        string         `yaml:"area"`
+	// LiquidServiceType is a string that can be used to identify the liquid when generating a liquidapi.NewClient.
```
majewsky, line 98: This description is rather vague:
```suggestion
	// LiquidServiceType is the service type under which the liquid is registered in the Keystone catalog.
```

internal/core/config.go
```diff
@@ -201,27 +208,22 @@ func (cluster ClusterConfiguration) validateConfig() (errs errext.ErrorSet) {
...
+		missing("liquids[]")
 	}
-	//NOTE: cluster.Capacitors is optional
 
-	for idx, srv := range cluster.Services {
+	//NOTE: Liquids[].FixedCapacityConfiguration and Liquids[].PrometheusCapacityConfiguration are optional
+	for idx, srv := range cluster.Liquids {
 		if srv.ServiceType == "" {
-			missing(fmt.Sprintf("services[%d].id", idx))
+			missing(fmt.Sprintf("services[%d].service_type", idx))
```
majewsky, line 218:
```suggestion
			missing(fmt.Sprintf("liquids[%d].service_type", idx))
```
Same for the other messages in this for-loop.

internal/collector/serialization.go
majewsky, line 1: Why are the functions in this file being moved into internal/core? We just split this from the LIQUID plugin in #708 because they are only used in internal/collector and thus make sense as an implementation detail over there.

internal/core/liquid.go
```diff
@@ -0,0 +1,471 @@
...
+	Area                            string
+	LiquidServiceType               string
+	ServiceType                     db.ServiceType
+	FixedCapacityConfiguration      Option[FixedCapacityConfiguration]      `yaml:"fixed_capacity_values"`
+	PrometheusCapacityConfiguration Option[PrometheusCapacityConfiguration] `yaml:"capacity_values_from_prometheus"`
+
+	// state
+	LiquidServiceInfo liquid.ServiceInfo `yaml:"-"`
+	LiquidClient      LiquidClient       `yaml:"-"`
+}
```
majewsky, line 58: 1. This type is not parsed from configuration, so the `yaml:` tags can be removed.
2. This type does not use the `Area` given to it, so we don't need to hold it here.

internal/core/liquid.go
```diff
@@ -0,0 +1,471 @@
...
+
+// MakeLiquidConnection is a factory to fill all necessary configuration fields
+func MakeLiquidConnection(lc LiquidConfiguration) LiquidConnection {
```
majewsky, line 61: By convention, factory functions are called `NewSomething` in Go.
```suggestion
// NewLiquidConnection is a factory to fill all necessary configuration fields
func NewLiquidConnection(lc LiquidConfiguration) LiquidConnection {
```

internal/core/liquid.go
```diff
@@ -0,0 +1,471 @@
...
+		LiquidServiceType:               lc.LiquidServiceType,
+		ServiceType:                     lc.ServiceType,
+		FixedCapacityConfiguration:      lc.FixedCapacityConfiguration,
+		PrometheusCapacityConfiguration: lc.PrometheusCapacityConfiguration,
+	}
+}
+
+// Init is called before any other interface methods, and allows the LiquidConnection to
+// perform first-time initialization.
+func (l *LiquidConnection) Init(ctx context.Context, client *gophercloud.ProviderClient, eo gophercloud.EndpointOpts, serviceType db.ServiceType) (err error) {
```
majewsky, line 76: The connection object already holds the ServiceType:

```suggestion
func (l *LiquidConnection) Init(ctx context.Context, client *gophercloud.ProviderClient, eo gophercloud.EndpointOpts) (err error) {
```

internal/core/liquid.go
```diff
@@ -0,0 +1,471 @@
...
+// ServiceInfo returns metadata for this liquid.
+// This includes metadata for all the resources and rates that this liquid scrapes.
+func (l *LiquidConnection) ServiceInfo() liquid.ServiceInfo {
+	return l.LiquidServiceInfo
+}
+
+// Resources returns the resources that this connections liquid is concerned with.
+func (l *LiquidConnection) Resources() map[liquid.ResourceName]liquid.ResourceInfo {
+	return l.LiquidServiceInfo.Resources
+}
```
majewsky, line 99: This does not appear to have any users, so it can probably be removed.

majewsky (changes requested):
First batch. I only looked at `internal/core` so far.

internal/collector/serialization.go
wagnerd3: I read in the comment `This logic lives here because it is shared between liquidCapacityPlugin and liquidQuotaPlugin.`. As there is no differentiation between `LiquidCapacityPlugin` and `LiquidQuotaPlugin` anymore, I searched for somewhere where it's more central. I get the rationale behind having it in collector though and will move it back + change the comment.

internal/core/liquid.go
wagnerd3: I am very sureI read somewhere, that `MakeSomething` is used when returning a struct and `NewSomething` is used when returning a pointer. Is this a thing? I see the code has this nowhere, so I'll change it.

internal/core/liquid.go
wagnerd3: Same is the case for `Rates()`, I will remove that as well.

majewsky (commented):
Thanks for the effort! I have pushed two commits, one for just whitespace changes and one for naming. Please check if you agree with these, and merge the PRs if you do.

majewsky (approved):
Wrong button :)

PR #715, opened 2025-05-13
move discovery plugins into core and merge method static and --test-static

Varsius:


wagnerd3 (approved):
LGTM

internal/core/config.go
```diff
@@ -63,10 +63,15 @@ func (cluster *ClusterConfiguration) GetLiquidConfigurationForType(serviceType d
...
 // and projects.
 type DiscoveryConfiguration struct {
-	Method          string                `yaml:"method"`
-	ExcludeDomainRx regexpext.PlainRegexp `yaml:"except_domains"`
-	IncludeDomainRx regexpext.PlainRegexp `yaml:"only_domains"`
-	Parameters      util.YamlRawMessage   `yaml:"params"`
+	Method                       string                               `yaml:"method"`
+	ExcludeDomainRx              regexpext.PlainRegexp                `yaml:"except_domains"`
+	IncludeDomainRx              regexpext.PlainRegexp                `yaml:"only_domains"`
+	StaticDiscoveryConfiguration Option[StaticDiscoveryConfiguration] `yaml:"static_config"`
```
majewsky, line 69: This does not need to be an Option. All the contained fields are safe as `nil`, and in fact, the initialization of StaticDiscoveryPlugin will use a zero-valued StaticDiscoveryConfiguration if this field is None, so we can go with that by default.

internal/core/config.go
```diff
@@ -63,10 +63,15 @@ func (cluster *ClusterConfiguration) GetLiquidConfigurationForType(serviceType d
 // DiscoveryConfiguration describes the method of discovering Keystone domains
 // and projects.
 type DiscoveryConfiguration struct {
-	Method          string                `yaml:"method"`
-	ExcludeDomainRx regexpext.PlainRegexp `yaml:"except_domains"`
-	IncludeDomainRx regexpext.PlainRegexp `yaml:"only_domains"`
-	Parameters      util.YamlRawMessage   `yaml:"params"`
```
majewsky, line 69: This was the last usage of `util.YamlRawMessage`, so that type can be removed.

majewsky (changes requested):

majewsky (approved):
Thank!

PR #716, opened 2025-05-15
Save service info in db

wagnerd3:
Checklist:

Outline:
* the `liquid.ServiceInfo` is distributed to the 3 tables `cluster_services, cluster_resources, cluster_rates` and persisted there on startup
* additionally, a graceful reload of the ServiceInfo on change of a liquid version is implemented (`ReconcileLiquidConnection` - I am open for better naming suggestions?)
* cluster_services and cluster_resources are not written by the consistency checks or the scrape itself, instead by the reconciliation

next steps:
* make sure that all liquids emit version numbers correctly (currently not done)
* use the data from the database on startup in case of unavailability of a liquid

### DIFF
--- a/internal/collector/consistency.go
+++ b/internal/collector/consistency.go
@@ -34,59 +34,27 @@ import (
 func (c *Collector) CheckConsistencyJob(registerer prometheus.Registerer) jobloop.Job {
 	return (&jobloop.CronJob{
 		Metadata: jobloop.JobMetadata{
-			ReadableName: "ensure that all active domains and projects in this cluster have a service entry for this liquid's service type",
+			ReadableName: "ensure that all active domains and projects in this cluster have a service entry for the liquid's service types",
 			CounterOpts: prometheus.CounterOpts{
 				Name: "limes_cron_consistency_runs",
 				Help: "Counter for consistency checks runs",
 			},
 		},
 		Interval: 1 * time.Hour,
-		// When new services or resources are added, we need this job to create the respective DB entries immediately upon deployment.
+		// When new services or resources are added, we need this job to populate the project level services densely.
+		// It does not take care of cluster level services, resources, or rates - they are added on demand from
+		// LiquidConnection.ReconcileLiquidConnection() or whenever the collect job is started. Project level resources
+		// and az_resources are created by the scraping job, which picks up the created project_services.
 		InitialDelay: 10 * time.Second,
-		Task:         c.checkConsistencyCluster,
+		Task:         c.CheckConsistencyAllDomains,
 	}).Setup(registerer)
 }
 
-func (c *Collector) checkConsistencyCluster(_ context.Context, _ prometheus.Labels) error {
-	// check cluster_services entries
-	var services []db.ClusterService
-	_, err := c.DB.Select(&services, `SELECT * FROM cluster_services`)
-	if err != nil {
-		return err
-	}
-	logg.Info("checking consistency for %d cluster services...", len(services))
-
-	// cleanup entries for services that have been disabled
-	seen := make(map[db.ServiceType]bool)
-	for _, service := range services {
-		seen[service.Type] = true
-
-		if !c.Cluster.HasService(service.Type) {
-			logg.Info("cleaning up %s cluster service entry", service.Type)
-			_, err := c.DB.Delete(&service)
-			if err != nil {
-				c.LogError(err.Error())
-			}
-		}
-	}
-
-	// create missing service entries
-	for _, serviceType := range c.Cluster.ServiceTypesInAlphabeticalOrder() {
-		if seen[serviceType] {
-			continue
-		}
-
-		logg.Info("creating missing %s cluster service entry", serviceType)
-		err := c.DB.Insert(&db.ClusterService{Type: serviceType, NextScrapeAt: c.MeasureTime()})
-		if err != nil {
-			c.LogError(err.Error())
-		}
-	}
-
+func (c *Collector) CheckConsistencyAllDomains(_ context.Context, _ prometheus.Labels) error {
 	// recurse into domains (with deterministic ordering for the unit test's sake;
 	// the DESC ordering is because I was too lazy to change the fixtures)
 	var domains []db.Domain
-	_, err = c.DB.Select(&domains, `SELECT * FROM domains ORDER BY name DESC`)
+	_, err := c.DB.Select(&domains, `SELECT * FROM domains ORDER BY name DESC`)
 	if err != nil {
 		return err
 	}
--- a/internal/collector/scrape_test.go
+++ b/internal/collector/scrape_test.go
@@ -125,6 +125,7 @@ func commonComplexScrapeTestSetup(t *testing.T) (s test.Setup, scrapeJob jobloop
 	mockLiquidClient, liquidServiceType := test.NewMockLiquidClient(srvInfo)
 	s = test.NewSetup(t,
 		test.WithConfig(fmt.Sprintf(testScrapeBasicConfigYAML, liquidServiceType)),
+		test.WithCollectorSetup,
 	)
 	prepareDomainsAndProjectsForScrape(t, s)
 
@@ -600,6 +601,7 @@ func Test_ScrapeButNoResources(t *testing.T) {
 	mockLiquidClient, liquidServiceType := test.NewMockLiquidClient(srvInfo)
 	s := test.NewSetup(t,
 		test.WithConfig(fmt.Sprintf(testNoopConfigYAML, liquidServiceType)),
+		test.WithCollectorSetup,
 	)
 	prepareDomainsAndProjectsForScrape(t, s)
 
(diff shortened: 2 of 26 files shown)


internal/core/liquid.go
```diff
@@ -84,6 +109,162 @@ func (l *LiquidConnection) Init(ctx context.Context, client *gophercloud.Provide
...
+	if err != nil {
+		return fmt.Errorf("cannot serialize UsageMetricFamilies for %s: %w", l.ServiceType, err)
+	}
+	serviceUpdate := db.SetUpdate[db.ClusterService, db.ServiceType]{
+		ExistingRecords: dbServices,
+		WantedKeys:      wantedServices,
+		KeyForRecord: func(service db.ClusterService) db.ServiceType {
+			return service.Type
+		},
+		Create: func(serviceType db.ServiceType) (db.ClusterService, error) {
```
wagnerd3, line 147: I would like to hand in a `next_scrape_at: time.Now()` here, but I am unsure how to do this elegantly so that I can mock it during testing. I see that other methods which have this problem have a receiver type `e.g. Collector` which get's handed this `nowFunction`. But the `LiquidConnection` is further down the stack and instantiated within all the setups, I think it's overhead to cascade the handover of the `nowFunction` through the levels `Collector → Cluster → LiquidConnection`. How to do this properly?

I thought about moving this reconciliation to a different place, but I think it belongs to the liquid. :-/

wagnerd3: Discussed offline with Stefan: Handing the parameter to the receiver types is the way to go. Will implement this now.

internal/datamodel/apply_computed_project_quota.go
```diff
@@ -83,7 +82,8 @@ func (c *projectLocalQuotaConstraints) AddMaxQuota(value Option[uint64]) {
 
 // ApplyComputedProjectQuota reevaluates auto-computed project quotas for the
 // given resource, if supported by its quota distribution model.
-func ApplyComputedProjectQuota(serviceType db.ServiceType, resourceName liquid.ResourceName, dbm *gorp.DbMap, cluster *core.Cluster, now time.Time) error {
+func ApplyComputedProjectQuota(serviceType db.ServiceType, resourceName liquid.ResourceName, cluster *core.
+	Cluster, now time.Time) error {
```
majewsky, line 86: Unnecessary line break.

internal/datamodel/apply_computed_project_quota.go
```diff
@@ -229,11 +229,9 @@ func ApplyComputedProjectQuota(serviceType db.ServiceType, resourceName liquid.R
 	if err != nil {
 		return fmt.Errorf("while marking updated %s/%s project quotas for sync in DB: %w", serviceType, resourceName, err)
 	}
-
-	return tx.Commit()
-
 	//NOTE: Quotas are not applied to the backend here because OpenStack is way too inefficient in practice.
 	// We wait for the next scrape cycle to come around and notice that `backend_quota != quota`.
+	return tx.Commit()
```
majewsky, line 234: After looking at this in context, we can actually drop the `NOTE` comment here. It applies to how things worked before we added SyncQuotaToBackendJob.

majewsky (changes requested):
I only looked at packages core, db, and datamodel so far.

internal/core/liquid.go
```diff
@@ -68,6 +100,163 @@ func (l *LiquidConnection) Init(ctx context.Context, client *gophercloud.Provide
 	return err
 }
 
+// ReconcileLiquidConnection ensures consistency of tables cluster_services, cluster_resources and cluster_rates
+// with the latest ServiceInfo of this LiquidConnection. It is called whenever the LiquidVersion changes
+// during Scrape or ScrapeCapacity. On startup of the collect task, this function is called from the Cluster
+// where additionally, orphaned entries are removed.
+func (l *LiquidConnection) ReconcileLiquidConnection() (err error) {
```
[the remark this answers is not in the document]
wagnerd3: Mhh, actually it does IMHO: Currently, this is called from `Cluster.ReconcileLiquidConnections` which does additionally the deletion of the orphaned LiquidConnections. If I was to move this call to `Init`, this would be called in `Cluster.Connect` and this would be done for all jobs (collect, api, metrics), because all have a cluster connection. But we only want the collector job to take care of the DB-writing for the LiquidConnections. Do I have a mistake in my train of thought?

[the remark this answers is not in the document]
majewsky: All of this is inside `package core`. Symbols only need to be public if you want to call them from other packages.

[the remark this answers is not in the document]
wagnerd3: Ah, sry - will change this

internal/core/liquid.go
```diff
@@ -60,6 +71,28 @@ func (l *LiquidConnection) Init(ctx context.Context, client *gophercloud.Provide
...
+
+// compareServiceInfoVersions compares a report version of the ServiceInfo with the saved version
+// and triggers the update and persisting if necessary.
+func (l *LiquidConnection) compareServiceInfoVersions(ctx context.Context, infoVersion int64) (err error) {
+	if infoVersion != l.LiquidServiceInfo.Version {
+		logg.Info("ServiceInfo version for %s changed from %d to %d; reloading and persisting ServiceInfo.", l.LiquidServiceType, l.LiquidServiceInfo.Version, infoVersion)
+		err = l.updateServiceInfo(ctx)
+		if err != nil {
+			return err
+		}
```
majewsky, line 85: After running updateServiceInfo, we should check that the ServiceInfo version that we got now matches the one in the report, and report an error if the liquid has now grown inconsistent with itself.

internal/core/liquid.go
```diff
@@ -60,6 +71,28 @@ func (l *LiquidConnection) Init(ctx context.Context, client *gophercloud.Provide
...
+	}
+	return nil
+}
+
+// updateServiceInfo queries the backend service for the latest ServiceInfo and validates it.
+func (l *LiquidConnection) updateServiceInfo(ctx context.Context) (err error) {
 	l.LiquidServiceInfo, err = l.LiquidClient.GetInfo(ctx)
```
majewsky, line 96: Since we allow updating `l.LiquidServiceInfo` as the program is running, and esp. while multiple goroutines are accessing it, access to it must be guarded to protect against data races. I recommend adding a `sync.RWMutex` for this case.

internal/core/liquid.go
```diff
@@ -68,6 +101,154 @@ func (l *LiquidConnection) Init(ctx context.Context, client *gophercloud.Provide
...
+				LiquidVersion: l.LiquidServiceInfo.Version,
+				Unit:          l.LiquidServiceInfo.Rates[rateName].Unit,
+				Topology:      l.LiquidServiceInfo.Rates[rateName].Topology,
+				HasUsage:      l.LiquidServiceInfo.Rates[rateName].HasUsage,
+			}, nil
+		},
+		Update: func(rate *db.ClusterRate) (err error) {
+			rate.LiquidVersion = l.LiquidServiceInfo.Version
+			rate.Unit = l.LiquidServiceInfo.Rates[rate.Name].Unit
+			rate.HasUsage = l.LiquidServiceInfo.Rates[rate.Name].HasUsage
```
majewsky, line 236: should also update Topology

internal/core/liquid.go
```diff
@@ -68,6 +101,154 @@ func (l *LiquidConnection) Init(ctx context.Context, client *gophercloud.Provide
...
+	_, err = rateUpdate.Execute(tx)
+	if err != nil {
+		return err
+	}
+
+	err = tx.Commit()
+	if err != nil {
+		return err
+	}
+	return nil
```
majewsky, line 249:
```suggestion
	return tx.Commit()
```

internal/collector/consistency.go
```diff
@@ -18,59 +18,27 @@ import (
...
-		}
-
-		logg.Info("creating missing %s cluster service entry", serviceType)
-		err := c.DB.Insert(&db.ClusterService{Type: serviceType, NextScrapeAt: c.MeasureTime()})
-		if err != nil {
-			c.LogError(err.Error())
-		}
-	}
-
+func (c *Collector) CheckConsistencyAllDomains(_ context.Context, _ prometheus.Labels) error {
```
majewsky, line 37: Does not need to be public.

internal/test/setup.go
```diff
@@ -80,6 +81,13 @@ func WithProject(p core.KeystoneProject) SetupOption {
 	}
 }
 
+// WithCollectorSetup is a SetupOption that sets up the Cluster the same ways
+// as the limes-collect would do. Namely, this reconciles the LiquidConnections
+// on startup.
+func WithCollectorSetup(params *setupParams) {
```
majewsky, line 87: I'm not happy with the name "WithCollectorSetup", because it is not descriptive of what it actually does (only what its intended audience is), but I'm also struggling to come up with a good and succinct name. Just flagging this for now so we can bikeshed the name some more.

majewsky (changes requested):

internal/core/liquid.go
wagnerd3: I added Topology after the fact because copilot forgot Topology. Goes to show that sometimes copilot disturbs more than it helps...

internal/test/setup.go
majewsky: We have `WithEmptyRecordsAsNeeded` for the project-level stuff. I suggest renaming that one to `WithEmptyProjectLevelRecordsAsNeeded` for clarity and calling the new option `WithClusterLevelRecords`.

internal/core/liquid.go
wagnerd3: I did use a raw `RWMutex` now instead of `liquidapi.State`, is this okay like this? I figured it would make sense to lock during the whole update operation (not just on write), because if multiple scrapes (capacity, usage, rate) are happening parallel, the would trigger updates unnecessarily when they are not waiting during the ongoing update before the comparison.

internal/collector/scrape_test.go
```diff
@@ -602,15 +605,16 @@ func Test_ScrapeButNoResources(t *testing.T) {
...
 		INSERT INTO projects (id, domain_id, name, uuid, parent_uuid) VALUES (1, 1, 'berlin', 'uuid-for-berlin', 'uuid-for-germany');
 	`,
-		scrapedAt.Unix(), scrapedAt.Add(scrapeInterval).Unix(),
+		initialTime.Unix(), scrapedAt.Unix(), scrapedAt.Add(scrapeInterval).Unix(),
 	)
 }
 
-////////////////////////////////////////////////////////////////////////////////
+// //////////////////////////////////////////////////////////////////////////////
 // test for empty UsageData
```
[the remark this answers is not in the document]
wagnerd3: I recently changed my IDE to force blanks on comment start, I guess this is an exception which I forgot 😬

PR #720, opened 2025-05-21
add correct version to liquids where missing

wagnerd3:
Checklist:

@majewsky Is there any reason, why these liquids did not yet have a correct `Version` in comparison to some of the other liquids, which already have it? I noticed this during local testing, but could not find any background in the code, nor in the respective commit messages. We need this step so that we can take like the next/ upcoming part where we read the `ServiceInfo` from the database. If we don't do this, the code chunk of code will not be fully functional.

### DIFF
--- a/internal/liquids/archer/liquid.go
+++ b/internal/liquids/archer/liquid.go
@@ -5,6 +5,7 @@ package archer
 
 import (
 	"context"
+	"time"
 
 	"github.com/gophercloud/gophercloud/v2"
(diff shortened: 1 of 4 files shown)


majewsky:
No, this behavior is intentional. The ServiceInfo version is defined as incrementing at least whenever the contents of ServiceInfo change. If the contents of ServiceInfo never change, no increment is needed. The intent here is that, on change, we increment the fixed ServiceInfo version as part of the PR.

PR #728, opened 2025-05-30
merge ScrapeJob and ScrapeRatesJob

Varsius:


internal/collector/scrape.go
```diff
@@ -152,18 +155,18 @@ func (c *Collector) processResourceScrapeTask(ctx context.Context, task projectS
...
-	task.Timing.FinishedAt = c.MeasureTimeAtEnd()
 
-	// write result on success; if anything fails, try to record the error in the DB
-	if task.Err == nil {
-		err := c.writeResourceScrapeResult(dbDomain, dbProject, task, resourceData, serializedMetrics)
-		if err != nil {
-			task.Err = fmt.Errorf("while writing results into DB: %w", err)
-		}
+	// perform rate scrape
+	rateData, ratesScrapeState, err := connection.ScrapeRates(ctx, project, c.Cluster.Config.AvailabilityZones, srv.RatesScrapeState)
```
wagnerd3, line 160: I don't think the order makes sense here: If the resource scrape fails, we will try the rates anyways. But then, we return with error and don't save the other result. In this case, shouldn't we just omit the second scrape? I understand we don't want to write one of both results when the other one fails, similar to the capacity scrapes from different sources.

internal/collector/scrape_test.go
```diff
@@ -417,6 +465,41 @@ func Test_ScrapeSuccess(t *testing.T) {
...
+	serviceUsageReport.Rates["firstrate"].PerAZ["any"].Usage = Some(big.NewInt(4096))
+	serviceUsageReport.Rates["secondrate"].PerAZ["any"].Usage = Some(big.NewInt(8192))
+	serviceUsageReport.SerializedState = []byte(`{"firstrate":4096,"secondrate":8192}`)
+	mockLiquidClient.SetUsageReport(serviceUsageReport)
+
+	mustT(t, job.ProcessOne(s.Ctx, withLabel))
+
+	scrapedAt1 = s.Clock.Now().Add(-5 * time.Second)
+	scrapedAt2 = s.Clock.Now()
+	// TODO: I do not see where the changes to project_az_resources are coming from.
```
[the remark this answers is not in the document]
Varsius: We found the issue: data retention period config was missing for resource capacity

majewsky (changes requested):

internal/collector/scrape.go
```diff
@@ -71,33 +72,35 @@ var (
...
-				Help: "Counter for resource scrape operations per Keystone project.",
+				Name: "limes_scrapes",
+				Help: "Counter for scrape operations per Keystone project.",
 			},
 			CounterLabels: []string{"service_type", "service_name"},
 		},
 		DiscoverTask: func(_ context.Context, labels prometheus.Labels) (projectScrapeTask, error) {
-			return c.discoverScrapeTask(labels, findProjectForResourceScrapeQuery)
+			return c.discoverScrapeTask(labels, findProjectForScrapeQuery)
 		},
```
majewsky, line 97: We can turn this into just `DiscoverTask: c.discoverScrapeTask`. Having this indirection through an anonymous function here was used to share code between resource and rate scraping, so this can be cleaned up.

internal/collector/scrape.go
```diff
@@ -71,33 +72,35 @@ var (
...
 		Metadata: jobloop.JobMetadata{
-			ReadableName: "scrape project quota and usage",
+			ReadableName: "scrape project quota, usage and rate usage",
 			CounterOpts: prometheus.CounterOpts{
-				Name: "limes_resource_scrapes",
-				Help: "Counter for resource scrape operations per Keystone project.",
+				Name: "limes_scrapes",
+				Help: "Counter for scrape operations per Keystone project.",
 			},
 			CounterLabels: []string{"service_type", "service_name"},
```
majewsky, line 93: As mentioned in a TODO on discoverScrapeTask, the `service_name` label is deprecated. Since we're changing metric family names here anyway, we should use the opportunity to remove that label.

PR #730, opened 2025-06-12
add UUID field to commitments

majewsky:
It is a historical mistake of mine that we exposed numerical IDs on commitments in the first place. We should have UUIDs here to avoid exposing the number of commitments to non-privileged users, and as defense in depth against enumeration attacks.

Because commitments of some sort will start appearing in LIQUID soon, I want to right this wrong before it seeps into the LIQUID API. LIQUID will only ever refer to commitments by UUID, and the Limes v2 API will also move to only using UUIDs. The v1 API will show both identifiers to aid in the migration.

For tests, a facility is added to generate UUIDs in deterministic ways, similar to what we already have for commitment transfer tokens.

An additional complication presents itself in type CommitmentWorkflowContext: Currently, commitments refer to other commitments by numerical ID here. I'm adding UUIDs here on top of the existing RelatedIDs field, with the intention of eventually fully replacing RelatedIDs with RelatedUUIDs. However, we will need a migration period here: We cannot backfill existing contexts with the UUIDs of the related commitments, since some of those superseded or expired commitments have already been garbage-collected from the DB.

~~**Warning:** Please merge https://github.com/sapcc/go-api-declarations/pull/50 first, and then let me rebase this PR to include that dependency bump before approving here.~~ Done.


internal/test/setup.go
```diff
@@ -123,6 +126,25 @@ func GenerateDummyToken() string {
...
 }
 
+// GenerateDummyCommitmentUUID generates a deterministic UUID from the given ID.
+func GenerateDummyCommitmentUUID(idx uint64) db.ProjectCommitmentUUID {
+	// e.g. idx = 5
+	//   -> str = hex(sha256("5")) = "ef2d127de37b942baad06145e54b0c619a1f22327b2ebbcfbec78f5564afe39d"
+	//   -> uuid = "ef2d127d-e37b-4942-baad-06145e54b0c6"
+	buf := sha256.Sum256([]byte(strconv.FormatUint(idx, 10)))
+	str := hex.EncodeToString(buf[:])
+	uuid := fmt.Sprintf("%s-%s-4%s-%s-%s", str[0:8], str[8:12], str[13:16], str[16:20], str[20:32])
```
wagnerd3, line 136: What's the rationale of this hardcoded `4` on first place of the third block?

wagnerd3 (dismissed):
No obvious errors to spot for me.

internal/test/setup.go
majewsky: It's a v4-format UUID.

VoigtS (approved):

Nuckal777 (approved):

PR #739, opened 2025-07-03
use respondwith.ObfuscatedErrorText in 500 responses

Varsius:


internal/api/api_test.go
```diff
@@ -583,9 +583,9 @@ func Test_ProjectOperations(t *testing.T) {
 		Method:       "PUT",
 		Path:         "/rates/v1/domains/uuid-for-germany/projects/uuid-for-berlin",
 		ExpectStatus: 500, // TODO: should be 403 (I don't care about fixing this in v1; v2 will be structured differently to allow for a fix)
-		ExpectBody: assert.StringData(
-			"no such rate: shared/service/shared/notexistent:bogus\n",
-		),
+		// ExpectBody: assert.StringData(
```
Varsius, line 586: Since a 500 error is returned, the message now contains a uuid. Is it fine to just ignore the body in this test for now? @majewsky

internal/api/commitment.go
```diff
@@ -419,15 +419,15 @@ func (p *v1Provider) CreateProjectCommitment(w http.ResponseWriter, r *http.Requ
...
 		return
 	}
 	err = tx.Commit()
-	if respondwith.ErrorText(w, err) {
+	if respondwith.ObfuscatedErrorText(w, err) {
 		return
 	}
 	maybeServiceInfo, err := p.Cluster.InfoForService(loc.ServiceType)
-	if respondwith.ErrorText(w, err) {
+	if respondwith.ObfuscatedErrorText(w, err) {
```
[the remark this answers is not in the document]
majewsky: Indeed. This is a good opportunity to address this concern. Please check for cases where `respondwith.{,Obfuscated}ErrorText` is called after `tx.Commit()` and update the logic to keep going as much as possible. For example, in the line commented here, we can move the serviceInfo loading to happen before `tx.Commit()`. Then down below, when triggering a capacity scrape, an error there can just be logged without aborting the request handler. Further down, when loading the confirmed commitment, if there is an error, we should only log it and then use the previous `dbCommitment` record to display the positive API response.

internal/api/commitment.go
```diff
@@ -973,22 +977,18 @@ func (p *v1Provider) StartCommitmentTransfer(w http.ResponseWriter, r *http.Requ
...
 		}
 		dbCommitment = transferCommitment
 	}
-	err = tx.Commit()
-	if respondwith.ErrorText(w, err) {
-		return
-	}
 
 	var loc core.AZResourceLocation
 	err = p.DB.QueryRow(findAZResourceLocationByIDQuery, dbCommitment.AZResourceID).
```
majewsky, line 994: here

majewsky (approved):

PR #756, opened 2025-08-05
delegate commitment acceptance: tests working

wagnerd3:
For now, this PR is a draft because of all the TODOs and clarifications contained herein.
Still, it would be good to get reviews for the behavior which is already built into the API and tested in the tests.

Outline:
* the /commitment APIs now call a function which delegates the validation of capacity constraints to the respective liquid or uses a combined form of the old `datamodel` functions
* for the commitment conversion API, this means that the conversions can only happen within one `serviceType`, else the interface of the liquids cannot capture this change in one API request. (see TODO)
** I discussed this briefly in the office with @majewsky and we figured that in reality, there won't be any conversions e.g. from BareMetal to Nova - because they also have separate capacities
* for testing, we tell the /commitment API to call the `MockLiquidClient` by using `HandlesCommitments=true`. These requests we can then check. In order to keep the testing effort manageable, we first `assert` the `datamodel`s response and then set this result to be the liquids response. This way, the test behavior remains 1:1 while checking the interface between API and datamodel/ liquid nicely.

TODOs:
* Make commitment conversion rules only work within 1 `serviceType` (validation on startup)
* Clarify the changes necessary in the liquid API
* remove old `datamodel` code (remove internal, other usages)


majewsky (changes requested):
Comments for the low-level packages. Will continue looking at `package api` tomorrow.

internal/core/cluster.go
```diff
@@ -38,15 +38,20 @@ type Cluster struct {
...
 // NewCluster creates a new Cluster instance also initializes the LiquidConnections - if configured.
 // Errors will be logged when the requested DiscoveryPlugin cannot be found.
-func NewCluster(config ClusterConfiguration, timeNow func() time.Time, dbm *gorp.DbMap, fillLiquidConnections bool) (c *Cluster, errs errext.ErrorSet) {
+func NewCluster(config ClusterConfiguration, provider *gophercloud.ProviderClient, eo gophercloud.EndpointOpts, timeNow func() time.Time, dbm *gorp.DbMap, fillLiquidConnections bool) (c *Cluster, errs errext.ErrorSet) {
 	c = &Cluster{
 		Config:            config,
 		LiquidConnections: make(map[db.ServiceType]*LiquidConnection),
 		DB:                dbm,
+		Provider:          provider,
+		EO:                eo,
```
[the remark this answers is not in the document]
wagnerd3: Ah, thanks for that hint. This did not appear to me, I thought the differentiation was simply the `Connect()` does all steps which involve API communication. But it makes sense to stick to what you described.

internal/datamodel/confirm_project_commitments.go
```diff
@@ -69,6 +71,74 @@ func CanMoveExistingCommitment(amount uint64, loc core.AZResourceLocation, sourc
...
+	for projectUUID, projectCommitmentChangeset := range req.ByProject {
+		distinctProjectsUUIDs[projectUUID] = struct{}{}
+		for resourceName := range projectCommitmentChangeset.ByResource {
+			distinctResources[resourceName] = struct{}{}
+		}
+	}
+	// internally, we only work with projectIDs, so we have to have a conversion ready
+	projectByUUID, err := db.BuildIndexOfDBResult(
+		dbi,
+		func(project db.Project) liquid.ProjectUUID { return liquid.ProjectUUID(project.UUID) },
```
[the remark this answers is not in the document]
wagnerd3: Created https://github.com/sapcc/limes/pull/758.
But TBH, also in there we still have to have a lot of type conversions.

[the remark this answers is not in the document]
wagnerd3: Rebase was done.

wagnerd3:
This needs some interface changes before i can take the PR out of draft: https://github.com/sapcc/go-api-declarations/pull/59

internal/datamodel/confirm_project_commitments.go
```diff
@@ -35,38 +37,71 @@ var (
...
+		for projectUUID, projectCommitmentChangeset := range req.ByProject {
+			project, exists := projectByUUID[projectUUID]
+			// defense in depth: technically, the request has been validated before, so this does not happen.
+			if !exists {
+				return false, fmt.Errorf("project %s not found in database", projectUUID)
+			}
+			for _, commitment := range projectCommitmentChangeset.ByResource[resourceName].Commitments {
+				newStatus, newExists := commitment.NewStatus.Unpack()
+				oldStatus, oldExists := commitment.OldStatus.Unpack()
+				if newExists && newStatus == liquid.CommitmentStatusConfirmed && (!oldExists || oldStatus != liquid.CommitmentStatusConfirmed) {
```
majewsky, line 73: The Unpack can be avoided:
```suggestion
				if Commitment.NewStatus == Some(liquid.CommitmentStatusConfirmed) && oldStatus != Some(liquid.CommitmentStatusConfirmed) {
```
Same below.

internal/core/config.go
```diff
@@ -208,13 +208,22 @@ func (cluster ClusterConfiguration) validateConfig() (errs errext.ErrorSet) {
 	}
 
 	// NOTE: Liquids[].FixedCapacityConfiguration and Liquids[].PrometheusCapacityConfiguration are optional
+	occupiedConversionIdentifiers := make([]string, 0)
```
majewsky, line 211: A zero-length allocation is always worthless (unless you want to ensure that the slice never serializes as `null` instead of `[]` in JSON). A nil slice behaves the same as an empty one, but does not need an allocation.
```suggestion
	var occupiedConversionIdentifiers []string
```

internal/core/commitment_behavior.go
```diff
@@ -42,8 +42,14 @@ func (b CommitmentBehavior) Validate(path string) (errs errext.ErrorSet) {
 			errs.Addf("invalid value: %s.until_percent may not be bigger than 100", path)
 		}
 	}
+	if conversionRule, ok := b.ConversionRule.Unpack(); ok {
+		identifier = conversionRule.Identifier
+		if slices.Contains(occupiedConversionIdentifiers, conversionRule.Identifier) {
+			errs.Addf("invalid value: %s.conversion_rule.identifier must be unique, but %q is already used by another serviceType", path, conversionRule.Identifier)
```
majewsky, line 48: This phrasing is imprecise:
```suggestion
			errs.Addf("invalid value: %s.conversion_rule.identifier values must be restricted to a single serviceType, but %q is already used by another serviceType", path, conversionRule.Identifier)
```

internal/api/commitment.go
```diff
@@ -295,16 +311,62 @@ func (p *v1Provider) CanConfirmNewProjectCommitment(w http.ResponseWriter, r *ht
 
 	// commitments can never be confirmed immediately if we are before the min_confirm_date
 	now := p.timeNow()
-	if !behavior.CanConfirmCommitmentsAt(now) {
+	if req.ConfirmBy != nil && req.ConfirmBy.Before(now) {
```
majewsky, line 314: This is within the `can-confirm` endpoint, which the API spec defines as taking a request body same as for commitment creation, "except that the `commitment.confirm_by` attribute must not be set". So this branch cannot be taken. In fact, if we don't have this yet, there should be a check that rejects `req.ConfirmBy != nil`. Consequently, the CommitmentChangeRequest shall always try to create a commitment in status `confirmed`.

internal/api/commitment.go
```diff
@@ -385,25 +446,64 @@ func (p *v1Provider) CreateProjectCommitment(w http.ResponseWriter, r *http.Requ
...
+					},
+				},
+			},
+		},
+	}, loc.ServiceType, *serviceInfo, p.DB)
+	if respondwith.ObfuscatedErrorText(w, err) {
+		return
+	}
+
+	if confirmBy.IsNone() {
```
majewsky, line 494: I think it would be safer to evaluate `commitmentRequest.RequiresConfirmation()` here, in order to avoid subtle drifts between when Limes expects a confirmation vs. when LIQUID mandates it (in case the RequiresConfirmation logic is adjusted in the future).

majewsky (changes requested):
I managed to start reading into the commitment API, but did not get through all (or the tests) before running out of time yet again...

internal/api/commitment.go
```diff
@@ -1649,6 +2058,51 @@ func (p *v1Provider) UpdateCommitmentDuration(w http.ResponseWriter, r *http.Req
...
+						TotalGuaranteedBefore: 0,
+						TotalGuaranteedAfter:  0,
+						Commitments: []liquid.Commitment{
+							{
+								UUID:      liquid.CommitmentUUID(dbCommitment.UUID),
+								OldStatus: Some(p.convertCommitmentStateToDisplayForm(dbCommitment.State)),
+								NewStatus: Some(p.convertCommitmentStateToDisplayForm(dbCommitment.State)),
+								Amount:    dbCommitment.Amount,
+								ConfirmBy: dbCommitment.ConfirmBy,
+								ExpiresAt: dbCommitment.ExpiresAt,
```
majewsky, line 2079: This should be setting OldExpiresAt.

internal/api/commitment.go
```diff
@@ -1668,3 +2122,123 @@ func (p *v1Provider) UpdateCommitmentDuration(w http.ResponseWriter, r *http.Req
...
+		return None[liquid.ProjectMetadata]()
+	}
+	return Some(liquid.ProjectMetadata{
+		UUID: string(dbProject.UUID),
+		Name: dbProject.Name,
+		Domain: liquid.DomainMetadata{
+			UUID: domain.UUID,
+			Name: domain.Name,
+		},
+	})
```
majewsky, line 2233:
```suggestion
	return Some(core.KeystoneProjectFromDB(dbProject, domain).ForLiquid())
```

internal/api/commitment.go
```diff
@@ -1668,3 +2122,123 @@ func (p *v1Provider) UpdateCommitmentDuration(w http.ResponseWriter, r *http.Req
...
+		return liquid.CommitmentChangeResponse{
+			RejectionReason: "not enough capacity!",
+			RetryAt:         None[time.Time](),
+		}, nil
+	}
+
+	return result, nil
+}
+
+func LiquidProjectMetadataFromDBProject(dbProject db.Project, domain db.Domain, serviceInfo liquid.ServiceInfo) Option[liquid.ProjectMetadata] {
```
majewsky, line 2222: Exported functions should have a docstring. But since these are small helpers for the current file only, they can be made private:
```suggestion
func liquidProjectMetadataFromDBProject(dbProject db.Project, domain db.Domain, serviceInfo liquid.ServiceInfo) Option[liquid.ProjectMetadata] {
```

Same for EvaluateRetryHeader below.

majewsky (changes requested):
I read through `internal/api/commitment.go`

internal/api/commitment.go
```diff
@@ -981,32 +1272,11 @@ func (p *v1Provider) StartCommitmentTransfer(w http.ResponseWriter, r *http.Requ
...
-	serviceInfo, ok := maybeServiceInfo.Unpack()
-	if !ok {
-		http.Error(w, "service not found", http.StatusNotFound)
-		return
+		dbCommitment = transferCommitment
 	}
-
 	err = tx.Commit()
-	if respondwith.ObfuscatedErrorText(w, err) {
+	if respondwith.ErrorText(w, err) {
```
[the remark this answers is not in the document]
wagnerd3: There was many places where this applied, I replaced all of them now.

internal/api/commitment.go
wagnerd3: I am so 👌🏻 close to building a lint for this ^^

internal/api/commitment.go
```diff
@@ -1668,3 +2118,123 @@ func (p *v1Provider) UpdateCommitmentDuration(w http.ResponseWriter, r *http.Req
 
 	respondwith.JSON(w, http.StatusOK, map[string]any{"commitment": c})
 }
+
+// DelegateCommitmentCheck decides whether LiquidClient.ChangeCommitments() should be called,
```
majewsky, line 2122:
```suggestion
// DelegateChangeCommitments decides whether LiquidClient.ChangeCommitments() should be called,
```

internal/api/commitment.go
```diff
@@ -1668,3 +2118,123 @@ func (p *v1Provider) UpdateCommitmentDuration(w http.ResponseWriter, r *http.Req
...
+	}
+	for projectUUID, projectCommitmentChangeset := range req.ByProject {
+		for resourceName, resourceCommitmentChangeset := range projectCommitmentChangeset.ByResource {
+			// this is just to make the tests deterministic because time.Local != local IANA time (after parsing)
+			for i, commitment := range resourceCommitmentChangeset.Commitments {
+				commitment.ExpiresAt = commitment.ExpiresAt.Local()
+				confirmBy, exists := commitment.ConfirmBy.Unpack()
+				if exists {
+					commitment.ConfirmBy = Some(confirmBy.Local())
+				}
```
majewsky, line 2148:
```suggestion
				commitment.ConfirmBy = options.Map(commitment.ConfirmBy, time.Time.Local)
```
I recently learned that you can refer to a method that way.

internal/api/commitment.go
```diff
@@ -1668,3 +2118,123 @@ func (p *v1Provider) UpdateCommitmentDuration(w http.ResponseWriter, r *http.Req
...
+	} else {
+		liquidClient = c.LiquidConnections[serviceType].LiquidClient
+	}
+	commitmentChangeResponse, err := liquidClient.ChangeCommitments(ctx, remoteCommitmentChanges)
+	if err != nil {
+		return result, fmt.Errorf("failed to retrieve liquid ChangeCommitment response for service %s: %w", serviceType, err)
+	}
+	if commitmentChangeResponse.RejectionReason != "" {
+		return commitmentChangeResponse, nil
+	}
```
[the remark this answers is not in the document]
wagnerd3: I added a test to check we don't do unnecessary requests whenever the commitments are not managed by the liquid.

Nuckal777 (approved):

PR #757, opened 2025-08-05
add cluster_resources.path and cluster_az_resources.path

majewsky:
For resources, this contains "servicetype/resourcename".
For AZ resources, this contains "servicetype/resourcename/az".

The original intent with this is to make interactive queries easier: When selecting on `project_az_resources`, one will only need to join `projects` and `cluster_az_resources` and then match on `projects.uuid` and `cluster_az_resources.path`, without having to join `cluster_services` and `cluster_resources` as well.

While updating the test assertions, I noticed that it also makes assertions more readable because each line in the SQL diff contains a full identification of the respective resource or AZ resource.

Also, following https://github.com/sapcc/go-makefile-maker/pull/335, this removes the workaround for that one dupword bug.

### DIFF
--- a/internal/collector/capacity_scrape_test.go
+++ b/internal/collector/capacity_scrape_test.go
@@ -641,11 +641,10 @@ func Test_ScanCapacityButNoResources(t *testing.T) {
 		INSERT INTO cluster_az_resources (id, resource_id, az, raw_capacity, path) VALUES (3, 1, 'az-two', 0, 'shared/capacity/az-two');
 		INSERT INTO cluster_az_resources (id, resource_id, az, raw_capacity, path) VALUES (4, 1, 'unknown', 0, 'shared/capacity/unknown');
 		INSERT INTO cluster_az_resources (id, resource_id, az, raw_capacity, path) VALUES (5, 2, 'any', 0, 'shared/things/any');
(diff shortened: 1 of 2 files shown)


.golangci.yaml
```diff
@@ -71,6 +71,10 @@ linters:
     - usetesting
     - whitespace
   settings:
+    dupword:
+      # Do not choke on SQL statements like `INSERT INTO things (foo, bar, baz) VALUES (TRUE, TRUE, TRUE)`.
+      # And yes, dupword considers the comma to be part of the word, because... I don't know.
+      ignore: [ "TRUE", "TRUE,", "FALSE", "FALSE,", "NULL", "NULL," ]
```
wagnerd3, line 77: You should be able to only add the version without `,`:
```suggestion
      ignore: [ "TRUE", "FALSE", "NULL" ]
```

wagnerd3 (commented):
Do we want to enforce this special constraint `path='{{TYPE}}/{{NAME}}` (or additionally with `/{{AZ}}`) through an insert-update-trigger? :-) As `cluster_services, cluster_resources and cluster_az_resources` change rather seldom, I think it would not hurt the performance. It adds an extra layer of security and makes the schema more understandable if you just look at the database. It would make me feel better for the intention to update joins to use the `path` for identifying entries.

.golangci.yaml
majewsky: Indeed. We can fix that in go-makefile-maker.

majewsky:
@wagnerd3 I added a trigger. Please take a look.

wagnerd3 (approved):

PR #760, opened 2025-08-07
rollup DB migrations

majewsky:
To better represent the current state of the DB. Also, this moves the baseline to a separate file, so that my code editor shows SQL syntax highlighting. :) I verified that running

```sh
rm -rf .testdb
go test ./internal/core
.testdb/pg_dump.sh -d test_clustersaveserviceinfo --schema-only
```

before and after this change produces no difference in output.


VoigtS (approved):
I can confirm the results.

PR #767, opened 2025-08-19
remove liquidServiceType (only used for testing), attempt 2

majewsky:
I have stacked some changes on top of #766 to not block tests from progressing in parallel because they might share the same service types for their mock liquids.


wagnerd3 (approved):
That's a simple and functional solution, to put the registry into the test setup. Very nice, thank you! :-)

Varsius (approved):

PR #768, opened 2025-08-22
liquid-ironic: limit memory usage when listing instances

majewsky:
There is so much crap in the JSON returned by Nova /servers/detail that it reliably blows over our 128 MiB memory limit in prod regions with large customer projects (one project has nearly 1k baremetal nodes).

Usually, I dislike pagination, because OpenStack is rather bad at implementing pagination well. But I'm not increasing that memory limit. More than that for a single liquid is preposterous.

Tested against eu-de-1:
```
$ limesctl liquid report-usage ironic $BIG_PROJECT_UUID -c -e http://localhost:8180/
ServiceUsageReports are identical
```


SchwarzM (approved):
lgtm

SuperSandro2000 (approved):

PR #769, opened 2025-08-25
internal/core: fix rates records created from configured rates

majewsky:
We only relied on the RateInfo returned by LIQUID when creating `rates` records, but for rates that are only configured (i.e. limit only and no usage), we do not have a RateInfo. In this case, we're now putting in UnitNone and FlatTopology as reasonable fallbacks. If anything else needs to be supported in the future, we need to extend the config format.


Varsius (approved):

SuperSandro2000 (approved):
geheiligt sei der Prozess

PR #771, opened 2025-08-29
the one wherein I started refactoring tests and it never ended send help

majewsky:
It appears that I have decided to join in on the nascent tradition of big huge PRs. This was not originally a conscious choice; I just kept going at this because nobody was around to review it in smaller pieces, and now here we are. Right out of the gate, I want to stress that I have been working very hard to separate all these changes out into small meaningful commits (as can be seen from the highly nonlinear order of commit timestamps on the branch). When reviewing, please take advantage of my efforts and review each commit diff separately.

The one thing that this branch does not do is the thing that I initially set out to do: Following the ground work in https://github.com/sapcc/go-api-declarations/pull/56, I was going to create `az_resources` and `project_az_resources` records for `liquid.AvailabilityZoneTotal` and move aggregate values from the resource level down into the AZ resource level, which would allow simplifying several pieces of business logic and expensive report queries.

The problem with that is that adding additional `{project_,}az_resources` records shifts existing IDs around and generates a lot of diffs in the test fixtures and assertions. This is especially problematic with setup-time DB fixtures (think `internal/api/fixtures/start-data.sql`) because those are 100% hand-written with no easy shortcut to automate updating them accordingly. While even just looking at those fixtures to judge the effort of updating them to include `az=total`, I saw how utterly inconsistent they were in many cases with how we are structuring our DB these days. For example, most `az=unknown` records were entirely missing, and sometimes we didn't even bother filling in records for legitimate AZs if we did not care about the respective records enough.

In order to get to a point where I could add `az=total` without having the respective commit balloon completely out of control, or having things blow up out of nowhere with no rhyme or reason, I decided to clean up the test setup phases. That is what is happening in this PR. I did try to limit myself to only test refactorings that contributed towards the grand goal of making the addition of `az=total` as small of a diff as reasonably possible, but I also ended up fixing a bunch of particularly abhorrent inconsistencies in the test setups as I happened across them. This is particularly true for the first third of commits on this branch, where you will notice that several commit messages talk about how cleaning up that one weird test setup uncovered a bug in the implementation that the old inconsistent test setup somehow side-stepped.

For reference, my actual change to introduce `az=total` is [in this other branch here](https://github.com/sapcc/limes/compare/db-test-refactors...az-total). I will put this up as a PR once this one is merged.


SuperSandro2000:
The report is here https://github.com/sapcc/limes/actions/runs/17320825689?pr=771

SuperSandro2000 (approved):

PR #778, opened 2025-09-22
consume publicly posted transfer commitments automatically

wagnerd3:
What this can do now:
* automatically consume `transfer_status=public` commitments
* send transfer mails, collated per project
* prevent confirmation mails, for commitments which were transferred after
* note the chart PR for the mail template: https://github.com/sapcc/helm-charts/pull/9831
* put audit entries for commitment confirmation and transfer

TODO:
* follow-up PR: Send mails when certain operations fail
* follow-up PR: Add `CommitmentChangeRequest` to liquid

### DIFF
--- a/internal/collector/capacity_scrape_test.go
+++ b/internal/collector/capacity_scrape_test.go
@@ -1530,7 +1530,7 @@ func Test_ScanCapacityWithCommitmentTakeover(t *testing.T) {
 		ProjectID:    dresden,
 		AZResourceID: firstCapacityAZOne,
 		Amount:       3,
-		CreatedAt:    s.Clock.Now().Add(-1 * time.Hour), // before the transfer commitment,
+		CreatedAt:    creation.Add(-1 * time.Hour), // before the transfer commitment
 		Duration:     committedForTwentyDays,
 	})
 	tr.DBChanges().Ignore()
@@ -1642,7 +1642,7 @@ func Test_ScanCapacityWithCommitmentTakeover(t *testing.T) {
 	s.Clock.StepBy(24 * time.Hour)
 	mustT(t, jobloop.ProcessMany(job, s.Ctx, len(s.Cluster.LiquidConnections)))
 
-	// now uuid17 is confirmed - as the confirmation of uuid18 is later, nothing else happens
+	// now uuid16 is confirmed - as the confirmation of uuid18 is later, nothing else happens
 	now = s.Clock.Now().Add(-5 * time.Second)
 	confirmation = now
 	tr.DBChanges().AssertEqualf(`
@@ -1655,11 +1655,13 @@ func Test_ScanCapacityWithCommitmentTakeover(t *testing.T) {
 	s.Clock.StepBy(24 * time.Hour)
 	mustT(t, jobloop.ProcessMany(job, s.Ctx, len(s.Cluster.LiquidConnections)))
 
-	// now the time progresses, uuid18 becomes pending and takes over amount=2 from uuid17 --> quota in berlin reduces by 2
-	// a leftover for amount=1 is created for the not-taken-over part of uuid17
+	// now the time progresses, uuid17 becomes pending and takes over amount=2 from uuid16 --> quota in berlin reduces by 2
+	// a leftover for amount=1 is created for the not-taken-over part of uuid16
 	now = s.Clock.Now().Add(-5 * time.Second)
 	confirmation2 := now
 	creation3 := now
+
+	uuid19 := test.GenerateDummyCommitmentUUID(19)
 	tr.DBChanges().AssertEqualf(`
 		UPDATE project_az_resources SET quota = 18 WHERE id = 12 AND project_id = 2 AND az_resource_id = 2;
 		UPDATE project_az_resources SET quota = 3 WHERE id = 2 AND project_id = 1 AND az_resource_id = 2;
@@ -1670,7 +1672,7 @@ func Test_ScanCapacityWithCommitmentTakeover(t *testing.T) {
 		UPDATE project_resources SET quota = 3 WHERE id = 1 AND project_id = 1 AND resource_id = 1;
 		UPDATE project_resources SET quota = 19 WHERE id = 5 AND project_id = 2 AND resource_id = 1;
 		%[10]s
-	`, now.Unix(), creation.Unix(), transferStartedAt.Unix(), confirmBy.Unix(), confirmation.Unix(), expiry.Unix(), uuid16, uuid17, test.GenerateDummyCommitmentUUID(19), timestampUpdates())
+	`, now.Unix(), creation.Unix(), transferStartedAt.Unix(), confirmBy.Unix(), confirmation.Unix(), expiry.Unix(), uuid16, uuid17, uuid19, timestampUpdates())
 
 	s.Clock.StepBy(24 * time.Hour)
 	mustT(t, jobloop.ProcessMany(job, s.Ctx, len(s.Cluster.LiquidConnections)))
@@ -1680,6 +1682,7 @@ func Test_ScanCapacityWithCommitmentTakeover(t *testing.T) {
 	now = s.Clock.Now().Add(-5 * time.Second)
 	creation2 := now
 
+	uuid20 := test.GenerateDummyCommitmentUUID(20)
 	tr.DBChanges().AssertEqualf(`
 		UPDATE project_az_resources SET quota = 17 WHERE id = 12 AND project_id = 2 AND az_resource_id = 2;
 		UPDATE project_az_resources SET quota = 1 WHERE id = 22 AND project_id = 3 AND az_resource_id = 2;
@@ -1691,7 +1694,7 @@ func Test_ScanCapacityWithCommitmentTakeover(t *testing.T) {
 		UPDATE project_resources SET quota = 1 WHERE id = 9 AND project_id = 3 AND resource_id = 1;
 		UPDATE project_services SET quota_desynced_at = %[1]d WHERE id = 5 AND project_id = 3 AND service_id = 1;
 		%[10]s
-	`, now.Unix(), creation.Unix(), transferStartedAt2.Unix(), confirmBy2.Unix(), confirmation2.Unix(), expiry2.Unix(), uuid17, uuid18, test.GenerateDummyCommitmentUUID(20), timestampUpdates())
+	`, now.Unix(), creation.Unix(), transferStartedAt2.Unix(), confirmBy2.Unix(), confirmation2.Unix(), expiry2.Unix(), uuid17, uuid18, uuid20, timestampUpdates())
 
 	s.Clock.StepBy(24 * time.Hour)
 	mustT(t, jobloop.ProcessMany(job, s.Ctx, len(s.Cluster.LiquidConnections)))
@@ -1725,7 +1728,7 @@ func Test_ScanCapacityWithCommitmentTakeover(t *testing.T) {
 		UPDATE project_resources SET quota = 17 WHERE id = 5 AND project_id = 2 AND resource_id = 1;
 		UPDATE project_resources SET quota = 2 WHERE id = 9 AND project_id = 3 AND resource_id = 1;
 		%[10]s
-	`, now.Unix(), creation2.Unix(), transferStartedAt2.Unix(), confirmBy2.Unix(), confirmation2.Unix(), expiry2.Unix(), uuid17, test.GenerateDummyCommitmentUUID(20), uuid21, timestampUpdates())
+	`, now.Unix(), creation2.Unix(), transferStartedAt2.Unix(), confirmBy2.Unix(), confirmation2.Unix(), expiry2.Unix(), uuid17, uuid20, uuid21, timestampUpdates())
 
 	// now we add one commitment larger than the rest of uuid16 (which is uuid19), consuming it fully
 	uuid22 := add(db.ProjectCommitment{
@@ -1751,7 +1754,7 @@ func Test_ScanCapacityWithCommitmentTakeover(t *testing.T) {
 		UPDATE project_resources SET quota = 2 WHERE id = 1 AND project_id = 1 AND resource_id = 1;
 		UPDATE project_resources SET quota = 4 WHERE id = 9 AND project_id = 3 AND resource_id = 1;
 		%[10]s
-	`, now.Unix(), creation3.Unix(), transferStartedAt.Unix(), confirmBy.Unix(), confirmation.Unix(), expiry.Unix(), uuid16, test.GenerateDummyCommitmentUUID(19), uuid22, timestampUpdates())
+	`, now.Unix(), creation3.Unix(), transferStartedAt.Unix(), confirmBy.Unix(), confirmation.Unix(), expiry.Unix(), uuid16, uuid19, uuid22, timestampUpdates())
 }
 
 func TestScanCapacityWithMailNotification(t *testing.T) {


wagnerd3:
@majewsky  I just realized that the current state of the PR does not yet tackle the mail sending for takeover of commitments (uncollated, as discussed in the ticket). I will therefore keep this PR on draft until I tackled that. 
The logic for commitment takeover itself could be reviewed anyways - it should be sufficiently covered by tests already.

majewsky (changes requested):
I read through `core`, `db`, `test`, `util` and most of `datamodel` so far. The notification part of `datamodel`, as well as `api` and `collector` are TBD.

internal/datamodel/confirm_project_commitments.go
```diff
@@ -171,43 +280,106 @@ func ConfirmPendingCommitments(loc core.AZResourceLocation, cluster *core.Cluste
 		}
 	}
 
+	// remove duplicates of multiple consecutive transferred commitments
+	transferredCommitmentIDsWithNotification := make(map[db.ProjectID]map[db.ProjectCommitmentID]core.CommitmentTransferLeftover)
+	for projectID, transfersPerProject := range transferredCommitmentIDs {
+		if _, exists := transferredCommitmentIDsWithNotification[projectID]; !exists {
+			transferredCommitmentIDsWithNotification[projectID] = make(map[db.ProjectCommitmentID]core.CommitmentTransferLeftover)
+		}
```
majewsky, line 288: I feel like this logic can be simplified. This loop goes over all relevant projects one by one, and then the next loop (where the actual notifs are generated) does the same, except for a larger set of projects.

I would like to ask for those loops to be merged, with the entire loop going over the union of projects with confirmations and transfers (like the second loop does now), on the condition that this actually ends up simplifying the code.

For example, then we do not need this double-layered map here. We would only need to hold a simple `map[db.ProjectCommitmentID]core.CommitmentTransferLeftover` within that loop.

Also, re `transferredCommitmentIDsWithNotification`: That's a mouthful. How about `notifiableTransfers`?

internal/datamodel/confirm_project_commitments.go
```diff
@@ -171,43 +280,106 @@ func ConfirmPendingCommitments(loc core.AZResourceLocation, cluster *core.Cluste
...
+		// we go through the transfers by ID descending, because that enables the linking operation in O(n)
+		for _, cID := range slices.Backward(slices.Sorted(maps.Keys(transfersPerProject))) {
+			// for commitments which get confirmed first and then transferred, we remove the mail for confirmation
+			committedPerProject := confirmedCommitmentIDs[projectID]
+			idx := slices.Index(committedPerProject, cID)
+			if idx != -1 {
+				confirmedCommitmentIDs[projectID] = append(confirmedCommitmentIDs[projectID][:idx], confirmedCommitmentIDs[projectID][idx+1:]...)
```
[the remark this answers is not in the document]
majewsky: Actually there's a better way to do this.

internal/datamodel/confirm_project_commitments.go
```diff
@@ -171,43 +280,106 @@ func ConfirmPendingCommitments(loc core.AZResourceLocation, cluster *core.Cluste
...
+			}
```
majewsky, line 296:
```suggestion
			committedPerProject = slices.DeleteFunc(committedPerProject, func(id db.ProjectCommitmentID) bool { return id == cID })
			confirmedCommitmentIDs[projectID] = committedPerProject
```

internal/datamodel/confirm_project_commitments.go
```diff
@@ -171,43 +280,106 @@ func ConfirmPendingCommitments(loc core.AZResourceLocation, cluster *core.Cluste
...
-	err := dbi.QueryRow("SELECT d.name, p.name FROM domains d JOIN projects p ON d.id = p.domain_id where p.id = $1", projectID).Scan(&n.DomainName, &n.ProjectName)
+	err = dbi.QueryRow("SELECT d.name, p.name FROM domains d JOIN projects p ON d.id = p.domain_id where p.id = $1", projectID).Scan(&n.DomainName, &n.ProjectName)
 	if err != nil {
-		return db.MailNotification{}, err
+		return result, err
 	}
 
-	var commitments []db.ProjectCommitment
-	_, err = dbi.Select(&commitments, `SELECT * FROM project_commitments WHERE id = ANY($1)`, pq.Array(confirmedCommitmentIDs))
+	commitmentsByID, err := db.BuildIndexOfDBResult(dbi, func(c db.ProjectCommitment) db.ProjectCommitmentID { return c.ID }, `SELECT * FROM project_commitments WHERE id = ANY($1)`, pq.Array(append(confirmedCommitmentIDs, slices.Collect(maps.Keys(transferredCommitmentIDs))...)))
```
majewsky, line 338:
```suggestion
	commitmentsByID, err := db.BuildIndexOfDBResult(dbi, func(c db.ProjectCommitment) db.ProjectCommitmentID { return c.ID }, `SELECT * FROM project_commitments WHERE id = ANY($1)`, pq.Array(slices.AppendSeq(confirmedCommitmentIDs, maps.Keys(transferredCommitmentIDs))))
```

internal/datamodel/confirm_project_commitments.go
```diff
@@ -171,43 +280,106 @@ func ConfirmPendingCommitments(loc core.AZResourceLocation, cluster *core.Cluste
...
+		})
+	}
+	if len(n.Commitments) == 0 {
+		return result, nil
+	}
+	newNotification, err := tplConfig.TransferredCommitments.Render(n, projectID, now)
+	if err != nil {
+		return result, err
+	}
+	result = append(result, newNotification)
```
majewsky, line 383: I know (and agree with) the notion of returning early, but in this case, I got confused because this looked different from L354-L360 above where the other notification is rendered. Since it is the same, I would like for it to look the same:
```suggestion
	if len(n.Commitments) != 0 {
		newNotification, err := tplConfig.TransferredCommitments.Render(n, projectID, now)
		if err != nil {
			return result, err
		}
		result = append(result, newNotification)
	}
```

majewsky (changes requested):
So I wrapped my head around the datamodel algorithm now, but there is too much stuff distracting me to go further rn.

internal/collector/capacity_scrape_test.go
```diff
@@ -1166,21 +1167,493 @@ func TestScanCapacityWithMailNotification(t *testing.T) {
...
+					UPDATE services SET scraped_at = %d, next_scrape_at = %d WHERE id = 1 AND type = 'first' AND liquid_version = 1;
+					UPDATE services SET scraped_at = %d, next_scrape_at = %d WHERE id = 2 AND type = 'second' AND liquid_version = 1;
+				`,
+			scrapedAt1.Unix(), scrapedAt1.Add(15*time.Minute).Unix(),
+			scrapedAt2.Unix(), scrapedAt2.Add(15*time.Minute).Unix(),
+		))
+	}
+
+	// fill `project_az_resources` with some usage data
+	// (we want to see how commitment confirmation reacts to existing usage)
```
majewsky, line 1195: This does not do what the comment says. Copy-paste error?

internal/collector/capacity_scrape_test.go
```diff
@@ -1166,21 +1167,493 @@ func TestScanCapacityWithMailNotification(t *testing.T) {
...
+		UPDATE project_commitments SET status = 'confirmed', confirmed_at = %[1]d WHERE id = 2 AND uuid = '%[3]s' AND transfer_token = NULL;
+		UPDATE project_resources SET quota = 1 WHERE id = 1 AND project_id = 1 AND resource_id = 1;
+		UPDATE project_resources SET quota = 1 WHERE id = 5 AND project_id = 2 AND resource_id = 1;
+		UPDATE project_services SET quota_desynced_at = %[1]d WHERE id = 1 AND project_id = 1 AND service_id = 1;
+		UPDATE project_services SET quota_desynced_at = %[1]d WHERE id = 3 AND project_id = 2 AND service_id = 1;
+		%[4]s
+	`, now.Unix(), UUID1, UUID2, timestampUpdates())
+
+	// now we place a commitment that is in the same project, so it cannot be consumed by the transfer one
+	// this situation can only occur, if the customer wants to get rid of a commitment, which is older than one they placed later
```
majewsky, line 1249: The first sentence is phrased the wrong way around. In the second sentence, "can only occur" is misleading. The sentence describes the specific problematic situation where this arrangement could arise.
```suggestion
	// now we place a commitment that is in the same project, so it cannot be consume the transferable one;
	// this checks that we avoid the loophole where the customer wants to get rid of an
	// old undeleteable commitment by having it be consumed by a newer deleteable one
```

majewsky (changes requested):
I got through the ConfirmPendingCommitments tests. 🎉 Only the API is left to go.

internal/datamodel/confirm_project_commitments.go
```diff
@@ -29,14 +31,32 @@ var (
...
+	// very small due to the atomicity of the API operation.
+	getTransferCommitmentsQuery = sqlext.SimplifyWhitespace(db.ExpandEnumPlaceholders(`
+		SELECT pc.*
+		FROM services cs
+		JOIN resources cr ON cr.service_id = cs.id
+		JOIN az_resources azr ON azr.resource_id = cr.id
+		JOIN project_commitments pc ON pc.az_resource_id = azr.id
+		WHERE cs.type = $1 AND cr.name = $2 AND azr.az = $3 AND pc.transfer_status = {{limesresources.CommitmentTransferStatusPublic}}
+		ORDER BY pc.transfer_started_at ASC, pc.created_at ASC, pc.id ASC
+	`))
```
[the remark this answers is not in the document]
wagnerd3: Checked with the regex, it was only this occasion:
```
➜ rg -w 'c(?:ar|azr|s|r|ra)' -l | grep -v vendor
internal/db/baseline.sql
internal/db/migrations.go
```

internal/core/mail.go
```diff
@@ -30,11 +30,17 @@ type AZResourceLocation struct {
 	AvailabilityZone limes.AvailabilityZone
 }
 
+type CommitmentTransferLeftover struct {
+	Amount uint64
+	ID     db.ProjectCommitmentID
```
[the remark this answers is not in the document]
wagnerd3: The `ID` is only being used internally to identify the commitment. If we would expose the `UUID` to the customer in the mail, we should also expose the `UUID` for all other notification mails on commitments, which is not being done right now. So I will hold off on this for now.

Instead, we just recognized that it would probably make sense to write this whole information about the automatic commitment takeover into the audit trail. I will include this directly in the PR.

internal/datamodel/confirm_project_commitments.go
```diff
@@ -136,31 +146,130 @@ func ConfirmPendingCommitments(loc core.AZResourceLocation, cluster *core.Cluste
...
+
+		overallTransferredAmount := uint64(0)
+		for idx, t := range transferableCommitments {
+			if overallTransferredAmount == c.Amount {
+				break
+			}
+			transfersPerProject, exists := transferredCommitmentIDs[t.ProjectID]
+			if _, exists2 := transfersPerProject[t.ID]; t.ProjectID == c.ProjectID || t.ExpiresAt.After(c.ExpiresAt) || (exists && exists2) {
+				continue
+			}
```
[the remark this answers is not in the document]
wagnerd3: I took this over, but placed the comments above the if-block. It looks better to me, because you would e.g. also explain a loop above, and not inside. 
Is that something we have a common opinion on in the group?

internal/db/models.go
```diff
@@ -218,11 +220,13 @@ type CommitmentWorkflowContext struct {
...
-	CommitmentReasonConvert CommitmentReason = "convert"
-	CommitmentReasonMerge   CommitmentReason = "merge"
-	CommitmentReasonRenew   CommitmentReason = "renew"
+	CommitmentReasonCreate   CommitmentReason = "create"
+	CommitmentReasonSplit    CommitmentReason = "split"
+	CommitmentReasonConvert  CommitmentReason = "convert"
+	CommitmentReasonMerge    CommitmentReason = "merge"
+	CommitmentReasonRenew    CommitmentReason = "renew"
+	CommitmentReasonTransfer CommitmentReason = "transfer"
+	CommitmentReasonConfirm  CommitmentReason = "confirm"
```
majewsky: This reason does not make sense within the context of this enum. The enum is defined as appearing in CommitmentWorkflowContext, and this struct is used in CreationContext, SupersedeContext and RenewContext. Continued in the next comment...

internal/util/resource_behavior.go
majewsky: Why are types ResourceBehavior and TranslationRule being moved? My guess is that there was an import cycle because of a new reference, but I cannot find which reference would cause such a problem.

majewsky (changes requested):
A few basic comments:

internal/datamodel/commitment.go
```diff
@@ -0,0 +1,118 @@
...
+	// UUID generation will only raise an error if reading from /dev/urandom fails,
+	// which is a wildly unexpected OS-level error and thus fine as a fatal error
+	return liquid.CommitmentUUID(must.Return(uuid.NewV4()).String())
+}
+
+// BuildSplitCommitment prepares a new commitment instance whose creation context
+// indicates that it was split from the given existing commitment. It is used in
+// the implementation of various API endpoints that can implicitly split commitments
+// if necessary.
+func BuildSplitCommitment(dbCommitment db.ProjectCommitment, amount uint64, now time.Time, generateProjectCommitmentUUID func() liquid.CommitmentUUID) (db.ProjectCommitment, error) {
```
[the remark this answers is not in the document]
wagnerd3: moved

wagnerd3:
After testing more thoroughly, I realized there is one more problem with my current implementation: Immediately confirmed commitments skip the transfer commitment consumption. I need to make the code which lived in `confirm_project_commitments.go` into a reusable separate struct, which has an interface that fits to the "bulk confirmation" logic of the loop but also to the single commitment fashion of the `POST /commitment` API.

docs/users/api-spec-resources.md
```diff
@@ -652,6 +652,18 @@ The response is a JSON of the commitment including the following fields that ide
...
 ```
+With the following request body, the endpoint can be used to withdraw a commitment from transfer.
+```json
+{
+  "commitment": {
+    "transfer_status": ""
+  }
+}
+```
+This operation only works on commitments that are either a) set to `transfer_status=unlisted` or b) set to `transfer_status=public` _and_ the point of posting was less than 24h in the past _and_ the commitment was not transferred yet.
```
majewsky: I recall we removed this requirement because we could not remember why we had it in the first place. I can see no reference in the implementation to it.
```suggestion
This operation only works on commitments that were not transferred yet. 
```

majewsky (commented):
I have read through the remaining piece, the algorithms in package datamodel. Once again, I have committed some nitpick changes directly, and the comments below are only for substantial comments. Once these (as well as the other outstanding) comments are addressed, I feel confident merging this.

On a related note, this has now been my 3rd or 4th time reading the confirmation/takeover algorithm, and I was happy to see that it felt easier to digest this time around. It's still pretty gnarly, but it feels like we got closer to dealing with just the essential complexity of the domain. As much as this changeset was/is a chore for all of us, I appreciate the good results.

internal/api/commitment_test.go
```diff
@@ -1283,15 +1359,77 @@ func Test_StartCommitmentTransfer(t *testing.T) {
...
+	delete(resp1, "transfer_status")
+	delete(resp1, "transfer_token")
+	assert.HTTPRequest{
+		Method:       "POST",
+		Path:         "/v1/domains/uuid-for-germany/projects/uuid-for-berlin/commitments/1/start-transfer",
+		ExpectStatus: http.StatusAccepted,
+		ExpectBody:   assert.JSONObject{"commitment": resp1},
+		Body:         assert.JSONObject{"commitment": assert.JSONObject{"amount": 10, "transfer_status": ""}},
+	}.Check(t, s.Handler)
+	s.Clock.StepBy(-25 * time.Hour)
```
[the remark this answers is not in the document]
wagnerd3: Correct, this situation is not necessary for proper coverage anymore. Will remove it.

internal/datamodel/consume_transferable_commitments.go
```diff
@@ -0,0 +1,309 @@
...
+	"github.com/sapcc/go-bits/sqlext"
+
+	"github.com/sapcc/limes/internal/audit"
+	"github.com/sapcc/limes/internal/core"
+	"github.com/sapcc/limes/internal/db"
+)
+
+// Commitments with transfer_status=public are selected to be matched with confirmable
+// commitments in the order they were posted in for transfer. The status of a
+// transferable commitment does not matter for this operation.
```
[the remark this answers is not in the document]
wagnerd3: `/api/commitment.go:1256` has a check for this on API level, but we just discussed that a check on DB level would be good to catch all the cases where something gets superseded which is in `transfer_status != ''`. Probably, I will do as mentioned and take a small action item for this.

wagnerd3:
After hours of rebasing, I now have a somewhat nice history which has the main iteration steps of the algorithm separated after each other, but I backported the correction of where which function goes (`util, datamodel, core`) into the original commits. This way, I could get rid of a lot of "added this, removed it X commits later".
@majewsky ready for another round.

PR #786, opened 2025-10-22
replace yaml with json in configs

Varsius:
Checklist:


docs/operators/config.md
```diff
@@ -325,24 +370,46 @@ Some special behaviors for resources can be configured in the `resource_behavior
...
+      "identity_in_v1_api": "cinder/$1"
+    },
+    {
+      // this service contains sets of resources for each share type
+      "resource": "manila/(shares|snapshots|share_capacity|snapshot_capacity)_(.+)",
+      "category": "share_type_$2"
+    }
+  ]
+}
 ```
```
SuperSandro2000: GitHub should understand this for rendering https://github.com/github-linguist/linguist/blob/main/samples/JSON5/package.json5

````suggestion
```json5
{
  "resource_behavior": [
    {
      // this entire service was renamed  
      "resource": "volumev2/(.*)",
      "identity_in_v1_api": "cinder/$1"
    },
    {
      // this service contains sets of resources for each share type
      "resource": "manila/(shares|snapshots|share_capacity|snapshot_capacity)_(.+)",
      "category": "share_type_$2"
    }
  ]
}
```
````

File: internal/collector/scrape_test.go [hunk 2]
```diff
@@ -64,31 +64,37 @@ func prepareDomainsAndProjectsForScrape(t *testing.T, s test.Setup) {
...
-			unittest:
-				area: testing
-				# to check how they are merged with the ServiceInfo of the liquids
-				rate_limits:
-					global:
-						- name: xOtherRate
-							limit:  5000
-							window: 1s
-		quota_distribution_configs:
-			# this is only used to check that historical_usage is tracked
```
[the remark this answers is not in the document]
Varsius: I looked into this today but did not reach a satisfactory solution:

Option 1: Move comments above of the config string (Sandro's suggestion)
Since the config files are oftentimes quite long, it is cumbersome to make accurate descriptions of where these comments belong. It would be much more readable and helpful if the comments were placed at the appropriate spots in the config.

Option 2: Use implicit string concatenation
This option works in theory, however the auto-formatter messes up the spacing if the comment should be indented (which is practically always the case).

Option 3: Write a parser that removes comments from json strings
This can be tricky with inline comments (which we could possibly forbid). Also this would be reinventing jsonc/json5 parsing.

I need further advise on this please.

internal/core/translation_rule.go
```diff
@@ -19,9 +19,9 @@ import (
 // that was generated by the old plugins for quota or capacity.
 type TranslationRule struct {
 	// If not nil, reports need to pass all `subcapacities` strings through this handler.
-	TranslateSubcapacities func(string, limes.AvailabilityZone, liquid.ResourceName, liquid.ResourceInfo) (string, error)
+	TranslateSubcapacities func(string, limes.AvailabilityZone, liquid.ResourceName, liquid.ResourceInfo) (string, error) `json:"-"`
 	// If not nil, reports need to pass all `subresources` strings through this handler.
-	TranslateSubresources func(string, limes.AvailabilityZone, liquid.ResourceName, liquid.ResourceInfo) (string, error)
+	TranslateSubresources func(string, limes.AvailabilityZone, liquid.ResourceName, liquid.ResourceInfo) (string, error) `json:"-"`
```
majewsky: These `json:"-"` tags are not necessary because `TranslationRule` implements `json.Unmarshaler` and therefore encoding/json never inspects the struct tags.

PR #787, opened 2025-10-29
introduce dummy AZ "total"

wagnerd3:
Todo:

- [x] extensive manual testing
- [x] check impact to monitoring queries

content:
* `az_resources` and `project_az_resources` get a `total` entry, which holds the sums for all the availability zones
* this makes some queries with regional and az values in the metrics shorter and shortens the insertion processes in some places, as only az-level has to be touched
* the fields `quota` and `backend_quota` are removed from `project_resources` 

how I tested:
* I had the metrics exporter running locally and observed the changes in a local prometheus
* First, I just deployed the migration and ensured there is no changes in the data
* Secondly, I had the collector run in parallel and observed that there is no anomalies in the exported data

follow-ups that I noticed might make sense:
* adjust the interface to the liquid, so that the sums are directly transmitted for all fields from the liquid and are not translated during scrape?

### DIFF
--- a/internal/liquids/archer/liquid.go
+++ b/internal/liquids/archer/liquid.go
@@ -76,8 +76,8 @@ func (l *Logic) ScanUsage(ctx context.Context, projectUUID string, req liquid.Se
 // SetQuota implements the liquidapi.Logic interface.
 func (l *Logic) SetQuota(ctx context.Context, projectUUID string, req liquid.ServiceQuotaRequest, serviceInfo liquid.ServiceInfo) error {
 	return l.Archer.putQuotaSet(ctx, projectUUID, quotaSet{
-		Endpoint: int64(req.Resources["endpoints"].Quota),
-		Service:  int64(req.Resources["services"].Quota),
+		Endpoint: int64(req.Resources["endpoints"].Quota), //nolint:gosec // uint64 -> int64 would only fail if quota is bigger than 2^63
+		Service:  int64(req.Resources["services"].Quota),  //nolint:gosec // uint64 -> int64 would only fail if quota is bigger than 2^63
 	})
 }
 
--- a/internal/liquids/designate/liquid.go
+++ b/internal/liquids/designate/liquid.go
@@ -101,12 +101,12 @@ func (l *Logic) ScanUsage(ctx context.Context, projectUUID string, req liquid.Se
 // SetQuota implements the liquidapi.Logic interface.
 func (l *Logic) SetQuota(ctx context.Context, projectUUID string, req liquid.ServiceQuotaRequest, serviceInfo liquid.ServiceInfo) error {
 	return l.DesignateV2.setQuota(ctx, projectUUID, quotaSet{
-		Zones:             int64(req.Resources["zones"].Quota),
-		RecordsetsPerZone: int64(req.Resources["recordsets_per_zone"].Quota),
+		Zones:             int64(req.Resources["zones"].Quota),               //nolint:gosec // uint64 -> int64 would only fail if quota is bigger than 2^63
+		RecordsetsPerZone: int64(req.Resources["recordsets_per_zone"].Quota), //nolint:gosec // uint64 -> int64 would only fail if quota is bigger than 2^63
 
 		// Designate has a records_per_recordset quota of default 20, so if we set
 		// ZoneRecords to 20 * ZoneRecordsets, this quota will not disturb us
-		RecordsPerZone: int64(req.Resources["recordsets_per_zone"].Quota * 20),
+		RecordsPerZone: int64(req.Resources["recordsets_per_zone"].Quota * 20), //nolint:gosec // uint64 -> int64 would only fail if quota is bigger than 2^63 / 20
 	})
 }
 
--- a/internal/liquids/nova/binpack_simulation.go
+++ b/internal/liquids/nova/binpack_simulation.go
@@ -119,7 +119,7 @@ func prepareHypervisorForBinpacking(h matchingHypervisor, pooledExtraSpecs map[s
 	}
 	result := binpackHypervisor{
 		Match:                h,
-		Nodes:                make([]*binpackNode, int(nodeCount)),
+		Nodes:                make([]*binpackNode, int(nodeCount)), //nolint:gosec // uint64 -> int conversion is okay, if there is more than 2^63 elements, we have other problems
 		AcceptsPooledFlavors: FlavorMatchesHypervisor(flavors.Flavor{ExtraSpecs: pooledExtraSpecs}, h),
 	}
 	for idx := range result.Nodes {
--- a/internal/liquids/swift/liquid.go
+++ b/internal/liquids/swift/liquid.go
@@ -109,7 +109,7 @@ func (l *Logic) ScanUsage(ctx context.Context, projectUUID string, req liquid.Se
 	}
 
 	// get quota and usage data from account headers
-	quota := int64(headers.BytesUsedQuota().Get())
+	quota := int64(headers.BytesUsedQuota().Get()) //nolint:gosec // uint64 -> int64 would only fail if quota is bigger than 2^63
 	if !headers.BytesUsedQuota().Exists() {
 		quota = -1
 	}


internal/datamodel/project_resource_update.go
```diff
@@ -143,15 +139,6 @@ func (u ProjectResourceUpdate) Run(dbi db.Interface, serviceInfo liquid.ServiceI
 			}
 		}
 		result = append(result, res)
-
-		// check if we need to arrange for SetQuotaJob to look at this project service
-		if resInfo.HasQuota && resInfo.Topology != liquid.AZSeparatedTopology {
-			backendQuota := res.BackendQuota.UnwrapOr(-1)
-			quota := res.Quota.UnwrapOr(0) // definitely not None, it was set above in validateResourceConstraints()
-			if backendQuota < 0 || uint64(backendQuota) != quota {
-				hasBackendQuotaDrift = true
```
majewsky: This was the only place where the variable has been set, so all logic related to it can be removed. This is safe because ACPQ has its own query for requesting quota syncs from AZ quota changes.

internal/api/inconsistencies_test.go
```diff
@@ -61,13 +68,15 @@ func TestInconsistencyReport(t *testing.T) {
 	}.Check(t, s.Handler)
 
 	// now put in some inconsistencies
-	s.MustDBExec(`UPDATE project_resources SET backend_quota = 10 WHERE project_id = $1 AND resource_id = $2`,
+	s.MustDBExec(`UPDATE project_az_resources SET backend_quota = 10 WHERE project_id = $1 AND az_resource_id IN (SELECT id FROM az_resources WHERE resource_id = $2 AND az = $3)`,
 		s.GetProjectID("dresden"),
 		s.GetResourceID("shared", "capacity"),
+		liquid.AvailabilityZoneTotal,
```
majewsky:
```suggestion
	s.MustDBExec(`UPDATE project_az_resources SET backend_quota = 10 WHERE project_id = $1 AND az_resource_id = $2`,
		s.GetProjectID("dresden"),
		s.GetAZResourceID("shared", "capacity", liquid.AvailabilityZoneTotal),
```
Same directly below.

internal/api/api_test.go
```diff
@@ -1610,8 +1627,14 @@ func Test_SeparatedTopologyOperations(t *testing.T) {
...
 		)
 	`, "az-one", "az-two")
+	s.MustDBExec(`
+		UPDATE project_az_resources SET backend_quota = 10, quota = 10, usage = 2 WHERE az_resource_id IN (
+			SELECT id FROM az_resources WHERE az = $1
+		)
+	`, liquid.AvailabilityZoneTotal)
 
 	// This test ensures that the consumable limes APIs do not break with the introduction (or further changes) of the az separated topology.
+	// TODO: the change in the following 2 assertion showed to me that the old implementation was incomplete? Or do we want to leave out the total values on purpose?
```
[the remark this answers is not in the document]
wagnerd3: Removed comment, as the old state was restored.

internal/collector/capacity_scrape_test.go
```diff
@@ -520,14 +535,16 @@ func TestScanCapacityReportsZeroValues(t *testing.T) {
...
+		UPDATE az_resources SET raw_capacity = 0, usage = 0 WHERE id = 6 AND resource_id = 2 AND az = 'any' AND path = 'shared/things/any';
+		UPDATE az_resources SET raw_capacity = 0, usage = 0 WHERE id = 7 AND resource_id = 2 AND az = 'total' AND path = 'shared/things/total';
 		UPDATE services SET scraped_at = %d, next_scrape_at = %d WHERE id = 1 AND type = 'shared' AND liquid_version = 1;
 	`,
 		s.Clock.Now().Unix(), s.Clock.Now().Add(15*time.Minute).Unix(),
 	)
 }
 
-func Test_ScanCapacityUnknownAZVanishes(t *testing.T) {
+func Test_ScanCapacityAZVanishes(t *testing.T) {
```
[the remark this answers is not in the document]
wagnerd3: Oh no, that's actually how I recognized:
When capacities from a real AZ vanish, we get the same problem as with unknown. Unknown is just the example I chose to continue using in the test.

internal/collector/capacity_scrape.go
```diff
@@ -171,7 +173,8 @@ func (c *Collector) processCapacityScrapeTask(ctx context.Context, task capacity
 	}
 	var dbAZResources []db.AZResource
 	whereClause, queryArgs := db.BuildSimpleWhereClause(map[string]any{"resource_id": dbResourceIDs}, 0)
-	_, err = tx.Select(&dbAZResources, `SELECT * FROM az_resources WHERE `+whereClause, queryArgs...)
+	_, err = tx.Select(&dbAZResources, `SELECT * FROM az_resources WHERE `+whereClause+
+		` ORDER BY CASE WHEN az != 'total' THEN 0 ELSE 1 END`, queryArgs...)
```
majewsky: For consistency with everything else, this should use db.ExpandEnumPlaceholders(). Note though that, if my previous suggestion is taken, this comment might be moot.

internal/collector/sync_quota_to_backend.go
```diff
@@ -173,50 +157,30 @@ func (c *Collector) performQuotaSync(ctx context.Context, srv db.ProjectService,
...
-					azSeparatedNeedsApply = true
-				}
-				return nil
-			})
-			if err != nil {
-				return err
-			}
+		targetQuota, targetQuotaExists := targetQuotaPtr.Unpack()
+		if !targetQuotaExists {
+			return fmt.Errorf("found unexpected NULL value in project_az_resources.quota for id = %d", projectAZResourceID)
```
majewsky: Since we have the full AZResourceLocation, please include it in the log:
```suggestion
			return fmt.Errorf("found unexpected NULL value in project_az_resources.quota for id = %d (location = %s/%s/%s)", projectAZResourceID, serviceType, resourceName, availabilityZone)
```

internal/collector/sync_quota_to_backend.go
```diff
@@ -173,50 +157,30 @@ func (c *Collector) performQuotaSync(ctx context.Context, srv db.ProjectService,
...
+		}
+		currentQuota, currentQuotaExists := currentQuotaPtr.Unpack()
+		// defense in depth: configured backend_quota for AZ any or unknown are not valid for the azSeparatedQuota topology.
+		if resInfo.Topology == liquid.AZSeparatedTopology && (availabilityZone == liquid.AvailabilityZoneAny || availabilityZone == liquid.AvailabilityZoneUnknown) && !currentQuotaExists {
+			return fmt.Errorf("detected invalid AZ: %s for resource: %s with topology: %s has backend_quota: %v", availabilityZone, resourceName, resInfo.Topology, currentQuota)
```
majewsky: Same here. `resourceName` alone might be ambiguous (e.g. `capacity`), so we should include the service type. Also, I'm not a huge fan of the colon-overusing message style. Also also, `currentQuota` is not well-defined in this branch because we know that `currentQuotaExists == false`.
```suggestion
			return fmt.Errorf("detected invalid AZ %q for resource %s/%s with topology %s (backend quota was nil)", availabilityZone, serviceType, resourceName, resInfo.Topology)
```

majewsky (changes requested):
I have now read through everything.

internal/collector/capacity_scrape.go
```diff
@@ -196,26 +199,32 @@ func (c *Collector) processCapacityScrapeTask(ctx context.Context, task capacity
...
 			if azResourceData.Capacity > 0 {
 				azRes.LastNonzeroRawCapacity = Some(azResourceData.Capacity)
 			}
+			// the total AZ gets the sum of all AZs' LastNonzeroRawCapacity
+			// NOTE: this is different to the sum of the current capacities if one is 0!
+			if azRes.AvailabilityZone != liquid.AvailabilityZoneTotal {
+				totalLastNonzeroRawCapacity += azRes.LastNonzeroRawCapacity.UnwrapOr(0)
+			} else if totalLastNonzeroRawCapacity > 0 {
+				azRes.LastNonzeroRawCapacity = Some(totalLastNonzeroRawCapacity)
+			}
```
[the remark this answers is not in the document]
wagnerd3: See my reply above: I think it's inconsistent if `sum(last_nonzero_raw_capacity)` is not equal to the value on `az=total`. But you are right, the value is unused. Would it be more correct for you, if we leave it `null`, like we leave `subcapacities/subresources=''` as they are unused on `az=total`?

internal/collector/scrape.go
```diff
@@ -581,26 +596,41 @@ func (c *Collector) writeDummyResources(dbDomain db.Domain, dbProject db.Project
...
 		if err != nil {
 			return err
 		}
 	}
 
-	// FIXME: These dummy resources do not conform to `resInfo.Topology` and are never AZ-aware.
-	//        I'm not fixing this right now because dummy resources are an extremely rare corner-case anyway.
-	// TODO:  When we rework the DB schema next year, we should build it so that dummy resources can be avoided entirely.
+	// TODO: Do we still want to find a way to make the datamodel work without dummy resources?
+	// with the total-AZ, we are kind of getting away from this desire, so at some point, we should discuss this again.
```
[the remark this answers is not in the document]
wagnerd3: I would note this for later and exclude it from this PR for now.

internal/collector/capacity_scrape.go
wagnerd3: see above.

PR #807, opened 2025-12-08
optimize writing queries in ACPQ by avoiding joins

majewsky:
This loads `services.id` and `az_resources.id` only once at the beginning to avoid JOINs in the UPDATE queries that get executed several times at the end.

### DIFF
--- a/internal/datamodel/apply_computed_project_quota.go
+++ b/internal/datamodel/apply_computed_project_quota.go
@@ -26,10 +26,11 @@ import (
 // since they represent individual steps of ApplyComputedProjectQuota.
 
 var (
-	acpqGetResourceIDQuery = sqlext.SimplifyWhitespace(`
-		SELECT r.id
+	acpqGetClusterwideIDsQuery = sqlext.SimplifyWhitespace(`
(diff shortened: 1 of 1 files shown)


internal/datamodel/apply_computed_project_quota.go
```diff
@@ -117,10 +114,24 @@ func ApplyComputedProjectQuota(serviceType db.ServiceType, resourceName liquid.R
 	defer sqlext.RollbackUnlessCommitted(tx)
 
 	// to avoid excessive joins, convert serviceType+resourceName into a ResourceID only once
-	var resourceID db.ResourceID
-	err = tx.QueryRow(acpqGetResourceIDQuery, serviceType, resourceName).Scan(&resourceID)
+	var (
+		serviceID     db.ServiceID
+		resourceID    db.ResourceID
+		azResourceIDs = make(map[liquid.AvailabilityZone]db.AZResourceID)
```
wagnerd3: I used a name like `azResourceIDbyAZ` in all other places with maps in my last PRs. Is this a good a idea or is it more idiomatic to name the variable via the values of the map?

PR #814, opened 2025-12-12
add transfer_status to /metrics commitment data

wagnerd3:
At some point, I noted down that we want some metrics for the commitments being in transfer. 
I recognized, that `state` is already in the grouping. IMHO it does not make sense to make a separate metric for the one's in transfer, because that does not give you the combination of both (e.g. `state=planned` but `transfer_status=public`).
Therefore, I combined them now.

Please tell me, whether this is desirable.
I also omitted `transfer_status=""`, because that somehow looks bad as a label to me. We could also invent an explicit `"None"` for that?

### DIFF
--- a/internal/collector/capacity_scrape_test.go
+++ b/internal/collector/capacity_scrape_test.go
@@ -1370,6 +1370,15 @@ func Test_ScanCapacityWithCommitmentTakeover(t *testing.T) {
 	})
 	tr.DBChanges().Ignore()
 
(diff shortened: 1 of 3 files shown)


majewsky (changes requested):
Unfortunately, I have to veto this because it's a backwards-incompatible change to customer-visible metrics. If a customer does e.g. `limes_project_committed_per_az{service="...",resource="...",state="confirmed"} < some_threshold`, they would get false-positive alerts after this change.

I have taken a note to make `transfer_status` a label dimension in the respective v2 metric family.

PR #815, opened 2025-12-12
pass through exact commitment rejection reason

wagnerd3:
At some point during the implementations of the automatic commitment transfer, I made a note from a discussion to pass through the exact `RejectionReason` everywhere. 

Please give bikeshed the name `commitmentChangeRequestWasRejected`. We need to have a path where no http error is written for the `/can-confirm` api, so I did not like something `respondWithCCRWasRejected`...

### DIFF
--- a/internal/api/commitment.go
+++ b/internal/api/commitment.go
@@ -457,11 +457,7 @@ func (p *v1Provider) CanConfirmNewProjectCommitment(w http.ResponseWriter, r *ht
 	if respondwith.ObfuscatedErrorText(w, err) {
 		return
 	}
-	result := true
-	if commitmentChangeResponse.RejectionReason != "" {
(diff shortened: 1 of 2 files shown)


internal/api/commitment.go
```diff
@@ -2357,10 +2345,19 @@ func liquidProjectMetadataFromDBProject(dbProject db.Project, domain db.Domain,
...
 		w.Header().Set("Retry-After", retryAt.Format(time.RFC1123))
 	}
+	if rejected && withHTTPResponse {
+		http.Error(w, response.RejectionReason, http.StatusConflict)
+		return true
+	}
+	if rejected {
+		return true
+	}
+	return false
```
majewsky:
```suggestion
	return rejected
```
Alternatively, the function can be rewritten to return early from `response.RejectionReason == ""` since everything else is only relevant when it isn't.

PR #835, opened 2026-02-19
/v2/info API

wagnerd3:
> That's one small step for man, one giant leap for mankind

This marks the start of development of the actual `/v2` API. We start with the info endpoint, which will include all static configuration in future, so it can be divided from movement data.


majewsky (changes requested):
First bunch of comments.

internal/api/v2_info.go
```diff
@@ -0,0 +1,266 @@
...
+	Version            int64                                `json:"version"`
+	DisplayName        string                               `json:"display_name"`
+	ResourceCategories map[string]V2ServiceCategoryReport   `json:"resources"`
+	Rates              map[liquid.RateName]V2RateInfoReport `json:"rates"`
+}
+
+// V2ServiceCategoryReport groups resources into categories, which are defined in the config.
+// It appears in V2ServiceInfoReport.
+type V2ServiceCategoryReport struct {
+	// DisplayName string `json:"display_name"` // TODO: needed?
```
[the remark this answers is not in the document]
wagnerd3: This is inherent of the latest change that we did in `go-api-declarations`, will rebase and implement that when the display-name PR was merged.

majewsky (changes requested):
I only had 30 minutes today, so this is just a quick sweep. I will try and get into the details tomorrow.

internal/api/api_v2/v2_api_spec.go
```diff
@@ -0,0 +1,154 @@
...
+
+// V2ResourcesResourceInfoReport contains details about a resource.
+// It appears in V2ResourcesCategoryReport.
+type V2ResourcesResourceInfoReport struct {
+	DisplayName      string                                         `json:"display_name"`
+	Unit             Option[liquid.Unit]                            `json:"unit,omitzero"`
+	Topology         liquid.Topology                                `json:"topology"`
+	HasCapacity      bool                                           `json:"has_capacity"`
+	HasQuota         bool                                           `json:"has_quota"`
+	CommitmentConfig Option[limesresources.CommitmentConfiguration] `json:"commitment_config,omitzero"`
```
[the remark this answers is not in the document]
wagnerd3: I refactored it so that `ScopedCommitmentBehavior.ForAPI` uses `ScopedCommitmentBehavior.ForV2API` and can be deleted when the old API is gone.

internal/api/api_v2/v2_info.go
```diff
@@ -0,0 +1,282 @@
...
+	if !token.Require(w, "project:show") {
+		return token, projectUUID, domainUUID, errors.New("Unauthorized")
+	}
+
+	// a cluster admin can see everything, regardless of the scope of his token
+	if token.Check("cluster:show") {
+		projectUUID = ""
+		domainUUID = ""
+	}
+	return token, projectUUID, domainUUID, nil
```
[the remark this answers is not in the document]
wagnerd3: Both points implemented. Do we need to mention this in the docs somehow/ somewhere? I only mentioned the `v2:[scope]:[endpoint]` scheme for the test configuration now.

internal/apideclarations/apiv2/rates/info.go
```diff
@@ -0,0 +1,57 @@
...
+// It contains all metadata information about the clusters services and rates.
+type InfoReport struct {
+	// The Area is a grouping of multiple services, which serve the same purpose.
+	// E.g. compute, storage, network, etc.
+	Areas map[string]AreaReport `json:"service_areas"`
+}
+
+// AreaReport groups services into areas, which are defined in the config.
+// It appears in InfoReport.
+type AreaReport struct {
```
majewsky: For consistency with the other levels:
```suggestion
// AreaInfoReport groups services into areas, which are defined in the config.
// It appears in InfoReport.
type AreaInfoReport struct {
```

Same for resourcesv2.AreaReport and resourcesv2.CategoryReport.

internal/apideclarations/apiv2/rates/info.go
```diff
@@ -0,0 +1,57 @@
...
+	DisplayName string                               `json:"display_name"`
+	Services    map[db.ServiceType]ServiceInfoReport `json:"services"`
+}
+
+// ServiceInfoReport contains details about a service.
+// It appears in AreaReport.
+type ServiceInfoReport struct {
+	Version     int64                              `json:"version"`
+	DisplayName string                             `json:"display_name"`
+	Rates       map[liquid.RateName]RateInfoReport `json:"rates"`
```
majewsky, line 47: Upon comparing both sides of the structure: Do we have an opinion on whether to add categories for rates, too? I don't think we have thought about this yet.

internal/apideclarations/apiv2/rates/info.go
```diff
@@ -0,0 +1,57 @@
...
+// RateLimitReport contains details about the limits of a rate.
+// Default limits might exist on cluster and project level.
+// Additionally, the local limit might be set on project level.
+// This object cannot exist on domain level.
+type RateLimitReport struct {
+	Limit         uint64                    `json:"limit,omitempty"`
+	Window        Option[limesrates.Window] `json:"window,omitzero"`
+	DefaultLimit  uint64                    `json:"default_limit,omitempty"`
+	DefaultWindow Option[limesrates.Window] `json:"default_window,omitzero"`
+}
```
majewsky: Within info reports, we are only showing defaults, so there is no usecase for the `limit` and `window` attributes. Also, within cluster-level and project-level reports, there is no point to repeat the default values that are already shown in the info reports. I therefore suggest to split both pieces apart, and to show `Default{Limit,Window}` directly within RateInfoReport:

```suggestion
	DefaultLimit  uint64                    `json:"default_limit,omitempty"`
	DefaultWindow Option[limesrates.Window] `json:"default_window,omitzero"`
}
```

internal/core/cluster.go
majewsky, line 752: Should this function be updated to use the ServiceInfoCache if available? If so, I'm not saying that we must do it in this PR, but we should leave a TODO comment and take an action item in the backlog.

majewsky (changes requested):
I have read `internal/apideclarations` and `internal/test` and most of `internal/core` so far. For the nitpicks, I pushed a commit; please counter-review. These are my bigger remarks:

internal/apideclarations/apiv2/rates/info.go
wagnerd3: The only use of the categories is really for display purposes, it does not provide any other benefit. We currently don't have rates in the limes UI, so I thought it's just unnecessary work that we can safe for the rates. What's the plan with regards to showing rates in the limes UI?

Regardless of our answer to that, if we want to do it for consistency I am also open for it. But I would note this in a separate follow up ticket together with the last point.

internal/apideclarations/apiv2/rates/info.go
wagnerd3: Mhhh, I saw a rate limit not as moving data. For me, it's configuration which belongs to `/info`. But I am fine with that split and will adjust it accordingly.

internal/core/cluster.go
wagnerd3: I created an action item in the backlog together with the above mentioned point.

internal/apideclarations/apiv2/rates/info.go
majewsky: As an advance concession, I would like to have categories for rates, but we just place all rates in the default category (i.e. CategoryName == ServiceType) throughout. That way, we can introduce categories properly later without having asymmetric API structure.

internal/api/api_v2/v2_info.go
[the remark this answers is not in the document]
majewsky: The example-policy.json is the documentation.

internal/api/api_v2/core.go
```diff
@@ -0,0 +1,115 @@
...
+			},
+		},
+	}
+
+	return p
+}
+
+// AddTo implements the httpapi.API interface.
+func (p *v2Provider) AddTo(r *mux.Router) {
+	r.Methods("HEAD", "GET").Path("/").HandlerFunc(func(w http.ResponseWriter, r *http.Request) {
```
majewsky: This path (`HEAD /` and `GET /`) is provided by both the v1 and v2 provider, which will not work as intended. The router will randomly choose whatever route it looks at first during evaluation.

Please move the VersionData responses (paths `/`, `/v1/` and `/v2/`) and the related method `v2Provider.Path` to a wholly separate API provider.

internal/api/api_v2/info.go
```diff
@@ -0,0 +1,238 @@
...
+	"github.com/sapcc/go-bits/sqlext"
+
+	ratesv2 "github.com/sapcc/limes/internal/apideclarations/apiv2/rates"
+	resourcesv2 "github.com/sapcc/limes/internal/apideclarations/apiv2/resources"
+	"github.com/sapcc/limes/internal/db"
+
+	. "github.com/majewsky/gg/option"
+)
+
+var findAllowedResources = sqlext.SimplifyWhitespace(`
```
majewsky: Please add the suffix `Query` on global vars holding query strings. Otherwise the name sounds like a function:

```suggestion
var findAllowedResourcesQuery = sqlext.SimplifyWhitespace(`
```

internal/api/api_v2/info.go
```diff
@@ -0,0 +1,238 @@
...
+
+	// a token.Require() check below cloud-level is only successful, when scope is in the context
+	// usually, this would come from vars in the URL
+	if projectUUID != "" {
+		token.Context.Request["project_id"] = projectUUID
+		token.Context.Request["domain_id"] = projectDomainUUID
+	}
+	if domainUUID != "" {
+		token.Context.Request["domain_id"] = domainUUID
+	}
```
majewsky: Why is this necessary? Policy rules can already access those attributes through `token.Context.Auth`.

internal/api/api_v2/info.go
```diff
@@ -0,0 +1,238 @@
...
+			if _, exists := serviceReport.Categories[category]; !exists {
+				serviceReport.Categories[category] = resourcesv2.CategoryInfoReport{
+					DisplayName: categoryDisplayName,
+					Resources:   make(map[liquid.ResourceName]resourcesv2.ResourceInfoReport),
+				}
+			}
+			unit := None[liquid.Unit]()
+			if !resource.Unit.IsZero() {
+				unit = Some(resource.Unit)
+			}
```
majewsky: What is the `None` case here? Resources should always have units. From memory, `u = resource.Unit` with `u.IsZero()` should be converted to UnitNone during serialization, so `u.IsZero()` should not be reachable here.

internal/apideclarations/apiv2/resources/info.go
```diff
@@ -0,0 +1,53 @@
...
+type CategoryInfoReport struct {
+	DisplayName string                                     `json:"display_name"`
+	Resources   map[liquid.ResourceName]ResourceInfoReport `json:"resources"`
+}
+
+// ResourceInfoReport contains details about a resource.
+// It appears in CategoryInfoReport.
+type ResourceInfoReport struct {
+	DisplayName      string                          `json:"display_name"`
+	Unit             Option[liquid.Unit]             `json:"unit,omitzero"`
```
majewsky: And then this should be:
```suggestion
	Unit             liquid.Unit                     `json:"unit"`
```
Or am I missing something?

internal/apideclarations/apiv2/rates/info.go
```diff
@@ -0,0 +1,47 @@
...
+	Version     int64                              `json:"version"`
+	DisplayName string                             `json:"display_name"`
+	Rates       map[liquid.RateName]RateInfoReport `json:"rates"`
+}
+
+// RateInfoReport contains details about a rate.
+// It appears in ServiceInfoReport.
+type RateInfoReport struct {
+	DisplayName   string                    `json:"display_name"`
+	Unit          Option[liquid.Unit]       `json:"unit,omitzero"`
```
majewsky: Same here.

internal/api/api_v2/info.go
```diff
@@ -0,0 +1,238 @@
...
+			scopedCommitmentBehavior := p.Cluster.CommitmentBehaviorForResource(serviceType, resourceName).ForCluster()
+			if domainUUID != "" && projectUUID == "" {
+				scopedCommitmentBehavior = p.Cluster.CommitmentBehaviorForResource(serviceType, resourceName).ForDomain(token.DomainScopeName())
+			}
+			if domainUUID != "" && projectUUID != "" {
+				scopedCommitmentBehavior = p.Cluster.CommitmentBehaviorForResource(serviceType, resourceName).ForDomain(token.ProjectScopeDomainName())
+			}
```
majewsky: We can avoid calling CommitmentBehaviorForResource multiple times:

```suggestion
			var (
				commitmentBehavior := p.Cluster.CommitmentBehaviorForResource(serviceType, resourceName)
				scopedCommitmentBehavior core.ScopedCommitmentBehavior
			)
			switch {
				case domainUUID == "":
					scopedCommitmentBehavior = commitmentBehavior.ForCluster()
				case projectUUID == "":
					scopedCommitmentBehavior = commitmentBehavior.ForDomain(token.DomainScopeName())
				default:
					scopedCommitmentBehavior = commitmentBehavior.ForDomain(token.ProjectScopeDomainName())
			}
```

internal/api/api_v2/info.go
```diff
@@ -0,0 +1,238 @@
...
+				DisplayName:      resource.DisplayName,
+				Unit:             unit,
+				Topology:         resource.Topology,
+				HasCapacity:      resource.HasCapacity,
+				HasQuota:         resource.HasQuota,
+				CommitmentConfig: scopedCommitmentBehavior.ForV2API(p.timeNow()),
+			}
+		}
+	}
+	respondwith.JSON(w, 200, report)
```
majewsky: Another case of "I was too lazy to go back and fix this in v1, but we should do it from the start in v2":
```suggestion
	respondwith.JSON(w, http.StatusOK, report)
```
Same in all other `respondwith.Something()` calls in the implementation and the `Expect{Status,Text,JSON}` Calls in the tests.

internal/api/api_v2/info.go
```diff
@@ -0,0 +1,238 @@
...
+				rir.Unit = Some(rate.Unit)
+			}
+			if rateConfig, ok := config.RateLimits.GetGlobalDefaultRateLimit(rateName); domainUUID == "" && projectUUID == "" && ok {
+				rir.DefaultLimit = rateConfig.Limit
+				rir.DefaultWindow = Some(rateConfig.Window)
+			}
+			if rateConfig, ok := config.RateLimits.GetProjectDefaultRateLimit(rateName); domainUUID != "" && projectUUID != "" && ok {
+				rir.DefaultLimit = rateConfig.Limit
+				rir.DefaultWindow = Some(rateConfig.Window)
+			}
```
majewsky: Ooooooh, now I understand why there were initially two different rate limits in the RateInfoReport... because there *are* two different rate limits on the cluster level.

I don't like that these show different values based on the level (and also there is no logic here for what to do in the domain-scoped case). Please always show the project-default rate limit here, since it is the one that we can safely show to everyone. The global rate limit will then go into the ClusterRateReport (in the same way that resource capacity will be in the ClusterResourceReport).

majewsky (changes requested):
Comments on `internal/api/api_v2`:

internal/api/api_v2/info.go
wagnerd3: I came from a v1 API, where we used request params to switch something in the API. Later I started to verify the scope from the token auth and then did not properly convert this. 😬

internal/api/api_v2/info.go
wagnerd3: Forgot about the `unitNone` handling, you are right.

internal/api/api_v2/info.go
```diff
@@ -0,0 +1,224 @@
...
+			var scopedCommitmentBehavior core.ScopedCommitmentBehavior
+
+			switch {
+			case domainUUID == "":
+				scopedCommitmentBehavior = commitmentBehavior.ForCluster()
+			case projectUUID == "":
+				scopedCommitmentBehavior = commitmentBehavior.ForDomain(domainName)
+			default:
+				scopedCommitmentBehavior = commitmentBehavior.ForDomain(domainName)
+			}
```
majewsky: Two of three branches are now identical, so this can be simplified to `if domainUUID == "" {...} else {...}`.

internal/apideclarations/apiv2/resources/info.go
```diff
@@ -0,0 +1,53 @@
...
+// InfoReport is the response type for GET /resources/v2/info.
+// It contains all metadata information about the clusters services and resources.
+type InfoReport struct {
+	// The Area is a grouping of multiple services, which serve the same purpose.
+	// E.g. compute, storage, network, etc.
+	Areas map[string]AreaInfoReport `json:"service_areas"`
+}
+
+// AreaInfoReport groups services into areas, which are defined in the config.
+// It appears in InfoReport.
```
majewsky: For easier navigation, type references within docstrings should be links. I recently did this for the LIQUID API in https://github.com/sapcc/go-api-declarations/pull/87, but this branch diverged before then.

```suggestion
// It appears in type [InfoReport].
```

Same on all other instances of `appears in` here and in `../rates/*.go`.

internal/db/refs.go
majewsky, line 1: General design remarks regarding the new types being added here (`ResourcePath`, `AZResourcePath` and `RatePath`):

1. Since we now have these types, do we want to change the type of `db.Resource.Path` et al to use those?
2. If yes, then I would suggest a different implementation approach. The current method implementation using `SplitN -> array index` can potentially panic if the path value is malformed and does not have the right amount of fields. For example, `strings.SplitN("foo", "/", 2)[1]` panics because SplitN only returns a slice with one element for this input. This can be avoided by holding the value in a structured form in code, and marshaling into the stringified form during SQL only, using the Valuer and Scanner interfaces. These allow raising errors if the DB returns malformed data (which it shouldn't because of the validation trigger, but still). For example:
   ```go
   type ResourcePath struct {
   	ServiceType ServiceType
   	ResourceName liquid.ResourceName
   }

   // Scan implements the [sql.Scanner] interface.
   func (r *ResourcePath) Scan(src any) error {
   	srcStr, ok := src.(string)
   	if !ok {
   		return fmt.Errorf("invalid input type for %T.Scan(): %T", r, src)
   	}
   	first, second, ok := strings.Cut(srcStr, "/")
   	if !ok || strings.Contains(second, "/") {
   		return fmt.Errorf("invalid value for %T: %q", r, srcStr)
   	}
   	r.ServiceType = db.ServiceType(first)
   	r.ResourceName = liquid.ResourceName(second)
   	return nil
   }

   // Value implements the [driver.Valuer] interface.
   func (r ResourcePath) Value() (driver.Value, error) {
   	return driver.Value(string(r.ServiceType) + "/" + string(r.ResourceName)), nil
   }
   ```

majewsky (changes requested):
Only ServiceInfoCache left to look at now:

internal/core/service_info_cache.go
```diff
@@ -0,0 +1,371 @@
...
+// during runtime.
+type ServiceInfoCache struct {
+	// state
+	DB                               *gorp.DbMap
+	listener                         *pq.Listener
+	servicesByTypeMutex              sync.RWMutex
+	resourcesByNameByTypeMutex       sync.RWMutex
+	azResourcesByAZByNameByTypeMutex sync.RWMutex
+	ratesByNameByTypeMutex           sync.RWMutex
+	categoriesByIDMutex              sync.RWMutex
```
majewsky: Since these mutexes are always write-locked together, I think we can just use one RWMutex for everything without losing anything. On read, the mutexes are not going to be contended a lot, so there is no big risk of performance loss except when the mutexes are locked for write, but then they are all locked anyway.

internal/core/service_info_cache.go
```diff
@@ -0,0 +1,371 @@
...
+func (s *ServiceInfoCache) signalInvalidation() {
+	if s.sendInvalidate != nil {
+		select {
+		case s.sendInvalidate <- struct{}{}:
+		default:
+		}
+	}
+}
+
+func (s *ServiceInfoCache) invalidateServices(serviceTypes ...db.ServiceType) error {
```
majewsky: Since this is only ever called with no arguments (to mean "all services") or one argument (to mean "that service"), the signature would make more sense as:

```suggestion
func (s *ServiceInfoCache) invalidateServices(serviceType Option[db.ServiceType]) error {
```

internal/core/service_info_cache.go
```diff
@@ -0,0 +1,371 @@
...
+	if err != nil {
+		return err
+	}
+	return nil
+}
+
+// fillData requires a write lock on all the variables holding data, if not called on init!
+// Also, it requires all maps of the variables holding data to be initialized on the first level.
+// It loads all service, resource, az_resource, rate, and category data for the requested serviceTypes.
+func (s *ServiceInfoCache) fillData(serviceTypes ...db.ServiceType) error {
```
majewsky: That's a lot of asterisks for `fillData` there. I suggest removing the lines marked by this comment and inlining `fillData` into `invalidateServices`. `New` can call `invalidateServices` for its initial fill; the mutex locking will not be expensive because the mutex is not contended at that point.

internal/core/service_info_cache.go
```diff
@@ -0,0 +1,371 @@
...
+	servicesByType, err := db.BuildIndexOfDBResult(s.DB, func(s db.Service) db.ServiceType { return s.Type }, "SELECT * FROM services WHERE type = ANY($1) OR $1 IS NULL", pq.Array(serviceTypes))
+	if err != nil {
+		return err
```
majewsky: Since this does multiple queries, it would be nice to report which query failed on error:
```suggestion
		return fmt.Errorf("while reading services for types %v: %w", serviceTypes, err)
```

internal/core/service_info_cache.go
```diff
@@ -0,0 +1,371 @@
...
+	categoriesByID, err := db.BuildIndexOfDBResult(
+		s.DB,
+		func(c db.Category) db.CategoryID { return c.ID },
+		"SELECT * FROM categories",
+	)
+	if err != nil {
+		return err
+	}
+	maps.Copy(s.categoriesByID, categoriesByID)
+	return nil
```
majewsky: If `invalidateServices` and `fillData` are merged as I suggested above, the extra map initialization above and the extra map copy here can be avoided:
```suggestion
	s.categoriesByID, err = db.BuildIndexOfDBResult(
		s.DB,
		func(c db.Category) db.CategoryID { return c.ID },
		"SELECT * FROM categories",
	)
	return err
```

majewsky (changes requested):
Comments on ServiceInfoCache overall implementation:

internal/db/refs.go
wagnerd3: I will switch the definition of the paths to be struct as you suggested, but will do the usage of it in the `db` artifacts in a separate PR later. So I will call the parsing function manually in `ServiceInfoCache` for now.

internal/apideclarations/apiv2/rates/info.go
majewsky: This was marked as resolved, but I don't see it being resolved in the code. I adjusted the code now.

majewsky (approved):
To fast-track this review, I have implemented the last few things myself and pushed a few commits. Please counter-review and ask @VoigtS for approval.

internal/apideclarations/apiv2/rates/info.go
wagnerd3: Ah, I noted a follow up for me to introduce this in a follow up. I thought it would make sense to have the api-declaration for it already too, so that it's ready to go if we were to need it on any rate. Sry, I did not note this down here.

wagnerd3:
@majewsky's changes look fine for me. We can merge as is and then I will take care of the rest of the small follow ups I noted down.

internal/db/migrations.go
```diff
@@ -35,4 +35,63 @@ var sqlMigrations = map[string]string{
...
+			RETURNS trigger AS $$
+			DECLARE
+				service_type TEXT;
+			BEGIN
+				IF TG_TABLE_NAME = 'services' THEN
+					FOR service_type IN
+						SELECT s.type FROM services s
+							WHERE s.type = ANY(ARRAY[NEW.type, OLD.type])
+					LOOP
+						PERFORM pg_notify('limes_service_update', service_type);
```
majewsky: I just had a brainwave. This identifier is the first time that the word "Limes" appears in the DB schema, and it will be annoying to change later:
```suggestion
						PERFORM pg_notify('limitas_service_update', service_type);
```
Same for all other uses, obviously.

majewsky (changes requested):

internal/db/migrations.go
wagnerd3: you are right 🤯

wagnerd3:
By now, we collected 11 commits, maybe we should merge-squash this time 😅

majewsky (approved):

PR #870, opened 2026-04-10
data metrics v2: add project-level resource metrics

majewsky:
Finally, we're getting to the cool part. What used to be a horrible mess of literally dozens of joins per query (see `projectMetricsQuery` and `projectAZMetricsQuery` in the v1 implementation) turns into three very simple OLAP queries that each do at most one join.

While sifting through the numbers in the fixture, I noticed that we do not test the case where `allocation = max(usage, committed)` is dominated by confirmed commitments rather than usage. To cover this case, I increased one of the test commitments.

Also, to test the `committable=true|false` label on some of the new metrics, I added a CommitmentBehavior for the resource that has test commitments.

Both of those changes to the test setup also lead to a corresponding diff in the v1 data metrics fixtures as part of this changeset.


internal/collector/metrics.go
```diff
@@ -895,6 +903,20 @@ var (
...
 	`)
+	dmv2ProjectInfoQuery = sqlext.SimplifyWhitespace(`
+		SELECT p.id AS project_id, p.name AS project_name, p.uuid AS project_uuid, d.name AS domain_name, d.uuid AS domain_uuid
+		  FROM domains d
+		  JOIN projects p ON p.domain_id = d.id
+	`)
+	dmv2ProjectCommitmentDataQuery = sqlext.SimplifyWhitespace(db.ExpandEnumPlaceholders(`
+		SELECT project_id, az_resource_id, uuid, amount, expires_at, status, transfer_status
+		  FROM project_commitments
+		 WHERE status NOT IN ({{liquid.CommitmentStatusSuperseded}}, {{liquid.CommitmentStatusExpired}}, {{util.CommitmentStatusDeleted}})
```
wagnerd3, line 915: I just noticed, that we have this "trinity" so often, that we could define a constant just for this `NOT IN ({{liquid.CommitmentStatusSuperseded}}, {{liquid.CommitmentStatusExpired}}, {{util.CommitmentStatusDeleted}})` 😄

internal/collector/metrics.go
```diff
@@ -977,50 +999,188 @@ func (d *DataMetricsV2Reporter) collectMetrics() (*dataMetricSet, error) {
...
+	// 2. to pre-compute resource committability per domain for later
+	type projectInfoRecord struct {
+		ProjectID   db.ProjectID `db:"project_id"`
+		ProjectName string       `db:"project_name"`
+		ProjectUUID string       `db:"project_uuid"`
+		DomainName  string       `db:"domain_name"`
+		DomainUUID  string       `db:"domain_uuid"`
+	}
+	var projectInfoRecords []projectInfoRecord
+	projectInfoRecordsByID := make(map[db.ProjectID]projectInfoRecord, len(projectInfoRecords))
```
wagnerd3: I think you want to move this definition after the `DB.Select` to utilize the `len(projectInfoRecord)`?

docs/operators/metrics.md
```diff
@@ -87,6 +87,12 @@ The data metrics service v2 will report the following metrics (all of type `Gaug
...
+| `limitas_project_resource_commitment_amount` | `az`, `domain`, `domain_id`, `project`, `project_id`, `resource`, `service`, `status`, `transfer_status`, `uuid` | For each present or future commitment, the committed amount of resource. The value is a multiple of the resource's unit. |
+| `limitas_project_resource_commitment_expires_at` | `az`, `domain`, `domain_id`, `project`, `project_id`, `resource`, `service`, `status`, `transfer_status`, `uuid` | For each present or future commitment, the UNIX timestamp of its expiry. This value will always be in the future, because expired commitments do not appear in the API. |
+| `limitas_project_resource_physical_usage` | `az`, `committable`, `domain`, `domain_id`, `project`, `project_id`, `resource`, `service` | For each project, resource and AZ, the amount of resource physically used by the project. Only reported for resources that have report physical usage separately from logical usage. The value is a multiple of the resource's unit. Only values > 0 are reported. |
```
wagnerd3: This issue is copy pasted to the help text as well.

```suggestion
| `limitas_project_resource_physical_usage` | `az`, `committable`, `domain`, `domain_id`, `project`, `project_id`, `resource`, `service` | For each project, resource and AZ, the amount of resource physically used by the project. Only reported for resources that report physical usage separately from logical usage. The value is a multiple of the resource's unit. Only values > 0 are reported. |
```

internal/collector/metrics.go
```diff
@@ -977,50 +999,188 @@ func (d *DataMetricsV2Reporter) collectMetrics() (*dataMetricSet, error) {
...
+
+			// emit limitas_project_resource_usage
+			if prr.Usage != 0 {
+				metric := dataMetric{Labels: labels, Value: float64(prr.Usage)}
+				result["limitas_project_resource_usage"] = append(result["limitas_project_resource_usage"], metric)
+			}
+		}
+
+		// emit limitas_project_resource_quota
+		if crr.AvailabilityZone != liquid.AvailabilityZoneTotal {
```
wagnerd3, line 1177: Is it correct that we don't guard this with `AZHasQuotaForTopology`?

internal/collector/metrics.go
```diff
@@ -977,50 +999,188 @@ func (d *DataMetricsV2Reporter) collectMetrics() (*dataMetricSet, error) {
...
+	}
+	isResourceCommittableForDomainName := make(map[string]map[db.AZResourceID]bool)
+	for _, pir := range projectInfoRecords {
+		projectInfoRecordsByID[pir.ProjectID] = pir
+
+		if _, ok := isResourceCommittableForDomainName[pir.DomainName]; !ok {
+			result := make(map[db.AZResourceID]bool, len(clusterResourceRecords))
+			for _, crr := range clusterResourceRecords {
+				cb := d.Cluster.CommitmentBehaviorForResource(crr.ServiceType, crr.ResourceName).ForDomain(pir.DomainName)
+				result[crr.AZResourceID] = len(cb.Durations) > 0 && cb.MinConfirmDate.IsNoneOr(is.Before(time.Now()))
```
wagnerd3: I guess for consistency it would be better to use something like `nowFunc` which we can propagate `s.Clock.Now()` into, so that one can test this behavior?

internal/collector/metrics.go
```diff
@@ -977,50 +999,188 @@ func (d *DataMetricsV2Reporter) collectMetrics() (*dataMetricSet, error) {
...
+				metric := dataMetric{Labels: labels, Value: float64(prr.Usage)}
+				result["limitas_project_resource_usage"] = append(result["limitas_project_resource_usage"], metric)
+			}
+		}
+
+		// emit limitas_project_resource_quota
+		if crr.AvailabilityZone != liquid.AvailabilityZoneTotal {
+			if quota, ok := prr.Quota.Unpack(); ok && quota != 0 {
+				metric := dataMetric{Labels: labels, Value: float64(quota)}
+				result["limitas_project_resource_quota"] = append(result["limitas_project_resource_quota"], metric)
```
wagnerd3, line 1180: I notice that we don't have any tests for these values, is this intentional?

wagnerd3 (changes requested):

internal/collector/metrics.go
majewsky: I thought about that and did not add the check. `project_az_resources.quota` is filled by ACPQ only for the AZs where it makes sense. If ACPQ somehow assigns quota in the wrong AZs, I would actually like for this to also show up in the metrics, so we have one more avenue to find incorrect values like that.

internal/collector/metrics.go
majewsky: We could define a function on SQL level that wraps this expression; if you want to make a PR for that, I can approve it.

internal/collector/metrics.go
majewsky: Indeed. We only have entries for `az=total`, which was sort of fine for the v1 metrics, but does not work with the v2 metrics. I extended the testcase to run ACPQ before reporting data metrics, so we have some plausible quota values.

wagnerd3 (approved):

PR #875, opened 2026-04-22
merge LiquidConnection.{Scrape, ScrapeRates}(), plus related refactors

majewsky:
I went down the rabbit hole of refactoring one thing and finding yet more things to refactor along the way.

### DIFF
--- a/internal/api/liquid.go
+++ b/internal/api/liquid.go
@@ -51,9 +51,10 @@ func (p *v1Provider) GetServiceCapacityRequest(w http.ResponseWriter, r *http.Re
 }
 
 var getServiceUsageRequestAttributesQuery = sqlext.SimplifyWhitespace(`
-	SELECT usage_report_needs_project_metadata
-	  FROM services
-	 WHERE type = $1
+	SELECT s.usage_report_needs_project_metadata, ps.serialized_scrape_state
+	  FROM project_services ps
+	  JOIN services s ON ps.service_id = s.id
+	  WHERE ps.project_id = $1 AND s.type = $2
 `)
 
 // GetServiceUsageRequest handles GET /admin/liquid/service-usage-request?service_type=:type&project_id=:id.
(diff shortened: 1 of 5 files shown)


internal/api/liquid.go
```diff
@@ -31,27 +31,32 @@ func (p *v1Provider) GetServiceCapacityRequest(w http.ResponseWriter, r *http.Re
...
+	serviceCapacityRequest, err := core.BuildServiceCapacityRequest(backchannel, p.Cluster.Config.AvailabilityZones, serviceType, serviceInfo.Resources)
 	if respondwith.ObfuscatedErrorText(w, err) {
 		return
 	}
 
 	respondwith.JSON(w, http.StatusOK, serviceCapacityRequest)
 }
 
+var getServiceUsageRequestAttributesQuery = sqlext.SimplifyWhitespace(`
+	SELECT s.usage_report_needs_project_metadata, ps.serialized_scrape_state
```
wagnerd3, line 54: In future `usageReportNeedsProjectMetadata` should come from the `serviceInfo` and not from this query. I want to file a PR to use the `ServiceInfoCache` instead of `ServiceInfo` next, where we have access to the `db.Service`. So we could skip the join in this query anyways by supplying the `s.type` from the cache :-)

But I am fine to leave it like this for this PR because it's consistent E2E and take note to adjust this within my PR. Just wanted to mention this here.

internal/api/liquid.go
```diff
@@ -31,27 +31,32 @@ func (p *v1Provider) GetServiceCapacityRequest(w http.ResponseWriter, r *http.Re
...
 	}
 
-	serviceInfos, err := p.Cluster.AllServiceInfos()
+	maybeServiceInfo, err := p.Cluster.InfoForService(serviceType)
 	if respondwith.ObfuscatedErrorText(w, err) {
 		return
 	}
-
-	// TODO prevent requests with `liquid-$SERVICE_TYPE` when limesctl was adjusted
-	normalizedServiceType := db.ServiceType(strings.TrimPrefix(string(serviceType), "liquid-"))
```
wagnerd3, line 40: I don't think this can be removed yet, we forgot to change `limesctl` for that?
https://github.com/sapcc/limesctl/blob/master/internal/cmd/liquid.go#L387

wagnerd3 (changes requested):
For the point about `liquid-$serviceType` we would at least need a separate limesctl PR which should be merged before this PR, if I am not mistaken?

internal/collector/scrape.go
wagnerd3, line 172: I just realized: We should return here, right?

internal/api/liquid.go
majewsky: This is not the relevant part of limesctl. For looking up the liquid in the Keystone catalog, we really do need to look for `liquid-$SERVICE_TYPE` (e.g. `liquid-keppel` because `keppel` would find the Keppel API, not the LIQUID API).

This is the relevant line, where only `serviceType` is used, not `"liquid-"+serviceType`: https://github.com/sapcc/limesctl/blob/b5ee65b3f390251b6054004a2c513411bac38b44/internal/cmd/liquid.go#L360

wagnerd3 (dismissed):
Just the PR-unrelated missing return.

wagnerd3 (approved):

PR #876, opened 2026-04-22
do not crash on impossible delete while starting collector

wagnerd3:
Basically, this prevents that we crash on startup when we cannot save the `ServiceInfo` changes to the DB.
This will not go unnoticed, as any subsequent scrape/ capacity scrape will fail, too.

### DIFF
--- a/internal/core/cluster.go
+++ b/internal/core/cluster.go
@@ -134,6 +134,13 @@ func (c *Cluster) Connect(ctx context.Context, provider *gophercloud.ProviderCli
 			continue
 		}
 		err = conn.Init(ctx, client)
+		if errors.Is(err, ErrLeftoverCommitment) {
+			// we just log this error here and ignore it, so that the startup does not fail
+			// this will produce errors on every scrape subsequently (as if the collector was
+			// already running) and those will trigger alerts subsequently.
+			logg.Error(`failed to initialize service %s: %v`, serviceType, err)
+			continue
+		}
 		if err != nil {
 			errs.Addf("failed to initialize service %s: %w", serviceType, gophercloudext.UnpackError(err))
 		}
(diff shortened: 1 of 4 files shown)


internal/core/cluster.go
```diff
@@ -340,9 +348,33 @@ func RatesForService(serviceInfos map[db.ServiceType]liquid.ServiceInfo, service
 	return serviceInfo.Rates
 }
 
+// ErrLeftoverCommitment is a custom error to define when a leftover commitment
+// prevents deletion of a service, resource or db_resource.
```
majewsky: I guess you mean this?
```suggestion
// prevents deletion of a service, resource or az_resource.
```

internal/core/cluster.go
```diff
@@ -340,9 +348,33 @@ func RatesForService(serviceInfos map[db.ServiceType]liquid.ServiceInfo, service
...
+var ErrLeftoverCommitment = errors.New("ErrLeftoverCommitment")
+
 ////////////////////////////////////////////////////////////////////////////////
 // Utility functions for working with ServiceInfo and DB
 
+func generateDeleteFunc[T any](dbm *gorp.DbMap, serviceType string, resourceNameFunc, azFunc func(o T) string) func(T) error {
+	return func(o T) error {
+		count, err := dbm.SelectInt(db.ExpandEnumPlaceholders(`
```
majewsky: Please use sqlext.SimplifyWhitespace (newlines in queries throw off the PostgreSQL logging format) and place the query in a global var to avoid running the simplify and expand steps repeatedly.

internal/core/cluster.go
```diff
@@ -340,9 +348,33 @@ func RatesForService(serviceInfos map[db.ServiceType]liquid.ServiceInfo, service
...
 }
 
+// ErrLeftoverCommitment is a custom error to define when a leftover commitment
+// prevents deletion of a service, resource or db_resource.
+var ErrLeftoverCommitment = errors.New("ErrLeftoverCommitment")
+
 ////////////////////////////////////////////////////////////////////////////////
 // Utility functions for working with ServiceInfo and DB
 
+func generateDeleteFunc[T any](dbm *gorp.DbMap, serviceType string, resourceNameFunc, azFunc func(o T) string) func(T) error {
```
majewsky: Suggestion to simplify the function signature:
```suggestion
func generateDeleteFunc[T any](dbm *gorp.DbMap, getAZResourcePathPattern func(o T) string) func(T) error {
```

And then call like
```go
generateDeleteFunc(dbm, func(s db.Service) string { return string(s.Type) + "/%" })
generateDeleteFunc(dbm, func(r db.Resource) string { return r.Path.String() + "/%" })
generateDeleteFunc(dbm, func(azr db.AZResource) string { return azr.Path.String() })

File: internal/db/setupdate.go [hunk 4]
```diff
@@ -82,6 +85,12 @@ func (u SetUpdate[R, K]) Execute(tx *gorp.Transaction) ([]R, error) {
 	for _, r := range u.ExistingRecords {
 		k := u.KeyForRecord(r)
 		if !slices.Contains(u.WantedKeys, k) {
+			if deleteFunc, ok := u.Delete.Unpack(); ok {
+				err := deleteFunc(r)
+				if err != nil {
+					return nil, fmt.Errorf("could not execute delete callback %T record with key %v: %w", r, k, err)
+				}
+			}
```
majewsky: The fieldname `SetUpdate.Delete` is misleading since it does not actually delete.
```suggestion
			if preDeleteFunc, ok := u.PreDelete.Unpack(); ok {
				err := preDeleteFunc(r)
				if err != nil {
					return nil, fmt.Errorf("could not execute pre-delete callback %T record with key %v: %w", r, k, err)
				}
			}
```

majewsky (changes requested):

internal/api/commitment_test.go
```diff
@@ -3099,7 +3099,7 @@ func Test_PublicCommitmentCloudAdminActions(t *testing.T) {
 		ExpectStatus: http.StatusBadRequest,
 	}.Check(t, s.Handler)
 
-	// Delete a non existing commitment
+	// PreDelete a non existing commitment
```
majewsky: Overeager search-and-replace?

majewsky (changes requested):

majewsky (approved):

PR #877, opened 2026-04-24
[manila] use netapp harvest recording rules for physical usage metrics

chuan137:
Switch sharePhysicalUsageQuery and snapshotPhysicalUsageQuery to use
the pre-aggregated _for_limes recording rules. Move the max/last_over_time/filter
logic into the recording rules where it belongs.

### DIFF
--- a/internal/liquids/manila/liquid.go
+++ b/internal/liquids/manila/liquid.go
@@ -248,8 +248,8 @@ const (
 	snapshotCapacityQuery = `sum   by (availability_zone_name, project_id, share_type_id) (max by (availability_zone_name, id, project_id, share_id, share_type_id) (last_over_time(openstack_manila_snapshot_size_gauge[15m])))`
 
 	// queries for netapp-exporter metrics
-	sharePhysicalUsageQuery      = `sum by (availability_zone, project_id, share_type) (max by (availability_zone, project_id, share_id, share_type) (last_over_time(netapp_volume_used_bytes         {project_id!="",share_type!="",volume_type!="dp",volume_state="online"}[15m])))`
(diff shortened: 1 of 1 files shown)


chuan137:
aggregation rules that generate the `for_limes` metrics, already deployed
https://github.com/sapcc/helm-charts/blob/72b17d852b02f6d216d7c9831db9eb66230d3a71/prometheus-exporters/netapp-exporter/charts/netapp-harvest-exporter/aggregations/limes.yaml#L61-L69

Carthaca (approved):

wagnerd3 (changes requested):
I looked at some examples for eu-de-1 and see the values are quite different for many entries, not just a few percentage points but factor X. AFAIK `snapshotPhysicalUsageQuery` is what ends up being billed for snapshots, so I have to ask whether the impact of this is aligned with financial management?

chuan137:
double checked the two version of exporters how snapshot usage are scraped, they are from the same source on the NetApp: volumeSpaceAttributes -> SizeUsedBySnapshots. The X factor you have seen is probably due to the metrics are scraped by different exporters at different timestamp. 

I picked a random project, and compare the differences for their shares, the values are consistent with fluctuations around zero.
`netapp_volume_snapshot_used_bytes{project_id="xxxx-xxxx", share_id!="", volume_type!="dp"} - on (share_id) group_left() netapp_volume_snapshots_size_used{share_id!="", volume_type!="dp"}`

wagnerd3 (approved):
You are right, thanks for the hint. I just looked at the current point in time, instead of the history. I see a delay of about 2 minutes maximum but no persistent differences, so I guess the change should be safe.

PR #878, opened 2026-04-27
RenderMailTemplate: change endpoint permission

VoigtS:
the smoke-test user that is used in the pipeline only has `show_basic` permissions, which conflicts with this endpoint. Instead of granting the user more permissions, the endpoint permission gets changed instead.

### DIFF
--- a/docs/example-policy.json
+++ b/docs/example-policy.json
@@ -25,5 +25,6 @@
     "cluster:edit": "rule:cluster_admin",
     "v2:cluster:info": "rule:cluster_admin",
     "v2:domain:info": "rule:domain_viewer",
-    "v2:project:info": "rule:project_viewer"
+    "v2:project:info": "rule:project_viewer",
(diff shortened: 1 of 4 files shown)


wagnerd3 (changes requested):
I just realized that `cluster:show_basic` is used in the old world to give simple project members access to cluster level access and not just to `cloud_` roles.

A nicer way would be to pick up the permission scheme for the `v2` API and introduce a policy for `v2:cluster:admin` and define this to match also the `limes-validation` user we have for this smoke testing. Let's discuss offline.

VoigtS:
the policy now uses the v2:cluster:validation from the proposal: https://github.com/sapcc/helm-charts/pull/11513

wagnerd3 (dismissed):
We need to wait for the chart PR before merging.

majewsky (approved):

majewsky (changes requested):
please adjust `docs/example-policy.json` to document this policy

majewsky (approved):
ok

PR #879, opened 2026-05-04
utilize service info cache everywhere

wagnerd3:
@majewsky asked in #835 what the plan is with the `ServiceInfoCache`, this PR should have the answer:
* Instead of `readServiceInfoFromDB`, we will now use the `ServiceInfoCache/ sic` everywhere
* The concept for invalidating the cache is:
  * non-collector tasks reload via postgres `NOTIFY` feature, so the `sic` might be outdated for a few seconds before the reload comes. This is acceptable because most of the time, the delay between a new `ServiceInfo.Version` and the results of the non-collect tasks is not critical.
  * The collector task reloads the `sic` explicitly, because it is the only source of changes for `ServiceInfo`. This way, we can rely on the `sic` begin 100% up to date in collect tasks, all the time.
* There is one ugly point, which I might want to postpone to a future PR: In some APIs, we need the results of `sic.GetServices()`, `sic.GetResources()` and others. Often times, we only need one part though. So when we load multiple, this is currently not transactional. This could mean, that the `sic` gets reloaded between `services` and `resources` and they are not in sync. Because we just generate reports, metrics etc. from this, I think this is acceptable for now, because we either do early return with defense in depth or we just default some properties (e.g. `HasQuota`) to a save value. Some function signatures really benefit from the split of `services, resources, az_resources, rates`, therefore I did not want to build a combined resultset. Maybe, it just takes a function `getServicesAndResources` (although we might need to many permutations of all options), or we make a function with a bitfield parameter like [here](https://github.com/sapcc/go-api-declarations/blob/main/internal/units/amount.go#L31) and a struct return type?
* In the non-collect tasks, in some areas we could also "yolo it" a bit more and remove `service, resourc etc.` function parameters here and there and call the `sic` multiple times by doing more defense-in-depth in those sub-functions. For simplicity of review I chose not to do this in this PR.


internal/api/commitment.go
```diff
@@ -281,10 +279,7 @@ func (p *v1Provider) GetPublicCommitments(w http.ResponseWriter, r *http.Request
 		if err != nil {
 			return err
 		}
-		// this check is defense in depth (the DB should be consistent with our config)
-		if core.HasResource(serviceInfos, loc.ServiceType, loc.ResourceName) {
-			azResourceLocationsByID[id] = loc
-		}
+		azResourceLocationsByID[id] = loc
```
wagnerd3, line 282: We don't use `ServiceInfo` from the liquid here anymore, so nothing can be inconsistent (both come from the database).

internal/core/service_info_cache_test.go
```diff
@@ -29,6 +30,38 @@ const configJSON = `{
 	}
 }`
 
+func BeOK[V any](val V, ok bool) V {
```
wagnerd3, line 34: I figured this would be handy to have globally, maybe we should add something like this to https://github.com/sapcc/go-bits/blob/master/must/must.go ?

wagnerd3:
~~Setting to draft, I found one more flaw with the fallback to the database when starting up the collector with the liquid being unavailable right now.~~

internal/db/migrations.go
```diff
@@ -100,4 +100,12 @@ var sqlMigrations = map[string]string{
 	`077_rate_category.down.sql`: `
 		ALTER TABLE rates DROP COLUMN category_id;
 	`,
+	`078_add_commitment_handling_needs_project_metadata.up.sql`: `
+		ALTER TABLE services ADD COLUMN commitment_handling_needs_project_metadata BOOLEAN NOT NULL DEFAULT FALSE;
```
wagnerd3, line 112: I think we forgot to add this in Limes when persisting the service info and we never recognized it, because we did not utilize it for interaction with the liquid.

internal/core/liquid.go
```diff
@@ -55,97 +53,84 @@ func MakeLiquidConnection(lc LiquidConfiguration, serviceType db.ServiceType, av
...
 }
 
 // retrieveServiceInfo queries the backend service for the latest ServiceInfo and validates it.
 // If the liquid is not reachable it can fall back to reading the ServiceInfo from the database
 // - if the dbFallback parameter is set. It is only called on init and when the InfoVersion changes.
-func (l *LiquidConnection) retrieveServiceInfo(ctx context.Context, dbFallback bool) (result liquid.ServiceInfo, apiSuccess bool, err error) {
-	apiSuccess = true
+func (l *LiquidConnection) retrieveServiceInfo(ctx context.Context) (result liquid.ServiceInfo, err error) {
 	result, err = l.LiquidClient.GetInfo(ctx)
 	// result, err := liquid.ServiceInfo{}, errors.New("some error")
```
majewsky: Do we want to clean up this line?

internal/core/service_info_cache.go
```diff
@@ -248,101 +276,236 @@ func deepCloneMap[M ~map[K]V, K comparable, V any](m M, cloneValue func(V) V) M
...
-func (s *ServiceInfoCache) GetServiceForType(serviceType db.ServiceType) db.Service {
+func (s *ServiceInfoCache) GetServiceForType(serviceType db.ServiceType) (db.Service, bool) {
 	s.dataMutex.RLock()
 	defer s.dataMutex.RUnlock()
-	return s.servicesByType[serviceType]
+	val, ok := s.servicesByType[serviceType]
+	return val, ok
+}
+
+// GetServiceForLoc returns the cached resource for the serviceType and resourceName of loc.
```
majewsky:
```suggestion
// GetServiceForLoc returns the cached service for the serviceType of loc.
```

Also, I doubt that `GetServiceForLoc(loc)` is clearer than `GetServiceForType(loc.ServiceType)`. Giving the entire `loc` feels like it would be doing something with all of it instead of just one single field.

internal/core/service_info_cache.go
```diff
@@ -248,101 +276,236 @@ func deepCloneMap[M ~map[K]V, K comparable, V any](m M, cloneValue func(V) V) M
...
-func (s *ServiceInfoCache) GetAZResourceForTypeNameAZ(serviceType db.ServiceType, resourceName liquid.ResourceName, az limes.AvailabilityZone) db.AZResource {
+func (s *ServiceInfoCache) GetAZResourceForTypeNameAZ(serviceType db.ServiceType, resourceName liquid.ResourceName, az limes.AvailabilityZone) (db.AZResource, bool) {
 	s.dataMutex.RLock()
 	defer s.dataMutex.RUnlock()
-	return s.azResourcesByAZByNameByType[serviceType][resourceName][az]
+	val, ok := s.azResourcesByAZNameType[serviceType][resourceName][az]
+	return val, ok
+}
+
+// GetAZResourceForLoc returns the cached resource for the serviceType and resourceName of loc.
```
majewsky:
```suggestion
// GetAZResourceForLoc returns the cached AZ resource for this location.
```

internal/core/service_info_cache_test.go
majewsky: Agreed. Please leave a `TODO: upstream this into go-bits/must, then replace it here`.

internal/test/setup.go
```diff
@@ -247,7 +247,7 @@ func NewSetup(t *testing.T, opts ...SetupOption) Setup {
 		if exists {
 			rateLimits = liquidConfig.RateLimits
 		}
-		_ = must.ReturnT(core.SaveServiceInfoToDB(serviceType, serviceInfo, s.Cluster.Config.AvailabilityZones, rateLimits, s.Clock.Now(), s.DB))(t)
+		must.Succeed(core.SaveServiceInfoToDB(serviceType, serviceInfo, s.Cluster.Config.AvailabilityZones, rateLimits, s.Clock.Now(), s.DB))
```
majewsky:
```suggestion
		must.SucceedT(t, core.SaveServiceInfoToDB(serviceType, serviceInfo, s.Cluster.Config.AvailabilityZones, rateLimits, s.Clock.Now(), s.DB))
```

majewsky (changes requested):
Comments for `internal/{core,datamodel,db,test}`. Also, please rebase to resolve the merge conflict on `internal/db/migrations.go`.

internal/core/liquid.go
wagnerd3: yes, will remove.

internal/core/service_info_cache.go
wagnerd3: I get it, it just felt more consistent to use `ForLoc` everywhere, where we have a `loc`, e.g. in the commitment API. I kept it for now, we can still get rid of this if we don't like it.

internal/collector/scrape.go
```diff
@@ -99,32 +100,34 @@ func (c *Collector) ScrapeJob(registerer prometheus.Registerer) jobloop.Job {
 type projectScrapeTask struct {
 	// data loaded during discoverScrapeTask
 	ProjectService db.ProjectService
-	Service        db.Service
+	// do not use the db.Service directly, as it might get updated during the scrape operation
+	ServiceType db.ServiceType
```
majewsky: If we only hold the ServiceType, then we don't need to hold anything at all here. `discoverScrapeTask` fills this with `labels["service_type"]`, and `processScrapeTask` can get the value in the same way without going through this struct.

majewsky (changes requested):

majewsky (approved):

PR #880, opened 2026-05-05
require resources.topology and rates.topology to be filled

majewsky:
This was not upheld in some test scenarios, which might cause confusion down the line.


wagnerd3 (approved):
Lucky you, I just recently introduced the fallback to `FlatTopology` on app layer with the changing-units PR 😄 https://github.com/sapcc/limes/blob/master/internal/core/cluster.go#L703

I think this should be safe to rollout without migrating any data, as the limit-only rates are always written on startup and should have been migrated.

PR #884, opened 2026-05-11
data metrics v2: add rate metrics

majewsky:
I adjusted the testcases to rename `xOtherRate` to a more plausible name to make the testcase more readable.


internal/core/config.go
```diff
@@ -122,24 +122,24 @@ type ServiceRateLimitConfiguration struct {
 	ProjectDefault []RateLimitConfiguration `json:"project_default"`
 }
 
-// GetGlobalRateLimit returns the default global-level rate limit for a given target type URI and a boolean whether one was found.
-func (svcRlConfig ServiceRateLimitConfiguration) GetGlobalRateLimit(name liquid.RateName) (RateLimitConfiguration, bool) {
+// GetGlobalRateLimit returns the global rate limit for a given rate name, or an error if not found.
```
VoigtS: Small remark, this doesn't return an error if not found, but an empty Option object, same for `GetProjectDefaultRateLimit`. I don't have a big opinion on this tho

VoigtS (dismissed):
The logic looks good to me, I only have a small remark on the `ServiceRateLimitConfiguration` method comments.

VoigtS (approved):

PR #887, opened 2026-05-19
apiv2: proofread and format API documentation

majewsky:
This appears to have been written under the expectation that godoc supports Markdown, but it only supports an extremely limited subset of Markdown (e.g. only first-level headings, not second level or below).

For "API structure", I copied and adjusted the equivalent section from the LIQUID API.


wagnerd3 (approved):
I had made some changes in the direction of this PR already in my branch for the /v2 APIs, but will just rebase onto this and go from here :-)

PR #888, opened 2026-05-21
add new DB column: project_commitments.updated_at

majewsky:
Cortex has requested an `updated_after` filter for commitment listing: When they reconcile their commitment data with ours, they only want to know about what has changed since their last reconciliation run. We will include this filter in the v2 API; this change is a prerequisite.


VoigtS (dismissed):

internal/db/migrations.go
```diff
@@ -122,4 +122,12 @@ var sqlMigrations = map[string]string{
 	`080_drop_rates_from_liquid.down.sql`: `
 		ALTER TABLE rates ADD COLUMN from_liquid BOOLEAN NOT NULL DEFAULT FALSE;
 	`,
+	`081_add_project_commitments_updated_at.up.sql`: `
+		ALTER TABLE project_commitments ADD COLUMN updated_at TIMESTAMPTZ;
+		UPDATE project_commitments SET updated_at = created_at;
```
wagnerd3: Making a better guess and `coalesce` with `superseded_at`, `confirmed_at` and so on is not too hard, should we do that?

wagnerd3 (dismissed):
Hard to judge whether we got all places, but LGTM

internal/db/migrations.go
majewsky: done

wagnerd3 (approved):

PR #890, opened 2026-05-28
replace core.AZResourceLocation with db.AZResourcePath

majewsky:
From the "Xyrill wanted to implement new stuff and ended up doing a refactor instead" department.

Both types do nearly the same thing. Historically, `core.AZResourceLocation` has existed for longer and was pervasive throughout commitment handling code. But `db.AZResourcePath` is a better fit because it implements SQL marshaling and thus is more convenient to use in queries. While replacing `core.AZResourceLocation` with `db.AZResourcePath`, I also rewrote several queries to use `az_resources.path` where I saw that it simplifies the join structure.

The one usecase where `core.AZResourceLocation` remains (with a changed name) is for email notifications, where having a separate type is warranted because those currently use the v1 API names.

### DIFF
--- a/internal/api/commitment.go
+++ b/internal/api/commitment.go
@@ -50,7 +50,7 @@ var (
 	`))
 
 	getAZResourceLocationsQuery = sqlext.SimplifyWhitespace(`
-		SELECT azr.id, s.type, r.name, azr.az
+		SELECT azr.id, azr.path
 		  FROM project_az_resources pazr
 		  JOIN az_resources azr on pazr.az_resource_id = azr.id
 		  JOIN resources r ON azr.resource_id = r.id {{AND r.name = $resource_name}}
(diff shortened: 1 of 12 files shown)


internal/db/refs.go
```diff
@@ -36,6 +36,11 @@ type ResourcePath struct {
 	ResourceName liquid.ResourceName
 }
 
+// InAZ builds an [AZResourcePath] below this ResourcePath.
+func (r ResourcePath) InAZ(az liquid.AvailabilityZone) AZResourcePath {
+	return AZResourcePath{r.ServiceType, r.ResourceName, az}
```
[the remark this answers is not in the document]
majewsky: Irrelevant. The type system prevents mistakes.

internal/db/refs.go
```diff
@@ -74,6 +79,11 @@ type AZResourcePath struct {
 	AvailabilityZone liquid.AvailabilityZone
 }
 
+// Resource extracts the [ResourcePath] portion of this AZResourcePath.
+func (r AZResourcePath) Resource() ResourcePath {
+	return ResourcePath{r.ServiceType, r.ResourceName}
```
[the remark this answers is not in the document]
majewsky: Irrelevant. The type system prevents mistakes.

internal/api/commitment.go
```diff
@@ -1602,13 +1596,13 @@ func (p *v1Provider) TransferCommitment(w http.ResponseWriter, r *http.Request)
...
 	)
-	err = p.DB.QueryRow(findAZResourceIDByLocationQuery, targetProject.ID, loc.ServiceType, loc.ResourceName, loc.AvailabilityZone).
+	err = p.DB.QueryRow(findAZResourceIDByLocationQuery, targetProject.ID, path).
 		Scan(&azResourceID, &resourceAllowsCommitments, &targetTotalConfirmed)
 	if respondwith.ObfuscatedErrorText(w, err) {
 		return
 	}
 	if !resourceAllowsCommitments {
-		msg := fmt.Sprintf("resource %s/%s is not enabled in the target project", loc.ServiceType, loc.ResourceName)
+		msg := fmt.Sprintf("resource %s is not enabled in the target project", path.Resource().String())
```
wagnerd3, line 1605: AFAIK we could get rid of the `String()` part, when `fmt.Stringer` is implemented this is called implicitly?
I think it's still well readable when doing that, do we want to do that consistently everywhere?

internal/datamodel/confirm_project_commitments.go
```diff
@@ -96,17 +94,17 @@ func CanAcceptCommitmentChangeRequest(req liquid.CommitmentChangeRequest, loc co
...
 		stats := statsByAZ[req.AZ]
 
-		behavior := cluster.CommitmentBehaviorForResource(loc.ServiceType, resourceName)
-		logg.Debug("checking additions in %s: overall amount %d",
-			loc.ShortScopeString(), resourceName, additionSum)
-		logg.Debug("checking subtractions in %s: overall amount %d",
-			loc.ShortScopeString(), resourceName, subtractionSum)
+		behavior := cluster.CommitmentBehaviorForResource(serviceType, resourceName)
+		logg.Debug("checking additions in %s/%s: overall amount %d",
+			serviceType, resourceName, additionSum)
```
wagnerd3: To keep consistent formatting for paths and shorten the log writing, we could also build a 
`path := db.ResourcePath{ServiceType: serviceType, ResourceName: resourceName}` at the top of the iteration once and then subsequently do

```suggestion
		logg.Debug("checking additions in %s: overall amount %d", path, additionSum)
```

internal/datamodel/confirm_project_commitments.go
```diff
@@ -260,10 +257,10 @@ func generateConfirmationMails(mailTemplate Option[core.MailTemplate], dbi db.In
 			Commitment: *c,
 			DateString: confirmedAt.Format(time.DateOnly),
 			// TODO: we actually don't want to have api-named props in AZResourceLocation. Replace the template and the code simultaneously.
```
wagnerd3: I think the direction we are now going in makes this `TODO` obsolete. We noted this back when we still had DB and api-facing namings mixed in this same struct, now it is nicely divided :-)

majewsky: I would like to keep _an_ `TODO` to remove `AZResourceLocationV1` entirely when the v2 API is ready and we switch mails to refer to resources by the v2 names instead. I have moved and updated the TODO thusly.

wagnerd3 (approved):

PR #894, opened 2026-06-04
fix association of leftover commitment with its parent, option 2

majewsky:
This is the minimal fix for the problem, based on the fact that the original bug stems from misunderstanding the semantics of indices returned by slices.Backward(). We need to merge either this or #893.


wagnerd3:
I think the other implementation is better, because we can get rid of the size check by using `option.Option`, so I merged that one.

PR #896, opened 2026-06-05
v2 API: base building

majewsky:
I broke off several pieces from the `v2-commitment-api` branch that will also be relevant to @wagnerd3's parallel work. Please look at the commit messages for details.

### DIFF
--- a/internal/api/api_v2/core.go
+++ b/internal/api/api_v2/core.go
@@ -4,6 +4,10 @@
 package api_v2
 
 import (
+	"cmp"
+	"database/sql"
+	"errors"
+	"fmt"
 	"net/http"
 	"time"
 
@@ -15,6 +19,7 @@ import (
 	"github.com/sapcc/go-bits/respondwith"
 
 	"github.com/sapcc/limes/internal/core"
+	"github.com/sapcc/limes/internal/db"
 )
 
 type v2Provider struct {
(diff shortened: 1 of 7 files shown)


internal/api/api_v2/core.go
```diff
@@ -35,8 +36,25 @@ func NewV2API(cluster *core.Cluster, tokenValidator gopherpolicy.Validator, audi
...
+	r.Methods("GET").Path("/rates/v2/info").HandlerFunc(handlerFunc(http.StatusOK, p.handleGetRatesInfo))
+}
+
+// Wrapper for request handlers that enforces a structure,
+// where the request handler function has an error return
+// instead of being able to randomly use `w` at any point.
+func handlerFunc[T any](successCode int, action func(r *http.Request) (T, error)) http.HandlerFunc {
+	return func(w http.ResponseWriter, r *http.Request) {
+		resp, err := action(r)
+		if respondwith.ObfuscatedErrorText(w, err) {
```
wagnerd3: This handler will only work for customer-facing APIs, because `ObfuscatedErrorText` is used. Can we add a bool-option for the error handling to handle "internal" APIs? E.g., I imagine it would be good to switch to this pattern for `/admin` APIs too, and there we want to return `error` unobfuscated. Wait... I looked the APIs up and we are also using `ObfuscatedErrorText` there. But I think this would be our chance to change this?

internal/api/api_v2/core.go
```diff
@@ -57,17 +62,67 @@ func handlerFunc[T any](successCode int, action func(r *http.Request) (T, error)
...
 // during this function are deferred until Require() is called.
-func (p *v2Provider) CheckToken(r *http.Request) *gopherpolicy.Token {
-	return CheckToken(r, p.tokenValidator)
+func (p *v2Provider) checkToken(r *http.Request) *gopherpolicy.Token {
+	return p.tokenValidator.CheckToken(r)
+}
+
+// checkProjectAccess authenticates and authorizes a project-scoped request using the given policy rule.
+// On success, returns the database records for the project scope, its containing domain and the authenticated token.
+func (p *v2Provider) checkProjectAccess(r *http.Request, projectUUID, policyRule string) (_ db.Domain, _ db.Project, _ *gopherpolicy.Token, err error) { //nolint:unused // TODO: remove this nolint once it is used
```
wagnerd3: I think for the functionality that we need for all reports endpoints, this is not complete. IIRC when we talked about this, we wanted to allow a `?domain_id=...` parameter on `/project` APIs. I know that we could put another function besides this and then switch within the endpoint, but I thought it would be nicer to have 1 "`Scope`-checking" function which you can configure to work for all different cases that we want to allow.

I already worked on this a little bit, and my idea was to have some kind of `ScopeConfig` which tells the "access-checker" what to allow in terms of params from the URL/token.
I started this on my branch here: [`master...v2-apis#diff-37238fe8a7`](https://github.com/sapcc/limes/compare/master...v2-apis#diff-37238fe8a7244bd9bf217ac6d061158cc1dfd3ccfe707cd4b507e39eb9c1d2a5R22-R112)
I am fine with the general structure to combine the `CheckToken` call into this, but I think we should solve the part about where to get the `Scope` from (what you hardcoded to `projectUUID`)?

wagnerd3 (changes requested):

internal/api/api_v2/core.go
majewsky: After talking about it, the report endpoints have a sufficiently different usecase to legitimately not be able to use these functions. We will have `checkProjectAccess` (for use by project-scoped endpoints like commitment handling and set-rate-limit or set-max-quota) be completely separate from `Scope` (for use by report endpoints, which should be renamed to `ReportScope` for clarity). Both are completely separate from the special case AuthZ handling for the info endpoints, which is sufficiently different yet again.

internal/api/api_v2/core.go
majewsky: Updated the wrapper to do the token checking and decide on obfuscation or not.

wagnerd3 (approved):

PR #897, opened 2026-06-08
implement POST /resources/v2/commitments/new

majewsky:
This has _some_ preparations for status "guaranteed" in it, but I decided to reject this status for now because handling it properly will require changes to `package datamodel` that I want to avoid within this epic.

### DIFF
--- a/internal/api/api_v2/commitment_create.go
+++ b/internal/api/api_v2/commitment_create.go
@@ -160,6 +160,7 @@ func (p *v2Provider) handlePostNewCommitment(r *http.Request, token *gopherpolic
 			}
 		} else {
 			ccr := liquid.CommitmentChangeRequest{
+				DryRun:      req.DryRun,
 				AZ:          path.AvailabilityZone,
 				InfoVersion: must.BeOK(sis.GetServiceForType(path.ServiceType)).LiquidVersion,
 				ByProject: map[liquid.ProjectUUID]liquid.ProjectCommitmentChangeset{
--- a/internal/api/api_v2/commitment_create.go
+++ b/internal/api/api_v2/commitment_create.go
@@ -1,0 +1,257 @@
+// SPDX-FileCopyrightText: 2026 SAP SE or an SAP affiliate company
+// SPDX-License-Identifier: Apache-2.0
+
+package api_v2
+
+import (
+	"encoding/json"
+	"fmt"
+	"net/http"
+	"time"
+
+	"github.com/go-gorp/gorp/v3"
+	"github.com/sapcc/go-api-declarations/cadf"
+	"github.com/sapcc/go-api-declarations/liquid"
+	"github.com/sapcc/go-bits/audittools"
+	"github.com/sapcc/go-bits/gopherpolicy"
+	"github.com/sapcc/go-bits/httpapi"
+	"github.com/sapcc/go-bits/logg"
+	"github.com/sapcc/go-bits/must"
+	"github.com/sapcc/go-bits/respondwith"
+	"github.com/sapcc/go-bits/sqlext"
+	. "go.xyrillian.de/gg/option"
+	"go.xyrillian.de/gg/options"
+
+	resourcesv2 "github.com/sapcc/limes/internal/apideclarations/apiv2/resources"
+	"github.com/sapcc/limes/internal/audit"
+	"github.com/sapcc/limes/internal/core"
+	"github.com/sapcc/limes/internal/datamodel"
+	"github.com/sapcc/limes/internal/db"
+	"github.com/sapcc/limes/internal/util"
+)
+
+func (p *v2Provider) handlePostNewCommitment(r *http.Request, token *gopherpolicy.Token) (resourcesv2.Commitment, error) {
+	httpapi.IdentifyEndpoint(r, "/resources/v2/commitments/new")
+	var (
+		none resourcesv2.Commitment // used on error return paths only
+		ctx  = r.Context()
+		sis  = p.Cluster.SIC.GetSnapshot()
+	)
+
+	// parse request
+	req, err := parseRequestBodyAs[resourcesv2.CommitmentRequest](r)
+	if err != nil {
+		return none, err
+	}
+	path := db.AZResourcePath{
+		ServiceType:      req.ServiceType,
+		ResourceName:     req.ResourceName,
+		AvailabilityZone: req.AvailabilityZone,
+	}
+	attrs := commitmentStatusAttributes{
+		Status:          req.Status,
+		ConfirmBy:       options.Map(req.ConfirmBy, util.FromUnixEncodedTime),
+		NotifyOnConfirm: req.NotifyOnConfirm,
+	}
+	now := p.timeNow()
+
+	// validate request contents
+	dbDomain, dbProject, err := p.checkProjectAccess(token, req.ProjectUUID, "v2:project:commitment_create")
+	if err != nil {
+		return none, err
+	}
+	azResource, behavior, err := p.validateCommittability(path, dbDomain, dbProject, req.Duration, sis)
+	if err != nil {
+		return none, err
+	}
+	if req.Amount == 0 {
+		return none, respondwith.CustomStatus(http.StatusUnprocessableEntity, errEmptyAmount)
+	}
+	err = p.validateStatusAttributesOnNewCommitment(attrs, behavior, now)
+	if err != nil {
+		return none, err
+	}
+
+	// prepare commitment and commitment request
+	creationContextJSON, err := json.Marshal(db.CommitmentWorkflowContext{Reason: db.CommitmentReasonCreate})
+	if err != nil {
+		return none, err
+	}
+	c := db.ProjectCommitment{
+		UUID:                datamodel.GenerateProjectCommitmentUUID(),
+		AZResourceID:        azResource.ID,
+		ProjectID:           dbProject.ID,
+		Amount:              req.Amount,
+		Duration:            req.Duration,
+		CreatedAt:           now,
+		UpdatedAt:           now,
+		CreatorUUID:         token.UserUUID(),
+		CreatorName:         fmt.Sprintf("%s@%s", token.UserName(), token.UserDomainName()),
+		ConfirmBy:           None[time.Time](), // may be set below
+		ConfirmedAt:         None[time.Time](), // may be set below
+		ExpiresAt:           req.Duration.AddTo(attrs.ConfirmBy.UnwrapOr(now)),
+		CreationContextJSON: json.RawMessage(creationContextJSON),
+		Status:              req.Status,
+		NotifyOnConfirm:     req.NotifyOnConfirm,
+	}
+	switch c.Status {
+	case liquid.CommitmentStatusConfirmed:
+		// special case: need to insert this with a different status initially;
+		// otherwise TransferableCommitmentCache will count it as an existing
+		// confirmed commitment when gathering stats (will change this later)
+		c.Status = liquid.CommitmentStatusPending
+	case liquid.CommitmentStatusPending:
+		c.ConfirmBy = Some(now)
+	default:
+		c.ConfirmBy = attrs.ConfirmBy
+	}
+
+	// using a transaction because we need to insert the commitment first to get
+	// its ID (for use in the SupersedeContext of consumed commitments),
+	// but we need to be able to revert this insertion if the CommitmentChangeRequest is rejected
+	var auditEvents []audittools.Event
+	err = withinTransactionThatMayBeADryRun(p.DB, req.DryRun, func(tx db.Interface) error {
+		stats, err := getCommitmentStats(tx, dbProject.ID, azResource.ID)
+		if err != nil {
+			return err
+		}
+
+		err = tx.Insert(&c)
+		if err != nil {
+			return err
+		}
+
+		var (
+			auditContext = audit.Context{UserIdentity: token, Request: r}
+		)
+
+		// creation of confirmed commitments requires a special codepath because we might be consuming other commitments
+		if req.Status == liquid.CommitmentStatusConfirmed {
+			mailTemplate := None[core.MailTemplate]()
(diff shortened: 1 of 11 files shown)


internal/api/api_v2/commitment_create.go
```diff
@@ -0,0 +1,256 @@
...
+		AvailabilityZone: req.AvailabilityZone,
+	}
+	attrs := commitmentStatusAttributes{
+		Status:          req.Status,
+		ConfirmBy:       options.Map(req.ConfirmBy, util.FromUnixEncodedTime),
+		NotifyOnConfirm: req.NotifyOnConfirm,
+	}
+	now := p.timeNow()
+
+	// validate request contents
```
VoigtS, line 58: shouldn't we check for the access before we do anything with the request body?

internal/api/api_v2/commitment_create.go
```diff
@@ -0,0 +1,256 @@
...
+		}
+
+		err = tx.Insert(&c)
+		if err != nil {
+			return err
+		}
+
+		var (
+			auditContext = audit.Context{UserIdentity: token, Request: r}
+			resources, _ = p.Cluster.SIC.GetResourcesForType(path.ServiceType) // TODO: ugly (we just pass this around; maybe callees should fetch this themselves)
```
VoigtS: should we check the `ok` value here regardless?

internal/api/api_v2/commitment_create_test.go
```diff
@@ -0,0 +1,745 @@
...
+	firstCapacityID := s.GetResourceID("first", "capacity")
+	for az, capacity := range map[liquid.AvailabilityZone]uint64{
+		"az-one": 20,
+		"az-two": 30,
+		"total":  50,
+	} {
+		s.MustDBExec(`UPDATE az_resources SET raw_capacity = $1 WHERE az = $2 AND resource_id = $3`,
+			capacity, az, firstCapacityID)
+	}
+	dresdenID := s.GetProjectID("berlin")
```
VoigtS: This looks weird.

VoigtS (changes requested):

internal/api/api_v2/commitment_create.go
majewsky: That is exactly what happens. The first call in this section is `checkProjectAccess`, which we do as soon as possible (we just need to parse the request body first because we get the project scope from there).

internal/api/api_v2/commitment_create.go
majewsky: It does not really make sense. The important part would be not that `resources` has anything at all, but rather that `resources[path.ResourceName]` exists. I will validate this at the usage site of `resources` in DelegateChangeCommitments.

internal/api/api_v2/commitment_create_test.go
majewsky: Yep, and what's even weirder is that the testcase below was nonsensical because of it. Fixed.

VoigtS (dismissed):

internal/api/api_v2/core.go
```diff
@@ -89,9 +93,38 @@ func handlerFunc[T any](successCode int, tv gopherpolicy.Validator, action func(
...
+
+	dec := json.NewDecoder(bytes.NewReader(buf))
+	dec.DisallowUnknownFields()
+	err = dec.Decode(&result)
+	if err != nil {
+		err = respondwith.CustomStatus(http.StatusBadRequest,
+			fmt.Errorf("request body is not valid JSON: %w", err),
+		)
+	}
+	return result, err
```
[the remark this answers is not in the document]
majewsky: I considered upping the const by 1, but I'm hoping that not extending that will keep the allocation aligned on page boundaries, which might be a bit more efficient.

internal/apideclarations/apiv2/resources/commitment.go
```diff
@@ -0,0 +1,84 @@
...
+
+	// Status must be one of:
+	//   - liquid.CommitmentStatusPlanned
+	//   - liquid.CommitmentStatusPending
+	//   - liquid.CommitmentStatusConfirmed
+	//   - liquid.CommitmentStatusGuaranteed
+	Status liquid.CommitmentStatus `json:"status"`
+	// ConfirmBy must be set for statuses "planned" and "guaranteed", and may not be set otherwise.
+	// Commitments created in status "pending" will have a ConfirmBy value equal to the current time.
+	ConfirmBy Option[limes.UnixEncodedTime] `json:"confirm_by,omitzero"`
```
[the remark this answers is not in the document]
majewsky: Intentional, because `package apideclarations` will move out into the go-api-declarations repo, and we will definitely forgot to update the docstring there once we do add support for `guaranteed`. This is the same way we're handling it for the LIQUID API right now.

internal/api/api_v2/commitment_create.go
```diff
@@ -0,0 +1,257 @@
...
+	"github.com/sapcc/limes/internal/core"
+	"github.com/sapcc/limes/internal/datamodel"
+	"github.com/sapcc/limes/internal/db"
+	"github.com/sapcc/limes/internal/util"
+)
+
+func (p *v2Provider) handlePostNewCommitment(r *http.Request, token *gopherpolicy.Token) (resourcesv2.Commitment, error) {
+	httpapi.IdentifyEndpoint(r, "/resources/v2/commitments/new")
+	var (
+		none resourcesv2.Commitment // used on error return paths only
```
wagnerd3, line 36: I would prefer a named return parameter `none` and then keeping the separate `result` at the end - just preference.

internal/api/api_v2/commitment_helpers.go
```diff
@@ -0,0 +1,209 @@
...
+	"go.xyrillian.de/gg/options"
+
+	resourcesv2 "github.com/sapcc/limes/internal/apideclarations/apiv2/resources"
+	"github.com/sapcc/limes/internal/core"
+	"github.com/sapcc/limes/internal/db"
+	"github.com/sapcc/limes/internal/util"
+)
+
+var (
+	errAZMayNotBeAny             = errors.New(`resource is AZ-aware, so the AZ may not be set to "any"`)
```
wagnerd3: Is there a reason to not use the `must` verb here?

```suggestion
	errAZMustNotBeAny             = errors.New(`resource is AZ-aware, so the AZ may not be set to "any"`)
```

internal/apideclarations/apiv2/resources/commitment.go
```diff
@@ -0,0 +1,84 @@
...
+	// NotifyOnConfirm may not be set for commitments that are created in status "confirmed".
+	NotifyOnConfirm bool `json:"notify_on_confirm,omitempty"` // may not be set for "confirmed"
```
wagnerd3: Double comment?
```suggestion
	NotifyOnConfirm bool `json:"notify_on_confirm,omitempty"`
```

internal/apideclarations/apiv2/resources/commitment.go
```diff
@@ -0,0 +1,84 @@
...
+	ProjectUUID      string                 `json:"project_id"`
+	ServiceType      db.ServiceType         `json:"service_type"`
+	ResourceName     liquid.ResourceName    `json:"resource_name"`
+	AvailabilityZone limes.AvailabilityZone `json:"availability_zone"`
+
+	// Status must be one of:
+	//   - liquid.CommitmentStatusPlanned
+	//   - liquid.CommitmentStatusPending
+	//   - liquid.CommitmentStatusConfirmed
+	//   - liquid.CommitmentStatusGuaranteed
```
wagnerd3: This is against the current implementation, the customer will get `errInvalidInitialStatus` back.
Someone looking into the new `/v2` API when it's "GA" might see this and wonder (or ask about the time horizon on this feature). You probably don't want a `TODO` in docs?

internal/api/api_v2/commitment_create.go
```diff
@@ -0,0 +1,257 @@
...
+	result := convertCommitmentToDisplayForm(c, path, dbProject, canBeDeleted)
+	if req.DryRun {
+		result.UUID = "00000000-0000-0000-0000-000000000000"
+	}
+	return result, nil
+}
+
+// withinTransactionThatMayBeADryRun starts a transaction such that the type system helps enforce that a dry run is not accidentally committed:
+// Within `action`, `tx` is only a generic interface handle that does not allow calling `tx.Commit()` directly.
+func withinTransactionThatMayBeADryRun(dbm *gorp.DbMap, dryRun bool, action func(tx db.Interface) error) error {
```
wagnerd3:
```suggestion
func withinDryRunnableTx(dbm *gorp.DbMap, dryRun bool, action func(tx db.Interface) error) error {
```

internal/api/api_v2/commitment_create_test.go
```diff
@@ -0,0 +1,759 @@
...
+	}
+}`
+
+// Helper function for a successful commitment creation. This always tests the dry-run first, and then does the actual creation.
+func createCommitmentAndExpectSuccess(t *testing.T, s test.Setup, tr *easypg.Tracker, ccrAcceptedByLiquid bool, request map[string]any, expected jsonmatch.Object, getTarget func() cadf.Resource) {
+	t.Helper()
+	ctx := t.Context()
+	mockLiquid := s.LiquidClients[db.ServiceType(request["service_type"].(string))]
+
+	if ccrAcceptedByLiquid {
```
wagnerd3:
```suggestion
	if liquidHandlesCommitments {
```

internal/api/api_v2/commitment_create_test.go
```diff
@@ -0,0 +1,759 @@
...
+			"duration":          "1 hour",
+			"project_id":        "uuid-for-berlin",
+			"service_type":      "first",
+			"resource_name":     "capacity",
+			"availability_zone": "az-one",
+			"status":            status,
+		}, func(r httptest.Response) {
+			r.ExpectText(t, http.StatusUnprocessableEntity, "confirm_by must be set for the requested initial commitment status\n")
+		})
+	}
```
wagnerd3:
```suggestion
	createCommitmentAndExpectError(t, s, tr, map[string]any{
		"amount":            1,
		"duration":          "1 hour",
		"project_id":        "uuid-for-berlin",
		"service_type":      "first",
		"resource_name":     "capacity",
		"availability_zone": "az-one",
		"status":            "planned",
	}, func(r httptest.Response) {
		r.ExpectText(t, http.StatusUnprocessableEntity, "confirm_by must be set for the requested initial commitment status\n")
	})
```

internal/apideclarations/apiv2/resources/commitment.go
```diff
@@ -0,0 +1,84 @@
...
+	TransferToken  Option[string]                          `json:"transfer_token,omitzero"`
+
+	CreatedAt limes.UnixEncodedTime `json:"created_at"`
+	// CreatorUUID and CreatorName identify the user who created this commitment.
+	// CreatorName is in the format `fmt.Sprintf("%s@%s", userName, userDomainName)`and intended for informational displays only.
+	CreatorUUID string `json:"creator_uuid,omitempty"`
+	CreatorName string `json:"creator_name,omitempty"`
+	// CanBeDeleted will be true if the commitment can be deleted by the same user who saw this object in response to a GET query.
+	CanBeDeleted bool `json:"can_be_deleted,omitempty"`
+	// ConfirmBy is unset if and only if the commitment was created in status "confirmed".
```
wagnerd3, line 39: This is subject to change with `guaranteed` - so the current documentation is a mix of 2 states (as-is and to-be with guaranteed) → see my other comment.

internal/apideclarations/apiv2/resources/commitment.go
```diff
@@ -0,0 +1,84 @@
...
+	// This means that a new commitment was created that will be confirmed when this commitment is set to expire.
+	WasRenewed bool `json:"was_renewed,omitempty"`
+}
+
+// CommitmentRequest is the request payload format for POST /v2/commitments/new.
+//
+// See documentation on [Commitment] for the semantics of all fields.
+// Documentation on this type's fields only mentions specifics related to the commitment creation process.
+type CommitmentRequest struct {
+	// DryRun can be set to true to avoid any side effects.
```
wagnerd3:
```suggestion
	// DryRun can be set to true to avoid any side effects - no data is saved in the system.
```

wagnerd3 (changes requested):

majewsky:
FYI: This force-rebase ^ is to bring in the ServiceInfoSnapshot change from #898. It does not change anything else. The only manual changes in it are within `internal/api/api_v2/commitment_*.go`, to use the new snapshot-based API.

internal/api/api_v2/commitment_create.go
majewsky: If we're talking preference, I would not like `none` to show up in the function signature because that would also be what appears in the package documentation (if the function was not private). An implementation detail of the function should not be in its signature.

internal/api/api_v2/commitment_helpers.go
majewsky: I was under the impression that RFC 2119 defines "MAY NOT", but it is indeed "MUST NOT" there, and that is some good precedent to follow.

internal/api/api_v2/commitment_create_test.go
majewsky: I'm adding a comment explaining that "guaranteed" needs to be added as an option there. That is why I originally built it as a loop.

internal/apideclarations/apiv2/resources/commitment.go
majewsky: No. When creating in status `guaranteed`, you need to set `confirm_by`, which is consistent with what this comment says.

wagnerd3: The status is now that the full documentation considers `guaranteed` to be implemented already, except for the one `TODO` on the enumeration of states.

wagnerd3 (dismissed):
My points are all fine, nice :-)

wagnerd3 (approved):

PR #898, opened 2026-06-09
introduce ServiceInfoSnapshot

wagnerd3:
The idea of this PR is to introduce the `ServiceInfoSnapshot`, which is a coherent set of objects of types `db.Service, db.Resource, db.AZResource, db.Rate, db.Category`.
From this, a `FilteredServiceInfoSnapshot` can be built, which has the same method-set for access as `ServiceInfoSnapshot`, denoted by interface `ServiceInfoReader`.

This change solves 2 things:
* when accessing `ServiceInfoCache` in the past, the cache could have theoretically changed between calls to different entities. 
* When doing all the reporting for `/v2`, we can operate on a `FilteredServiceInfoSnapshot`, which we can directly build from query params.

I have a few questions:
* Do we want to pass `ServiceInfoSnapshot` as pointer like I did? I figured this would cause much less memory load...
* Do we want to reserve usage of `FilteredServiceInfoSnapshot` for cases, where we iterate the `services, resources, ....` etc. - which is mostly reports? When accessing individual objects, operating on a filtered version does not really make sense. I sometimes used the filtered version to express, that we evaluated certain contents are there (e.g. on `/commitment`), but this does not really benefit us, because when accessing the individual `service/resource`, the call-sites don't get shorter.
* bikeshedding of naming

### DIFF
--- a/internal/api/commitment.go
+++ b/internal/api/commitment.go
@@ -223,7 +223,7 @@ func (p *v1Provider) GetPublicCommitments(w http.ResponseWriter, r *http.Request
 		return
 	}
 
-	resource, rExists := sis.GetResourceForTypeName(dbServiceType, dbResourceName)
+	resource, rExists := sis.GetResourceForPath(db.ResourcePath{ServiceType: dbServiceType, ResourceName: dbResourceName})
 	// this check is defense in depth (when a name mapping was found, the resource must be found too)
 	if !rExists {
 		msg := fmt.Sprintf("could not load underlying resource: %q", resource.Path)
@@ -329,6 +329,11 @@ func (p *v1Provider) parseAndValidateCommitmentRequest(w http.ResponseWriter, r 
 		http.Error(w, msg, http.StatusUnprocessableEntity)
 		return nil, nil, nil, filteredSIS
 	}
+	path := db.AZResourcePath{
+		ServiceType:      dbServiceType,
+		ResourceName:     dbResourceName,
+		AvailabilityZone: req.AvailabilityZone,
+	}
 	filteredSIS = sis.Filter(core.ServiceInfoFilter{ServiceType: Some(dbServiceType), ResourceName: Some(dbResourceName)})
 
 	behavior := p.Cluster.CommitmentBehaviorForResource(dbServiceType, dbResourceName).ForDomain(dbDomain.Name)
@@ -336,7 +341,7 @@ func (p *v1Provider) parseAndValidateCommitmentRequest(w http.ResponseWriter, r 
 		http.Error(w, "commitments are not enabled for this resource", http.StatusUnprocessableEntity)
 		return nil, nil, nil, filteredSIS
 	}
-	resource := must.BeOK(filteredSIS.GetResourceForTypeName(dbServiceType, dbResourceName))
+	resource := must.BeOK(filteredSIS.GetResourceForPath(path.Resource()))
 	if resource.Topology == liquid.FlatTopology {
 		if req.AvailabilityZone != limes.AvailabilityZoneAny {
 			http.Error(w, `resource does not accept AZ-aware commitments, so the AZ must be set to "any"`, http.StatusUnprocessableEntity)
@@ -359,11 +364,6 @@ func (p *v1Provider) parseAndValidateCommitmentRequest(w http.ResponseWriter, r 
 		return nil, nil, nil, filteredSIS
 	}
 
-	path := db.AZResourcePath{
-		ServiceType:      dbServiceType,
-		ResourceName:     dbResourceName,
-		AvailabilityZone: req.AvailabilityZone,
-	}
 	// when the name mapping succeeded, service and resource must be found
 	return &req, &path, &behavior, filteredSIS
 }
@@ -1756,7 +1756,7 @@ func (p *v1Provider) GetCommitmentConversions(w http.ResponseWriter, r *http.Req
 		http.Error(w, msg, http.StatusUnprocessableEntity)
 		return
 	}
-	sourceResource, rExists := sis.GetResourceForTypeName(sourceServiceType, sourceResourceName)
+	sourceResource, rExists := sis.GetResourceForPath(db.ResourcePath{ServiceType: sourceServiceType, ResourceName: sourceResourceName})
 	if !rExists {
 		http.Error(w, "service not found", http.StatusNotFound)
 		return
--- a/internal/api/commitment.go
+++ b/internal/api/commitment.go
@@ -310,7 +310,7 @@ func (p *v1Provider) GetPublicCommitments(w http.ResponseWriter, r *http.Request
 // parseAndValidateCommitmentRequest parses and validates the request body for a commitment creation or confirmation.
 // This function in its current form should only be used if the serviceInfo is not necessary to be used outside
 // of this validation to avoid unnecessary database queries.
-func (p *v1Provider) parseAndValidateCommitmentRequest(w http.ResponseWriter, r *http.Request, dbDomain db.Domain) (_ *limesresources.CommitmentRequest, _ *db.AZResourcePath, _ *core.ScopedCommitmentBehavior, filteredSIS *core.FilteredServiceInfoSnapshot) {
+func (p *v1Provider) parseAndValidateCommitmentRequest(w http.ResponseWriter, r *http.Request, dbDomain db.Domain) (_ *limesresources.CommitmentRequest, _ *db.AZResourcePath, _ *core.ScopedCommitmentBehavior, filteredSIS core.FilteredServiceInfoSnapshot) {
 	// parse request
 	var parseTarget struct {
 		Request limesresources.CommitmentRequest `json:"commitment"`
--- a/internal/api/commitment.go
+++ b/internal/api/commitment.go
@@ -731,7 +731,7 @@ func (p *v1Provider) MergeProjectCommitments(w http.ResponseWriter, r *http.Requ
 	sis := p.Cluster.SIC.GetSnapshot()
 	resource, rExists := sis.GetResourceForPath(path.Resource())
 	if !rExists { // checking the deepest level is enough
-		http.Error(w, "service not found", http.StatusNotFound)
+		http.Error(w, "service or resource not found", http.StatusNotFound)
 		return
 	}
 	service := must.BeOK(sis.GetServiceForType(path.ServiceType))
@@ -949,7 +949,7 @@ func (p *v1Provider) RenewProjectCommitments(w http.ResponseWriter, r *http.Requ
 	sis := p.Cluster.SIC.GetSnapshot()
 	resource, rExists := sis.GetResourceForPath(path.Resource())
 	if !rExists { // checking the deepest level is enough
-		http.Error(w, "service not found", http.StatusNotFound)
+		http.Error(w, "service or resource not found", http.StatusNotFound)
 		return
 	}
 	service := must.BeOK(sis.GetServiceForType(path.ServiceType))
@@ -1101,7 +1101,7 @@ func (p *v1Provider) DeleteProjectCommitment(w http.ResponseWriter, r *http.Requ
 	sis := p.Cluster.SIC.GetSnapshot()
 	_, rExists := sis.GetResourceForPath(path.Resource())
 	if !rExists { // checking the deepest level is enough
-		http.Error(w, "service not found", http.StatusNotFound)
+		http.Error(w, "service or resource not found", http.StatusNotFound)
 		return
 	}
 	service := must.BeOK(sis.GetServiceForType(path.ServiceType))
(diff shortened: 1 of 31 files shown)


internal/core/service_info_cache.go
```diff
@@ -23,34 +23,496 @@ import (
...
+// ServiceInfoFilter is a set of possible attributes by which ServiceInfoSnapshot
+// can be filtered. Most of the time, these come directly or indirectly from an API
+// requests' scope, e.g. query filters or attributes of a processed commitment.
+type ServiceInfoFilter struct {
+	ServiceArea  Option[string]
+	ServiceType  Option[db.ServiceType]
+	Category     Option[liquid.CategoryName]
+	ResourceName Option[liquid.ResourceName]
+	RateName     Option[liquid.RateName]
+}
```
majewsky, line 37: Does the v2 API not allow listing multiple values for each of these fields? The v1 API allows e.g. `?service=network&resource=security_groups&resource=security_group_rules`, and this is indeed used (e.g. by the Elektra quota_usage display).
```suggestion
type ServiceInfoFilter struct {
	ServiceAreas  Option[[]string]
	ServiceTypes  Option[[]db.ServiceType]
	Categories    Option[[]liquid.CategoryName]
	ResourceNames Option[[]liquid.ResourceName]
	RateNames     Option[[]liquid.RateName]
}
```

internal/core/service_info_cache.go
```diff
@@ -23,34 +23,496 @@ import (
...
+
+///////////////////////////////////////////////////////////////////////
+
+// ServiceInfoReader defines shared methods for reading
+// filtered or unfiltered ServiceInfoSnapshots.
+type ServiceInfoReader interface {
```
majewsky, line 44: I think we can lose all the `...ForTypeName` and `...ForTypeNameAZ` methods. I replaced most usages with `...ForPath`. A quick grep suggests that only one usage remains (except for test coverage in `service_info_cache_test.go`).

internal/core/service_info_cache.go
```diff
@@ -23,34 +23,496 @@ import (
...
 	RatesByName map[liquid.RateName]db.Rate
-	// CategoriesByID are all categories, indexed by their id.
+	// CategoriesByID are categories, indexed by their id.
 	CategoriesByID map[db.CategoryID]db.Category
 )
 
+// removeType removes all data of a given serviceType, making the cache ready
+// for populating this serviceType from scratch. Categories are always flushed
+// completely.
+func (s *ServiceInfoSnapshot) removeType(serviceType db.ServiceType) {
```
majewsky: Naming suggestion:
```suggestion
func (s *ServiceInfoSnapshot) removeDataForType(serviceType db.ServiceType) {
```

internal/core/service_info_cache.go
```diff
@@ -23,34 +23,496 @@ import (
...
+	}
+	if serviceType, ok := filter.ServiceType.Unpack(); ok && f.filter.ServiceType.IsNone() {
+		f.filter.ServiceType = Some(serviceType)
+	}
+	if resourceCategory, ok := filter.Category.Unpack(); ok && f.filter.Category.IsNone() {
+		f.filter.Category = Some(resourceCategory)
+	}
+	if azResource, ok := filter.ResourceName.Unpack(); ok && f.filter.ResourceName.IsNone() {
+		f.filter.ResourceName = Some(azResource)
+	}
```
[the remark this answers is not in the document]
wagnerd3: solved after adding tests for it.

internal/core/name_mapping.go
```diff
@@ -37,12 +37,12 @@ type dbRateRef struct {
...
 
 // BuildResourceNameMapping constructs a new ResourceNameMapping instance.
-func BuildResourceNameMapping(cluster *Cluster, resources ResourcesByNameType) ResourceNameMapping {
+func BuildResourceNameMapping(cluster *Cluster, sis *ServiceInfoSnapshot) ResourceNameMapping {
 	nm := ResourceNameMapping{
 		fromAPIToDB: make(map[ResourceRef]dbResourceRef),
 		fromDBToAPI: make(map[dbResourceRef]ResourceRef),
 	}
-	for dbServiceType, resByName := range resources {
+	for dbServiceType, resByName := range sis.GetResources() {
```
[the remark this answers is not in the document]
wagnerd3: This is on purpose. The first clone is to make it easier to filter down the data.

internal/core/name_mapping.go
```diff
@@ -54,13 +54,13 @@ func BuildResourceNameMapping(cluster *Cluster, resources ResourcesByNameType) R
...
 // BuildRateNameMapping constructs a new RateNameMapping instance.
-func BuildRateNameMapping(cluster *Cluster, rates RatesByNameType) RateNameMapping {
+func BuildRateNameMapping(cluster *Cluster, sis *ServiceInfoSnapshot) RateNameMapping {
 	nm := RateNameMapping{
 		cluster:     cluster,
 		fromAPIToDB: make(map[RateRef]dbRateRef),
 		fromDBToAPI: make(map[dbRateRef]RateRef),
 	}
-	for dbServiceType, rateByName := range rates {
+	for dbServiceType, rateByName := range sis.GetRates() {
```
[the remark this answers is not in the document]
wagnerd3: This is on purpose. The first clone is to make it easier to filter down the data.

internal/core/service_info_cache.go
```diff
@@ -23,34 +23,496 @@ import (
...
+}
+
+// Filter applies the filter to the ServiceInfoSnapshot and produces an
+// eagerly filtered FilteredServiceInfoSnapshot.
+func (s *ServiceInfoSnapshot) Filter(filter ServiceInfoFilter) *FilteredServiceInfoSnapshot {
+	f := &FilteredServiceInfoSnapshot{
+		snapshot: s, // gets cloned by AddToFilter
+	}
+	return f.AddToFilter(filter)
+}
```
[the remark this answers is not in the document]
wagnerd3: added tests

internal/core/service_info_cache.go
```diff
@@ -23,34 +23,496 @@ import (
...
+// GetFilter returns the current filter.
+func (f *FilteredServiceInfoSnapshot) GetFilter() ServiceInfoFilter {
+	return f.filter
+}
+
+// AddToFilter merges the given filter to the existing filter.
+// The filter can only become more specific.
+// If a property was already filtered before, the additional filter is
+// omitted.
+func (f *FilteredServiceInfoSnapshot) AddToFilter(filter ServiceInfoFilter) *FilteredServiceInfoSnapshot {
```
majewsky: This is only called within this file and thus can be private. Also, the return value does not appear to be needed:

```suggestion
func (f *FilteredServiceInfoSnapshot) addToFilter(filter ServiceInfoFilter) {
```

Following from that, I would like for all other methods on this type to take a non-pointer receiver, since they do not modify the snapshot. And then I would like for the snapshot type to be passed by-value everywhere else, since we don't have a need to pass by-pointer.

majewsky (changes requested):
First small batch of comments:

internal/datamodel/project_resource_update.go
```diff
@@ -48,9 +52,9 @@ func (u ProjectResourceUpdate) Run(dbi db.Interface, project db.Project, srv db.
 
 	// collect existing project_resources for this service
 	var projectResources []db.ProjectResource
-	_, err := dbi.Select(&projectResources, `SELECT pr.* FROM project_resources pr JOIN resources r ON pr.resource_id = r.id WHERE r.service_id = $1 AND pr.project_id = $2`, srv.ID, project.ID)
+	_, err := dbi.Select(&projectResources, `SELECT pr.* FROM project_resources pr JOIN resources r ON pr.resource_id = r.id WHERE r.service_id = $1 AND pr.project_id = $2`, service.ID, project.ID)
```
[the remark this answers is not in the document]
wagnerd3: This was only true in an old go version, since 1.22 each iteration should get a fresh version.

internal/core/service_info_cache.go
wagnerd3: I quote from the design codi: 

> looking at `limes_report_specificity`, there does not seem to be demand for anything except filtering for a single service and for a single resource; if there is potential to simplify the implementation, we could consider limiting filters like that

Which direction do we go now?

internal/core/service_info_cache.go
wagnerd3: I see a lot of usages, did you commit something and forgot to push?

For sure, we can change all access to `Path`s. I will wait for your feedback, so we don't do double work if you already did something.

internal/core/service_info_cache.go
wagnerd3: I wanted to offer the possibility to filter on a `ServiceInfoSnapshot` step by step, but you are right - it's unused. And I also don't think that we will get a usecase for that with `v2`. This means I can move the content of this function to `ServiceInfoSnapshot` and implement it directly in `Filter()`.

The pointer receivers I will remove entirely.

internal/core/service_info_cache.go
majewsky: Genuinely thank you for quoting my own design docs back at me. :) Past Stefan has a point.

internal/core/service_info_cache.go
majewsky: Huh, I ran the same grep again, and today it finds a ton of stuff. Nevermind then.

internal/core/liquid.go
```diff
@@ -298,7 +303,9 @@ func prometheusScrapeOneResource(p PrometheusCapacityConfiguration, ctx context.
 // BuildServiceCapacityRequest generates the request body payload for querying the LIQUID API
 // endpoint /v1/report-capacity. In order to be reusable for exposing an API which prints the
 // request for admin purposes, it does not use the LiquidConnection as receiver type.
-func BuildServiceCapacityRequest(backchannel CapacityScrapeBackchannel, allAZs []limes.AvailabilityZone, serviceType db.ServiceType, resources ResourcesByName) (liquid.ServiceCapacityRequest, error) {
+func BuildServiceCapacityRequest(backchannel CapacityScrapeBackchannel, allAZs []limes.AvailabilityZone, filteredSIS FilteredServiceInfoSnapshot) (liquid.ServiceCapacityRequest, error) {
```
majewsky: The fact that this function takes a FilteredServiceInfoSnapshot that specifically must have a service type filter feels weird, and not documented in the function. I can see that this was posed as a question in the PR description, and I would indeed lean in the direction of continuing to pass the service type explicitly here.
```suggestion
func BuildServiceCapacityRequest(backchannel CapacityScrapeBackchannel, allAZs []limes.AvailabilityZone, sis core.ServiceInfoReader, serviceType db.ServiceType) (liquid.ServiceCapacityRequest, error) {
```
Same on all other helper functions that take a FilteredServiceInfoSnapshot argument: parseAndValidateCommitmentRequest, enrichUsageReportTotals, NewTransferableCommitmentCache and ProjectResourceUpdate.Run.

internal/api/commitment.go
```diff
@@ -1097,12 +1099,13 @@ func (p *v1Provider) DeleteProjectCommitment(w http.ResponseWriter, r *http.Requ
 		return
 	}
 
-	service, sExists := p.Cluster.SIC.GetServiceForType(path.ServiceType)
-	resources, _ := p.Cluster.SIC.GetResourcesForType(path.ServiceType)
-	if !sExists {
+	filteredSIS := p.Cluster.SIC.GetSnapshot().Filter(core.ServiceInfoFilter{ServiceType: Some(path.ServiceType)})
+	_, rExists := filteredSIS.GetResourceForPath(path.Resource())
+	if !rExists { // checking the deepest level is enough
 		http.Error(w, "service not found", http.StatusNotFound)
```
majewsky: This error message does not make sense anymore with the updated check (unless I'm missing the broader context that includes an earlier check somewhere else):
```suggestion
		http.Error(w, "service or resource not found", http.StatusNotFound)
```

internal/api/commitment.go
```diff
@@ -1844,11 +1848,12 @@ func (p *v1Provider) ConvertCommitment(w http.ResponseWriter, r *http.Request) {
...
-	resources := p.Cluster.SIC.GetResources()
-	sourceService, sExists := p.Cluster.SIC.GetServiceForType(sourcePath.ServiceType)
-	if !sExists {
-		msg := "no such service and/or resource: " + sourcePath.Resource().String()
-		http.Error(w, msg, http.StatusUnprocessableEntity)
+
+	sis := p.Cluster.SIC.GetSnapshot()
+	filteredSIS := sis.Filter(core.ServiceInfoFilter{ServiceType: Some(sourcePath.ServiceType)})
+	sourceService, rExists := filteredSIS.GetServiceForType(sourcePath.ServiceType)
+	if !rExists { // checking the deepest level is enough
```
majewsky:
```suggestion
	sourceService, sExists := filteredSIS.GetServiceForType(sourcePath.ServiceType)
	if !sExists { // checking the deepest level is enough
```

Also, not sure what the comment is about (this also applies to other instances of this same comment on a service-level check).

majewsky (changes requested):
I pushed a few commits with boring refactoring; please take a look. Also, some more small notes:

internal/api/commitment.go
wagnerd3: The message is historically grown, it's what the `/commitment` API returns since the beginning and I did not dare to change it during all the past refactorings. I can include this change here.

internal/api/commitment.go
wagnerd3: Yes, that's a c&p error. Will remove the comment for service checks.

majewsky (approved):

PR #899, opened 2026-06-11
v2 report types

wagnerd3:
I figured to make development easier and break down the PRs, we should have a discussion about the rough typing first, so that we can spot any major problems before I am too deep in the implementation.

Some specific questions of mine are:
* Do we want to nest `opts` like I did? It's easier to read for the API consumers if we copy-paste the params around, but it's much longer...
* For the report structure, I currently have it so that `cluster` has no level for the cluster itself anymore (because Limes dropped support for multiple clusters) which `domains` and `projects` reports can have multiple. What do you think about the nesting? I was unsure whether we want the projects/domains to be a map, also (made them an array right now). In the array setup, I have the `domain_info` for projects as named nested struct. Please give feedback on that.
* I was unsure about the placement of quota properties depending on the `topology` on `ProjectResourceReport` and its childs. Please check whether this is fine. 
* Do we mention about the special `AvailabilityZoneAny, AvailabilityZoneUnknown` in the respective cases when they can appear?
* I saw in the old world, we don't summed rate reports on domain + cluster level, most likely because we store the values as string. We would have to do the summing on application layer. I think it would be good to have the endpoints available, but I can also drop them?

### DIFF
--- a/internal/apideclarations/apiv2/common/objects.go
+++ b/internal/apideclarations/apiv2/common/objects.go
@@ -1,0 +1,24 @@
+// SPDX-FileCopyrightText: 2026 SAP SE or an SAP affiliate company
+// SPDX-License-Identifier: Apache-2.0
+
+package common
+
+// DomainMetadata contains the metadata for a domain.
+// It appears in types [resourcesv2.DomainReport] and [ratesv2.DomainReport].
+type DomainMetadata struct {
+	// UUID is what OpenStack commonly refers to as domain_id.
+	UUID string `json:"uuid"`
+	Name string `json:"name"`
+}
+
+// ProjectMetadata contains the metadata for a project.
+// It appears in types [resourcesv2.ProjectReport] and [ratesv2.ProjectReport].
+type ProjectMetadata struct {
+	// UUID is what OpenStack commonly refers to as project_id.
+	UUID string `json:"uuid"`
+	Name string `json:"name"`
+	// The ParentUUID is the ID of the domain in case the project is not a subproject,
+	// else it is the UUID of another project. This hierarchy can have multiple levels.
+	ParentUUID string         `json:"parent_uuid"`
+	DomainInfo DomainMetadata `json:"domain"`
+}
--- a/internal/apideclarations/apiv2/common/opts.go
+++ b/internal/apideclarations/apiv2/common/opts.go
@@ -1,0 +1,100 @@
+// SPDX-FileCopyrightText: 2026 SAP SE or an SAP affiliate company
+// SPDX-License-Identifier: Apache-2.0
+
+package common
+
+import (
+	"time"
+
+	"github.com/sapcc/go-api-declarations/liquid"
+	. "go.xyrillian.de/gg/option"
+
+	"github.com/sapcc/limes/internal/db"
+)
+
+// GenericReportOpts contains query parameter options shared between
+// resources and rates reports.
+// It appears in types ResourceReportOpts and RateReportOpts.
+type GenericReportOpts struct {
+	// Area is a grouping, used to filter for multiple services
+	Area Option[string] `q:"area"`
+	// ServiceType filters services by type
+	ServiceType Option[db.ServiceType] `q:"service"`
+	// Category is a grouping, used to filter for multiple resources or rates
+	Category Option[liquid.CategoryName] `q:"category"`
+	// enrich response with InfoReport
+	WithInfo bool `q:"with,value:info"`
+}
+
+// ResourceReportOpts contains query parameter options for resource reports.
+// It appears in types ClusterResourceReportOpts, DomainResourceReportOpts and ProjectResourceReportOpts.
+type ResourceReportOpts struct {
+	GenericReportOpts
+	// ResourceName filters resources by name
+	ResourceName Option[liquid.ResourceName] `q:"resource"`
+	// WithUserSpecifiedConstraints enriches the response with MaxQuota and ForbidAutogrowth values
+	WithUserSpecifiedConstraints bool `q:"with,value:constraints"`
+	// WithCommitmentStats enriches the response with Committed values
+	WithCommitmentStats bool `q:"with,value:commitment_stats"`
+	// MinScrapedAt and MaxScrapedAt allow to filter services by their latest successful scrape time
+	MinScrapedAt Option[time.Time] `q:"min_scraped_at"`
+	MaxScrapedAt Option[time.Time] `q:"max_scraped_at"`
+}
+
+// ClusterResourceReportOpts contains query parameter options for cluster
+// resource reports.
+type ClusterResourceReportOpts struct {
+	ResourceReportOpts
+	// WithTiming enriches the response with ScrapedAt values
+	WithTiming bool `q:"with,value:timing"`
+	// WithSubcapacities enriches the response with Subcapacities values which is only allowed for users with certain permissions
+	WithSubcapacities bool `q:"with,value:subcapacities"`
+}
+
+// DomainResourceReportOpts contains query parameter options for domain
+// resource reports.
+type DomainResourceReportOpts struct {
+	ResourceReportOpts
+}
+
+// ProjectResourceReportOpts contains query parameter options for project
+// resource reports.
+type ProjectResourceReportOpts struct {
+	ResourceReportOpts
+	// WithTiming enriches the response with ScrapedAt values which is only allowed for users with certain permissions
+	WithTiming bool `q:"with,value:timing"`
+	// WithSubresources enriches the response with Subresources values which is only allowed for users with certain permissions
+	WithSubresources bool `q:"with,value:subresources"`
+	// DomainUUID is a special entity filter which is only allowed for users with certain permissions
+	DomainUUID Option[string] `q:"domain_uuid"`
+}
+
+// RateReportOpts contains query parameter options for rate reports.
+// It appears in type ClusterRateReportOpts, DomainRateReportOpts and ProjectRateReportOpts.
+type RateReportOpts struct {
+	GenericReportOpts
+	// RateName filters rates by name
+	RateName Option[liquid.RateName] `q:"rate"`
+}
+
+// ClusterRateReportOpts contains query parameter options for cluster
+// rate reports.
+type ClusterRateReportOpts struct {
+	RateReportOpts
+}
+
+// DomainRateReportOpts contains query parameter options for domain
+// rate reports.
+type DomainRateReportOpts struct {
+	RateReportOpts
+}
+
+// ProjectRateReportOpts contains query parameter options for project
+// rate reports.
+type ProjectRateReportOpts struct {
+	RateReportOpts
+	// WithTiming enriches the response with ScrapedAt values which is only allowed for users with certain permissions
+	WithTiming bool `q:"with,value:timing"`
+	// DomainUUID is a special entity filter which is only allowed for users with certain permissions
+	DomainUUID Option[string] `q:"domain_uuid"`
+}
--- a/internal/apideclarations/apiv2/rates/cluster.go
+++ b/internal/apideclarations/apiv2/rates/cluster.go
@@ -1,0 +1,54 @@
+// SPDX-FileCopyrightText: 2026 SAP SE or an SAP affiliate company
+// SPDX-License-Identifier: Apache-2.0
+
+package ratesv2
+
+import (
+	"github.com/sapcc/go-api-declarations/liquid"
+	. "go.xyrillian.de/gg/option"
+
+	"github.com/sapcc/limes/internal/db"
+)
+
+// ClusterGetResponse is the response type for GET /rates/v2/cluster.
+// It contains the latest state of all dynamic rate data for the whole cluster.
+// This can only contain summed data from project level.
+type ClusterGetResponse struct {
+	// InfoReport is what GET /rates/v2/info would report.
+	// It is only returned when the respective query option with=info is set.
+	InfoReport    Option[InfoReport] `json:"info,omitzero"`
+	ClusterReport ClusterReport      `json:"cluster_report"`
+}
+
+// ClusterReport groups services of the whole cluster into areas,
+// which are defined in the config.
+// It appears in [ClusterGetResponse].
+type ClusterReport struct {
+	Areas map[string]ClusterAreaReport `json:"service_areas"`
+}
+
+// ClusterAreaReport contains data for one area.
+// It appears in [ClusterReport].
+type ClusterAreaReport struct {
+	Services map[db.ServiceType]ClusterServiceReport `json:"services"`
+}
+
+// ClusterServiceReport contains data for one service.
+// It appears in [ClusterAreaReport].
+type ClusterServiceReport struct {
+	Categories map[liquid.CategoryName]ClusterCategoryReport `json:"categories"`
+}
+
+// ClusterCategoryReport groups rates into categories, which are defined in the config.
+// It appears in [ClusterServiceReport].
+type ClusterCategoryReport struct {
+	Rates map[liquid.RateName]ClusterRateReport `json:"rates"`
+}
+
+// ClusterRateReport contains data for one rate.
+// It appears in [ClusterCategoryReport].
+type ClusterRateReport struct {
+	// UsageAsBigint is a big.Int that has to be serialized as string.
+	// If the rate does not report usage, the rate is omitted from the ClusterCategoryReport.Rates map.
+	UsageAsBigint string `json:"usage_as_bigint"`
+}
--- a/internal/apideclarations/apiv2/rates/domain.go
+++ b/internal/apideclarations/apiv2/rates/domain.go
@@ -1,0 +1,57 @@
+// SPDX-FileCopyrightText: 2026 SAP SE or an SAP affiliate company
+// SPDX-License-Identifier: Apache-2.0
+
+package ratesv2
+
+import (
+	"github.com/sapcc/go-api-declarations/liquid"
+
+	"github.com/sapcc/limes/internal/apideclarations/apiv2/common"
+	"github.com/sapcc/limes/internal/db"
+
+	. "go.xyrillian.de/gg/option"
+)
+
+// DomainGetResponse is the response type for GET /rates/v2/domains(/:domain_uuid)?.
+// It contains the latest state of all dynamic rate data for one or many domains.
+// This can only contain summed data from project level.
+type DomainGetResponse struct {
+	// InfoReport is what GET /rates/v2/info would report.
+	// It is only returned when the respective query option with=info is set.
+	InfoReport    Option[InfoReport]      `json:"info,omitzero"`
+	DomainReports map[string]DomainReport `json:"domains"`
+}
+
+// DomainReport groups all services for one domain into areas,
+// which are defined in the config.
+// It appears in [DomainGetResponse].
+type DomainReport struct {
+	common.DomainMetadata
+	Areas map[string]DomainAreaReport `json:"service_areas"`
+}
+
+// DomainAreaReport contains data for one area.
+// It appears in [DomainReport].
+type DomainAreaReport struct {
+	Services map[db.ServiceType]DomainServiceReport `json:"services"`
+}
+
+// DomainServiceReport contains data for one service.
+// It appears in [DomainAreaReport].
+type DomainServiceReport struct {
+	Categories map[liquid.CategoryName]DomainCategoryReport `json:"categories"`
+}
+
+// DomainCategoryReport groups rates into categories, which are defined in the config.
+// It appears in [DomainServiceReport].
+type DomainCategoryReport struct {
+	Rates map[liquid.RateName]DomainRateReport `json:"rates"`
+}
+
+// DomainRateReport contains data for one rate.
+// It appears in [DomainCategoryReport].
+type DomainRateReport struct {
+	// UsageAsBigint is a big.Int that has to be serialized as string.
+	// If the rate does not report usage, the rate is omitted from the DomainCategoryReport.Rates map.
+	UsageAsBigint string `json:"usage_as_bigint"`
+}
--- a/internal/apideclarations/apiv2/rates/project.go
+++ b/internal/apideclarations/apiv2/rates/project.go
@@ -1,0 +1,74 @@
+// SPDX-FileCopyrightText: 2026 SAP SE or an SAP affiliate company
+// SPDX-License-Identifier: Apache-2.0
+
+package ratesv2
+
+import (
+	"github.com/sapcc/go-api-declarations/limes"
+	limesrates "github.com/sapcc/go-api-declarations/limes/rates"
+	"github.com/sapcc/go-api-declarations/liquid"
+
+	"github.com/sapcc/limes/internal/apideclarations/apiv2/common"
+	"github.com/sapcc/limes/internal/db"
+
+	. "go.xyrillian.de/gg/option"
+)
+
+// ProjectGetResponse is the response type for GET /rates/v2/projects(/:project_uuid)?.
+// It contains the latest state of all dynamic rate data for one or many projects.
+// The data comes directly from the project level.
+type ProjectGetResponse struct {
+	// InfoReport is what GET /rates/v2/info would report.
+	// It is only returned when the respective query option with=info is set.
+	InfoReport    Option[InfoReport]                `json:"info,omitzero"`
+	DomainReports map[string]ProjectsByDomainReport `json:"domains"`
+}
+
+// ProjectsByDomainReport groups all projects for one domain.
+// It appears in [ProjectGetResponse].
+type ProjectsByDomainReport struct {
+	ProjectReports map[string]ProjectReport `json:"projects"`
+}
+
+// ProjectReport groups all services for one project into areas,
+// which are defined in the config.
+// It appears in [ProjectGetResponse].
+type ProjectReport struct {
+	common.ProjectMetadata
+	Areas map[string]ProjectAreaReport `json:"service_areas"`
+}
+
+// ProjectAreaReport contains data for one area.
+// It appears in [ProjectReport].
+type ProjectAreaReport struct {
+	Services map[db.ServiceType]ProjectServiceReport `json:"services"`
+}
+
+// ProjectServiceReport contains data for one service.
+// It appears in [ProjectAreaReport].
+type ProjectServiceReport struct {
+	// ScrapedAt is the most recent time at which usage data for this service within this
+	// project was successfully collected from the backend, or None if no usage data has been collected yet.
+	// It is only returned when the respective query option with=timing is set.
+	ScrapedAt  Option[limes.UnixEncodedTime]                 `json:"scraped_at,omitzero"`
+	Categories map[liquid.CategoryName]ProjectCategoryReport `json:"categories"`
+}
+
+// ProjectCategoryReport groups rates into categories, which are defined in the config.
+// It appears in [ProjectServiceReport].
+type ProjectCategoryReport struct {
+	Rates map[liquid.RateName]ProjectRateReport `json:"rates"`
+}
+
+// ProjectRateReport contains data for one rate.
+// It appears in [ProjectCategoryReport].
+type ProjectRateReport struct {
+	// UsageAsBigint is a big.Int that has to be serialized as string.
+	// If the rate does not report usage and has no ProjectLimit, the rate is omitted from the ProjectCategoryReport.Rates map.
+	// If the rate does not report usage and has a ProjectLimit, this will be None.
+	UsageAsBigint Option[string] `json:"usage_as_bigint,omitzero"`
+	// ProjectLimit and ProjectWindow can be set by a project admin.
+	// If the rate does not have a local project rate limit, this will be None.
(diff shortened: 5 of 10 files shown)


internal/apideclarations/apiv2/resources/project.go
```diff
@@ -0,0 +1,114 @@
...
+	InfoReport     Option[InfoReport] `json:"service_info,omitzero"`
+	ProjectReports []ProjectReport    `json:"project_reports"`
+}
+
+// ProjectReport groups all services for one project into areas,
+// which are defined in the config.
+// It appears in [ProjectGetResponse].
+type ProjectReport struct {
+	apiv2.ProjectInfo
+	DomainInfo apiv2.DomainInfo             `json:"domain_info"`
```
majewsky: `DomainInfo` should be within `apiv2.ProjectInfo`, since a ProjectInfo does not occur without a DomainInfo accompanying it (and indeed, it would be somewhat incomplete without it).

internal/apideclarations/apiv2/resources/project.go
```diff
@@ -0,0 +1,114 @@
...
+	// BackendQuota is the quota value which is currently set in the service as reported last by the backend.
+	// It will appear on ProjectResourceReport, when the topology of the service does not allow for az-separated quotas.
+	// It is only returned when the resource has quota.
+	BackendQuota Option[uint64] `json:"backend_quota,omitzero"`
+	// Committed and the following values are the sum of committed amounts across projects
+	// grouped by their CommitmentDuration.
+	Committed           map[limesresources.CommitmentDuration]uint64 `json:"committed,omitempty"`
+	CommittedUnutilized map[limesresources.CommitmentDuration]uint64 `json:"committed_unutilized,omitempty"`
+	CommittedPending    map[limesresources.CommitmentDuration]uint64 `json:"committed_pending,omitempty"`
+	CommittedPlanned    map[limesresources.CommitmentDuration]uint64 `json:"committed_planned,omitempty"`
```
majewsky: This is already four separate keys, going on five when we add `guaranteed`. What do you think about simplifying this declaration like so?
```suggestion
	Committed           map[liquid.CommitmentStatus]map[limesresources.CommitmentDuration]uint64 `json:"committed,omitempty"`
	ConfirmedCommittedUnutilized map[limesresources.CommitmentDuration]uint64 `json:"confirmed_committed_unutilized,omitempty"`
```

internal/apideclarations/apiv2/resources/project.go
```diff
@@ -0,0 +1,114 @@
...
+	// It is only returned when the resource has quota.
+	Quota Option[uint64] `json:"quota,omitzero"`
+	// BackendQuota is the quota value which is currently set in the service as reported last by the backend.
+	// It will appear on ProjectResourceReport, when the topology of the service does not allow for az-separated quotas.
+	// It is only returned when the resource has quota.
+	BackendQuota Option[uint64] `json:"backend_quota,omitzero"`
+	// Committed and the following values are the sum of committed amounts across projects
+	// grouped by their CommitmentDuration.
+	Committed           map[limesresources.CommitmentDuration]uint64 `json:"committed,omitempty"`
+	CommittedUnutilized map[limesresources.CommitmentDuration]uint64 `json:"committed_unutilized,omitempty"`
```
majewsky: Can this actually be split by duration? That would require Limes to start attributing specific pieces of usage to specific commitments, which currently only happens downstream in billing.
```suggestion
	CommittedUnutilized uint64 `json:"committed_unutilized,omitempty"`
```

majewsky: Going further than the previous comment, I don't think we want to report this at all on project level, since it is redundant and can readily be computed from `Usage - sum_values(Committed)`. It's only a required report on domain and cluster level, where the report itself does not allow the user to make this computation without looking at all project reports.

internal/apideclarations/apiv2/resources/project.go
majewsky: I think I remember now we had these fields as `map[string]uint64` in v1. Can you please check if `map[limesresources.CommitmentDuration]uint64` serializes in the intended way? I have a vague suspicion that `MarshalJSON/UnmarshalJSON` is not used for map keys.

...

Actually, I checked the code really quick, and it turns out I'm right. `json.Marshaler/Unmarshaler` are not used for map keys. **However,** `encoding.TextMarshaler/TextUnmarshaler` are indeed used. So we can use them as map keys if we add implementations for those interfaces.

internal/apideclarations/apiv2/opts.go
```diff
@@ -0,0 +1,99 @@
...
+// ResourceReportOpts contains query parameter options for resource reports.
+// It appears in types ClusterResourceReportOpts, DomainResourceReportOpts and ProjectResourceReportOpts.
+type ResourceReportOpts struct {
+	GenericReportOpts
+	// ResourceName filters resources by name
+	ResourceName Option[liquid.ResourceName] `q:"resource"`
+	// WithConstraints enriches the response with MaxQuota and ForbidAutogrowth values
+	WithConstraints bool `q:"with,value:constraints"`
+	// WithCommitmentDetails enriches the response with Committed values
+	WithCommitmentDetails bool `q:"with,value:commitment_details"`
```
majewsky: Name bikeshedding: WithCommitmentStats.

internal/apideclarations/apiv2/opts.go
```diff
@@ -0,0 +1,99 @@
...
+}
+
+// ResourceReportOpts contains query parameter options for resource reports.
+// It appears in types ClusterResourceReportOpts, DomainResourceReportOpts and ProjectResourceReportOpts.
+type ResourceReportOpts struct {
+	GenericReportOpts
+	// ResourceName filters resources by name
+	ResourceName Option[liquid.ResourceName] `q:"resource"`
+	// WithConstraints enriches the response with MaxQuota and ForbidAutogrowth values
+	WithConstraints bool `q:"with,value:constraints"`
```
majewsky: Name bikeshedding: WithUserSpecifiedConstraints.

internal/apideclarations/apiv2/opts.go
```diff
@@ -0,0 +1,99 @@
...
+type DomainResourceReportOpts struct {
+	ResourceReportOpts
+}
+
+// ProjectResourceReportOpts contains query parameter options for project
+// resource reports.
+type ProjectResourceReportOpts struct {
+	ResourceReportOpts
+	// WithTiming enriches the response with ScrapedAt values
+	WithTiming bool `q:"with,value:timing"`
```
majewsky: The `only allowed for users with certain permissions` note that some filters have shoud also be on various flags:

- WithTiming
- WithSubresources
- WithSubcapacities

majewsky (commented):
> For the report structure, I currently have it so that cluster has no level for the cluster itself anymore 

Looks good to me.

> I was unsure whether we want the projects/domains to be a map, also (made them an array right now).

We can have this be a map as well.

> I think it would be good to have the endpoints available, but I can also drop them?

We can drop the domain-level endpoint for now, and only show usage totals on cluster level for the time being. Summing rate limits across projects is a bit iffy anyway, so I would like to get away with not doing it.

> Do we mention about the special AvailabilityZoneAny, AvailabilityZoneUnknown in the respective cases when they can appear?

Yes, we should have documentation about that. For now, a TODO will suffice. At some point, I will put my technical-writer hat on and go through the entire API documentation to make it nice.

internal/apideclarations/apiv2/resources/project.go
```diff
@@ -0,0 +1,114 @@
...
+}
+
+// ProjectResourceReport contains data for one resource.
+// It appears in [ProjectCategoryReport].
+type ProjectResourceReport struct {
+	AvailabilityZones map[limes.AvailabilityZone]ProjectAvailabilityZoneReport `json:"availability_zones"`
+	// BackendQuota is the quota value which is currently set in the service as reported last by the backend.
+	// It will appear on ProjectAvailabilityZoneReport, when the topology of the service allows for az-separated quotas.
+	// It is only returned when the resource has quota.
+	BackendQuota Option[uint64] `json:"backend_quota,omitzero"`
```
majewsky: We are not using this in the current version of the UI at all, so I would lean towards not surfacing it at all and seeing if someone screams. (The v2 data metrics do not have backend quota in them, either.)

majewsky (commented):
Regarding the quota structuring question:

wagnerd3:
Regarding the general comments:

> We can have this be a map as well.

☑️ → please check the json keys (I have `domain_reports` → `project_reports` now)

> We can drop the domain-level endpoint for now, and only show usage totals on cluster level for the time being. Summing rate limits across projects is a bit iffy anyway, so I would like to get away with not doing it.

Mhh, my idea was anyways to not sum the rate limits and only offer the usage. I would tend to keep the domain level report then. If I have to solve the usage-summing for cluster, we might offer it on domain for consistency. I imagine when rates would get more usage for e.g. network (and they can only be charged by usage, probably?), this requirement would surface very fast

> Yes, we should have documentation about that. For now, a TODO will suffice. At some point, I will put my technical-writer hat on and go through the entire API documentation to make it nice.

☑️

internal/apideclarations/apiv2/resources/project.go
wagnerd3: * I don't know for which use cases we need all the values. I know the billing folks need `Committed` to be the confirmed amount. So maybe this just needs a renaming anyways
* If we want to make this explicit, I would argue for the following naming (will include this for now):
```suggestion
	CommittedConfirmed                     map[limesresources.CommitmentDuration]uint64 `json:"committed_confirmed,omitempty"`
	CommittedConfirmedUnutilized    map[limesresources.CommitmentDuration]uint64 `json:"committed_confirmed_unutilized,omitempty"`
```
* I could also imagine doing a map of maps to simplify?

internal/apideclarations/apiv2/resources/project.go
wagnerd3: In billing, the huge complexity of the logic lies in the task to attribute each usage with "covered by CR" or "covered by PAYG". If you look at it from the sum of commitments it's fairly simple: CRs get used up by duration descending. So it is possible to implement this map with a few lines in the app layer.

I just came from the POV on the API signature. I will include the simple form for now.

wagnerd3: I think you meant both the values `CommittedConfirmedUnutilized` and `UncommittedUsage` here - you need the one value to calculate the other.
I will remove both from project and adjust the domain/cluster level comments to make this clear.

internal/apideclarations/apiv2/resources/project.go
wagnerd3: PR here: https://github.com/sapcc/go-api-declarations/pull/97

internal/apideclarations/apiv2/opts.go
wagnerd3: I realized the respective values need to be `option.Option`. Will change this, too.

internal/apideclarations/apiv2/resources/project.go
wagnerd3: removed the fields.

internal/apideclarations/apiv2/rates/project.go
```diff
@@ -0,0 +1,71 @@
...
+
+// ProjectGetResponse is the response type for GET /rates/v2/projects(/:project_uuid)?.
+// It contains the latest state of all dynamic rate data for one or many projects.
+// The data comes directly from the project level.
+type ProjectGetResponse struct {
+	// InfoReport is what GET /rates/v2/info would report.
+	// It is only returned when the respective query option with=info is set.
+	InfoReport    Option[InfoReport]                `json:"service_info,omitzero"`
+	DomainReports map[string]ProjectsByDomainReport `json:"project_reports"`
+}
```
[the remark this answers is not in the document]
wagnerd3: I could swear I already fixed this locally, but must have screwed up one of the rebases in the process...

internal/apideclarations/apiv2/resources/project.go
```diff
@@ -0,0 +1,106 @@
...
+	// Usage is the usage in this project as reported by the service.
+	Usage uint64 `json:"usage"`
+	// Quota is the quota value which was calculated in Limes for this availability zone.
+	// It might differ to the value set in the service, when the backend of the service was not yet updated.
+	// It is only returned when the resource has quota.
+	Quota Option[uint64] `json:"quota,omitzero"`
+	// CommittedConfirmed is the sum of committed and confirmed amounts across projects grouped by their CommitmentDuration.
+	// It is only returned when the respective query option with=commitment_stats is set.
+	CommittedConfirmed map[limesresources.CommitmentDuration]uint64 `json:"committed_confirmed,omitempty"`
+	// PhysicalUsage is usage of physical hardware resources which results from the Usage.
```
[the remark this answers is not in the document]
wagnerd3: `limesresources.CommitmentDuration` implements `encoding.TextMarshaller` and `encoding.TextUnmarshaller` now which are used by the map implementation, so this comment is wrong.

internal/apideclarations/apiv2/resources/domain.go
```diff
@@ -0,0 +1,78 @@
...
+
+// DomainAvailabilityZoneReport contains the data for an availability zone.
+// It appears in [DomainResourceReport].
+type DomainAvailabilityZoneReport struct {
+	// Usage is the sum of the usages across projects in this domain as reported by the service.
+	Usage uint64 `json:"usage"`
+	// CommittedConfirmed is the sum of committed and confirmed amounts across projects grouped by their CommitmentDuration.
+	// It is only returned when the respective query option with=commitment_stats is set.
+	CommittedConfirmed map[limesresources.CommitmentDuration]uint64 `json:"committed_confirmed,omitempty"`
+	// CommittedConfirmedUnutilized is the sum of CommittedConfirmed - Usage for each project in this domain.
```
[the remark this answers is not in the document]
wagnerd3: see previous comment.

internal/apideclarations/apiv2/resources/cluster.go
```diff
@@ -0,0 +1,92 @@
...
+	RawCapacity uint64 `json:"raw_capacity"`
+	// OverallUsage is what the service reports as overall usage, including usages that cannot be attributed to
+	// OpenStack-projects (like management workload). This is only shown if the backend does report this figure.
+	OverallUsage Option[uint64] `json:"overall_usage,omitzero"`
+	// Usage is the sum of the usages across all projects as reported by the service.
+	Usage uint64 `json:"usage"`
+	// CommittedConfirmed is the sum of committed and confirmed amounts across projects grouped by their CommitmentDuration.
+	// It is only returned when the respective query option with=commitment_stats is set.
+	CommittedConfirmed map[limesresources.CommitmentDuration]uint64 `json:"committed_confirmed,omitempty"`
+	// CommittedConfirmedUnutilized is the sum of CommittedConfirmed - Usage for each project in this cluster.
```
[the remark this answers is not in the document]
wagnerd3: see previous comment.

internal/apideclarations/apiv2/resources/project.go
```diff
@@ -0,0 +1,106 @@
...
+	Subresources json.RawMessage `json:"subresources,omitempty"`
+}
+
+// ProjectHistoricalReport contains historical summed data for an availability zone.
+// It appears in [ProjectAvailabilityZoneReport].
+type ProjectHistoricalReport struct {
+	MinUsage uint64        `json:"min_usage,omitempty"`
+	MaxUsage uint64        `json:"max_usage,omitempty"`
+	Duration time.Duration `json:"duration"`
+}
```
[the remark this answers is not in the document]
wagnerd3: Should have been `limesrates.Window`, what it really is. Will change.

wagnerd3:
@SuperSandro2000 ready for re-review.

internal/apideclarations/apiv2/rates/project.go
```diff
@@ -0,0 +1,71 @@
...
+// ProjectRateReport contains data for one rate.
+// It appears in [ProjectCategoryReport].
+type ProjectRateReport struct {
+	// UsageAsBigint is a big.Int that has to be serialized as string.
+	UsageAsBigint string `json:"usage_as_bigint"`
+	// ProjectLimit and ProjectWindow can be set by a project admin.
+	// They are only returned when the respective query option with=constraints is set.
+	ProjectLimit  Option[uint64]            `json:"project_limit,omitzero"`
+	ProjectWindow Option[limesrates.Window] `json:"project_window,omitzero"`
+}
```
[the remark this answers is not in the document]
wagnerd3: Good catch. I realized on `cluster/domain` level this is the only figure, so I can omit the whole rate when there is no usage. For project, we might have the limit though.

@majewsky What do you think should be the behavior? Do we make usage an option on project level and only omit the whole object when all are `isNone()`?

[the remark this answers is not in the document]
wagnerd3: I went with that now.

internal/apideclarations/apiv2/rates/domain.go
```diff
@@ -0,0 +1,56 @@
...
+type DomainCategoryReport struct {
+	Rates map[liquid.RateName]DomainRateReport `json:"rates"`
+}
+
+// DomainRateReport contains data for one rate.
+// It appears in [DomainCategoryReport].
+type DomainRateReport struct {
+	// UsageAsBigint is a big.Int that has to be serialized as string.
+	UsageAsBigint string `json:"usage_as_bigint"`
+}
```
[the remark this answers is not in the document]
wagnerd3: I clarified the behavior for this with a comment.

internal/apideclarations/apiv2/rates/cluster.go
```diff
@@ -0,0 +1,53 @@
...
+type ClusterCategoryReport struct {
+	Rates map[liquid.RateName]ClusterRateReport `json:"rates"`
+}
+
+// ClusterRateReport contains data for one rate.
+// It appears in [ClusterCategoryReport].
+type ClusterRateReport struct {
+	// UsageAsBigint is a big.Int that has to be serialized as string.
+	UsageAsBigint string `json:"usage_as_bigint"`
+}
```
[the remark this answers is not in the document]
wagnerd3: I clarified the behavior for this with a comment.

internal/apideclarations/apiv2/resources/project.go
```diff
@@ -0,0 +1,106 @@
...
+	Subresources json.RawMessage `json:"subresources,omitempty"`
+}
+
+// ProjectHistoricalReport contains historical summed data for an availability zone.
+// It appears in [ProjectAvailabilityZoneReport].
+type ProjectHistoricalReport struct {
+	MinUsage uint64            `json:"min_usage,omitempty"`
+	MaxUsage uint64            `json:"max_usage,omitempty"`
+	Duration limesrates.Window `json:"duration"`
+}
```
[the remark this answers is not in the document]
wagnerd3: Fair point.

internal/apideclarations/apiv2/common/objects.go
```diff
@@ -0,0 +1,24 @@
...
+
+// ProjectInfo contains the metadata for a project.
+// It appears in types [resourcesv2.ProjectReport] and [ratesv2.ProjectReport].
+type ProjectInfo struct {
+	// UUID is what OpenStack commonly refers to as project_id.
+	UUID string `json:"uuid"`
+	Name string `json:"name"`
+	// The ParentUUID is the ID of the domain in case the project is not a subproject,
+	// else it is the UUID of another project. This hierarchy can have multiple levels.
+	ParentUUID string     `json:"parent_uuid"`
```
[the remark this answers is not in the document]
wagnerd3: The suggestion would actually be wrong formatting?!

internal/apideclarations/apiv2/rates/cluster.go
```diff
@@ -0,0 +1,54 @@
...
+	"github.com/sapcc/limes/internal/db"
+)
+
+// ClusterGetResponse is the response type for GET /rates/v2/cluster.
+// It contains the latest state of all dynamic rate data for the whole cluster.
+// This can only contain summed data from project level.
+type ClusterGetResponse struct {
+	// InfoReport is what GET /rates/v2/info would report.
+	// It is only returned when the respective query option with=info is set.
+	InfoReport    Option[InfoReport] `json:"service_info,omitzero"`
```
majewsky:
```suggestion
	InfoReport    Option[InfoReport] `json:"info,omitzero"`
```
Since the endpoint it matches is called `/v2/info` and not `/v2/service_info`.

internal/apideclarations/apiv2/common/objects.go
```diff
@@ -0,0 +1,24 @@
+// SPDX-FileCopyrightText: 2026 SAP SE or an SAP affiliate company
+// SPDX-License-Identifier: Apache-2.0
+
+package common
+
+// DomainInfo contains the metadata for a domain.
+// It appears in types [resourcesv2.DomainReport] and [ratesv2.DomainReport].
+type DomainInfo struct {
```
majewsky: These are only internal names, so it's not super-important. But since this is not part of `/v2/info`, I suggest using a different name for clarity:
```suggestion
// DomainMetadata contains the metadata for a domain.
// It appears in types [resourcesv2.DomainReport] and [ratesv2.DomainReport].
type DomainMetadata struct {
```
Same for `Project{Info => Metadata}`.

internal/apideclarations/apiv2/common/objects.go
```diff
@@ -0,0 +1,24 @@
...
+// ProjectInfo contains the metadata for a project.
+// It appears in types [resourcesv2.ProjectReport] and [ratesv2.ProjectReport].
+type ProjectInfo struct {
+	// UUID is what OpenStack commonly refers to as project_id.
+	UUID string `json:"uuid"`
+	Name string `json:"name"`
+	// The ParentUUID is the ID of the domain in case the project is not a subproject,
+	// else it is the UUID of another project. This hierarchy can have multiple levels.
+	ParentUUID string     `json:"parent_uuid"`
+	DomainInfo DomainInfo `json:"domain_info"`
```
majewsky: Following from the previous note, I would like to not have "info" here, also to clarify that it is not affected by `?with=info`. Just saying "domain" should be reasonably clear.
```suggestion
	DomainInfo DomainInfo `json:"domain"`
```

internal/apideclarations/apiv2/rates/project.go
```diff
@@ -0,0 +1,73 @@
...
+// ProjectAreaReport contains data for one area.
+// It appears in [ProjectReport].
+type ProjectAreaReport struct {
+	Services map[db.ServiceType]ProjectServiceReport `json:"services"`
+}
+
+// ProjectServiceReport contains data for one service.
+// It appears in [ProjectAreaReport].
+type ProjectServiceReport struct {
+	// ScrapedAt expresses when this service/project-combination was last successfully queried from the backend.
```
majewsky: Weird phrasing:
```suggestion
	// ScrapedAt is the most recent time at which usage data for this service within this project was successfully collected from the backend, or None if no usage data has been collected yet.
```

majewsky (changes requested):
A batch of comments. This might be incomplete, I forgot to submit it yesterday and do not remember where I left off.

internal/apideclarations/apiv2/resources/project.go
majewsky: "Pending" and "Planned" are displayed in the UI.
image

wagnerd3: I have included `pending` and `planned` again. I was under the impression that we don't need them because the initial suggestion did not include them.

internal/apideclarations/apiv2/rates/cluster.go
```diff
@@ -0,0 +1,54 @@
...
+// It appears in [ClusterServiceReport].
+type ClusterCategoryReport struct {
+	Rates map[liquid.RateName]ClusterRateReport `json:"rates"`
+}
+
+// ClusterRateReport contains data for one rate.
+// It appears in [ClusterCategoryReport].
+type ClusterRateReport struct {
+	// UsageAsBigint is a big.Int that has to be serialized as string.
+	// When the rate has no usage, the rate is omitted from the ClusterCategoryReport.Rates map.
```
majewsky: To clarify that "no usage" is different from "zero usage":
```suggestion
	// If the rate does not report usage, the rate is omitted from the ClusterCategoryReport.Rates map.
```

internal/apideclarations/apiv2/common/opts.go
majewsky, line 1: Please move this into `package common`.

The reason is that I want to split up `doc.go` in this folder into two separate documents, similar to how the v1 API is split in twain, with the main entrypoint only having two links to the individual specs: https://github.com/sapcc/limes/blob/master/docs/users/api-v1-specification.md

However, because docstring links are based on imports, this would create an import cycle: `apiv2 -> resourcesv2` (for docstring link to overall API spec) and `resourcesv2 -> apiv2` (for docstring link to opts types).

internal/apideclarations/apiv2/rates/domain.go
```diff
@@ -0,0 +1,57 @@
...
+)
+
+// DomainGetResponse is the response type for GET /rates/v2/domains(/:domain_uuid)?.
+// It contains the latest state of all dynamic rate data for one or many domains.
+// This can only contain summed data from project level.
+type DomainGetResponse struct {
+	// InfoReport is what GET /rates/v2/info would report.
+	// It is only returned when the respective query option with=info is set.
+	InfoReport    Option[InfoReport]      `json:"info,omitzero"`
+	DomainReports map[string]DomainReport `json:"domain_reports"`
```
majewsky: While "Report" is a word that occurs frequently in our implementation, it does not appear in API naming anywhere else, so for consistency, `domain_reports` and `project_reports` should just be `domains` and `projects`.

internal/apideclarations/apiv2/rates/project.go
```diff
@@ -0,0 +1,74 @@
...
+type ProjectCategoryReport struct {
+	Rates map[liquid.RateName]ProjectRateReport `json:"rates"`
+}
+
+// ProjectRateReport contains data for one rate.
+// It appears in [ProjectCategoryReport].
+type ProjectRateReport struct {
+	// UsageAsBigint is a big.Int that has to be serialized as string.
+	// When the rate has no usage and no ProjectLimit, the rate is omitted from the ProjectCategoryReport.Rates map.
+	// When the rate has no usage but the project has a ProjectLimit, the UsageAsBigint.IsNone() will be true.
```
majewsky: Besides the other comment on phrasing for the first half of the sentence:
```suggestion
	// When the rate has no usage but the project has a ProjectLimit, this will be None.
```

internal/apideclarations/apiv2/resources/cluster.go
```diff
@@ -0,0 +1,98 @@
...
+type ClusterAreaReport struct {
+	Services map[db.ServiceType]ClusterServiceReport `json:"services"`
+}
+
+// ClusterServiceReport contains data for one service.
+// It appears in [ClusterAreaReport].
+type ClusterServiceReport struct {
+	// ScrapedAt is the most recent time at which capacity and usage data for this service within this
+	// cluster was successfully collected from the backend, or None if no usage data has been collected yet.
+	ScrapedAt  Option[limes.UnixEncodedTime]                 `json:"scraped_at,omitzero"`
```
majewsky, line 46: "Only included for with=timing"

Also, this should also be in the respective report for rates, since rate usage data is also scraped.

internal/apideclarations/apiv2/resources/project.go
majewsky: Reopened. We checked because I was confused by the answer, and it turns out Daniel misread my suggestion.

internal/apideclarations/apiv2/opts.go
```diff
@@ -0,0 +1,102 @@
...
+}
+
+// ProjectRateReportOpts contains query parameter options for project
+// rate reports.
+type ProjectRateReportOpts struct {
+	RateReportOpts
+	// WithTiming enriches the response with ScrapedAt values which is only allowed for users with certain permissions
+	WithTiming bool `q:"with,value:timing"`
+	// WithUserSpecifiedConstraints enriches the response with local rate limit values
+	WithUserSpecifiedConstraints bool `q:"with,value:constraints"`
```
majewsky: This constraint can go. Most requests to get a project's rates will be to see the rate limits (supposing that the ratelimit-middleware ever gets deployed at scale with customizable rate limits), so it is not particularly useful to be able to skip this. It's like having `?with=quota` on a resource report.

majewsky (changes requested):

internal/apideclarations/apiv2/resources/project.go
wagnerd3: I over-read the status in the initial suggestion 🤦🏻‍♂️ changed it.

internal/apideclarations/apiv2/rates/cluster.go
wagnerd3: changed it on all occurences.

internal/apideclarations/apiv2/resources/cluster.go
wagnerd3: There is no rate cluster scrape, so I don't want to post a scraped_at here, as it would be an aggregated values. The first point is valid though.

majewsky (approved):

PR #903, opened 2026-06-21
bump gg to v1.10.1, replace go-bits/assert with gg/assert

majewsky:
The main part of the diff is replacing gobits_assert.DeepEqual() with gg_assert.Equal(), which drops the argument containing a description of the compared value. Where this argument was used to disambiguate output from individual assertions, I replaced this with t.Run()-based subtests.


wagnerd3 (approved):
LGTM. @VoigtS wanna have a look here, too?

VoigtS (approved):

PR #904, opened 2026-06-22
v2 report API scaffold and /rates implementations

wagnerd3:
Finally the start of the implementation of the v2 report APIs.
Please review commit by commit, then it should be easily digestible.

Good to know for reviewing: We decided to rename `project_id` to `project_uuid` consistently in the API, likewise for domain.
This makes some changes necessary for the `policy.json`.
I am not 100% certain about the new structuring of that yet.

### DIFF
--- a/docs/example-policy.json
+++ b/docs/example-policy.json
@@ -2,6 +2,7 @@
     "project_scope": "project_domain_id:%(domain_id)s and project_id:%(project_id)s",
     "domain_scope": "domain_id:%(domain_id)s",
     "cluster_admin": "role:cloud_resource_admin",
+    "cluster_viewer": "role:cloud_resource_viewer or rule:cluster_admin",
     "domain_editor": "rule:cluster_admin or (rule:domain_scope and role:resource_admin)",
     "domain_viewer": "rule:domain_editor or (rule:domain_scope and role:resource_viewer)",
     "project_editor": "rule:domain_editor or (rule:project_scope and role:admin)",
@@ -27,13 +28,22 @@
     "cluster:show_errors": "rule:cluster_admin",
     "cluster:edit": "rule:cluster_admin",
 
+    "v2:project:scope": "project_domain_id:%(domain_uuid)s and project_id:%(project_uuid)s",
+    "v2:domain:scope": "domain_id:%(domain_uuid)s",
+    "v2:project:role": "role:member or role:admin",
+    "v2:domain:role": "role:resource_viewer or role:resource_admin",
     "v2:meta:no_error_obfuscation": "rule:cluster_admin",
 
-    "v2:project:info": "rule:project_viewer",
-    "v2:project:commitment_create": "rule:project_editor",
+    "v2:project:info": "rule:'v2:domain:info' or rule:'v2:project:role'",
+    "v2:project:report_single": "rule:'v2:project:report_multiple' or (rule:'v2:project:scope' and rule:'v2:project:role')",
+    "v2:project:report_multiple": "rule:cluster_viewer or (rule:'v2:domain:scope' and rule:'v2:domain:role')",
+    "v2:project:commitment_create": "rule:cluster_admin or (rule:'v2:domain:scope' and role:resource_admin) or (rule:'v2:project:scope' and role:admin)",
 
-    "v2:domain:info": "rule:domain_viewer",
+    "v2:domain:info": "rule:'v2:cluster:info' or rule:'v2:domain:role'",
+    "v2:domain:report_single": "rule:'v2:domain:report_multiple' or (rule:'v2:domain:scope' and rule:'v2:domain:role')",
+    "v2:domain:report_multiple": "rule:cluster_viewer",
 
-    "v2:cluster:info": "rule:cluster_admin",
+    "v2:cluster:info": "rule:cluster_viewer",
+    "v2:cluster:report_single": "rule:cluster_viewer",
     "v2:cluster:validation": "project_name:service and project_domain_name:Default and user_name:limes-validation and user_domain_name:Default"
 }
--- a/internal/api/api_v2/core.go
+++ b/internal/api/api_v2/core.go
@@ -58,8 +58,19 @@ func (p *v2Provider) AddTo(r *mux.Router) {
 
 	tv := p.tokenValidator
 	resRouter.Methods("GET").Path("/info").HandlerFunc(handlerFunc(http.StatusOK, tv, p.handleGetResourcesInfo))
-	ratesRouter.Methods("GET").Path("/info").HandlerFunc(handlerFunc(http.StatusOK, tv, p.handleGetRatesInfo))
+	resRouter.Methods("GET").Path("/cluster").HandlerFunc(handlerFunc(http.StatusOK, tv, p.handleGetResourcesCluster))
+	resRouter.Methods("GET").Path("/domains").HandlerFunc(handlerFunc(http.StatusOK, tv, p.handleGetResourcesDomains))
+	resRouter.Methods("GET").Path("/domains/{domain_uuid}").HandlerFunc(handlerFunc(http.StatusOK, tv, p.handleGetResourcesDomain))
+	resRouter.Methods("GET").Path("/projects").HandlerFunc(handlerFunc(http.StatusOK, tv, p.handleGetResourcesProjects))
+	resRouter.Methods("GET").Path("/projects/{project_uuid}").HandlerFunc(handlerFunc(http.StatusOK, tv, p.handleGetResourcesProject))
 	resRouter.Methods("POST").Path("/commitments/new").HandlerFunc(handlerFunc(http.StatusCreated, tv, p.handlePostNewCommitment))
+
+	ratesRouter.Methods("GET").Path("/info").HandlerFunc(handlerFunc(http.StatusOK, tv, p.handleGetRatesInfo))
+	ratesRouter.Methods("GET").Path("/cluster").HandlerFunc(handlerFunc(http.StatusOK, tv, p.handleGetRatesCluster))
+	ratesRouter.Methods("GET").Path("/domains").HandlerFunc(handlerFunc(http.StatusOK, tv, p.handleGetRatesDomains))
+	ratesRouter.Methods("GET").Path("/domains/{domain_uuid}").HandlerFunc(handlerFunc(http.StatusOK, tv, p.handleGetRatesDomain))
+	ratesRouter.Methods("GET").Path("/projects").HandlerFunc(handlerFunc(http.StatusOK, tv, p.handleGetRatesProjects))
+	ratesRouter.Methods("GET").Path("/projects/{project_uuid}").HandlerFunc(handlerFunc(http.StatusOK, tv, p.handleGetRatesProject))
 }
 
 // Wrapper for request handlers that enforces a structure,
--- a/internal/api/api_v2/core.go
+++ b/internal/api/api_v2/core.go
@@ -71,6 +71,7 @@ func (p *v2Provider) AddTo(r *mux.Router) {
 func handlerFunc[T any](successCode int, tv gopherpolicy.Validator, action func(*http.Request, *gopherpolicy.Token) (T, error)) http.HandlerFunc {
 	return func(w http.ResponseWriter, r *http.Request) {
 		t := tv.CheckToken(r)
+		t.Context.Request = mux.Vars(r)
 
 		var (
 			resp T
@@ -164,8 +165,8 @@ func (p *v2Provider) checkProjectAccess(t *gopherpolicy.Token, projectUUID liqui
 	switch {
 	case err == nil:
 		t.Context.Request = map[string]string{
-			"domain_id":  domain.UUID,
-			"project_id": string(projectUUID),
+			"domain_uuid":  domain.UUID,
+			"project_uuid": string(projectUUID),
 		}
 		err = t.Enforce(policyRule)
 		if err != nil {
@@ -173,8 +174,8 @@ func (p *v2Provider) checkProjectAccess(t *gopherpolicy.Token, projectUUID liqui
 		}
 	case errors.Is(err, sql.ErrNoRows):
 		t.Context.Request = map[string]string{
-			"domain_id":  "unknown",
-			"project_id": string(projectUUID),
+			"domain_uuid":  "unknown",
+			"project_uuid": string(projectUUID),
 		}
 		err = t.Enforce(policyRule)
 		if err == nil {
--- a/internal/api/api_v2/domain.go
+++ b/internal/api/api_v2/domain.go
@@ -69,37 +69,24 @@ func (p *v2Provider) handleGetResourcesDomain(r *http.Request, token *gopherpoli
 // handleGetRatesDomains handles GET /rates/v2/domains.
 func (p *v2Provider) handleGetRatesDomains(r *http.Request, token *gopherpolicy.Token) (_ ratesv2.DomainGetResponse, err error) {
 	httpapi.IdentifyEndpoint(r, "/rates/v2/domains")
-	none := ratesv2.DomainGetResponse{}
-
-	err = token.Enforce("v2:domain:report_multiple")
-	if err != nil {
-		return none, err
-	}
-	_, err = reports_v2.NewScope(false, r, None[string](), token, p.DB)
-	if err != nil {
-		return none, err
-	}
-	options, err := opts.ParseQueryString[common.DomainRateReportOpts](r.URL.Query())
-	if err != nil {
-		return none, err
-	}
-	_, err = reports_v2.FilterFromRateOpts(p.Cluster, options.RateReportOpts)
-	if err != nil {
-		return none, err
-	}
-	return none, nil
+	return p.commonHandleGetRatesDomain(r, token, "v2:domain:report_multiple")
 }
 
 // handleGetRatesDomain handles GET /rates/v2/domains/:domain_uuid.
 func (p *v2Provider) handleGetRatesDomain(r *http.Request, token *gopherpolicy.Token) (_ ratesv2.DomainGetResponse, err error) {
 	httpapi.IdentifyEndpoint(r, "/rates/v2/domains/:domain_uuid")
+	return p.commonHandleGetRatesDomain(r, token, "v2:domain:report_single")
+}
+
+// commonHandleGetRatesDomain handles single- and multi-domain rate calls.
+func (p *v2Provider) commonHandleGetRatesDomain(r *http.Request, token *gopherpolicy.Token, rule string) (_ ratesv2.DomainGetResponse, err error) {
 	none := ratesv2.DomainGetResponse{}
 
-	err = token.Enforce("v2:domain:report_single")
+	err = token.Enforce(rule)
 	if err != nil {
 		return none, err
 	}
-	_, err = reports_v2.NewScope(false, r, None[string](), token, p.DB)
+	scope, err := reports_v2.NewScope(false, r, None[string](), token, p.DB)
 	if err != nil {
 		return none, err
 	}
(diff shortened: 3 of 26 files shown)


docs/example-policy.json
```diff
@@ -27,13 +28,20 @@
...
+    "v2:domain_user": "role:resource_viewer or role:resource_admin",
     "v2:meta:no_error_obfuscation": "rule:cluster_admin",
 
-    "v2:project:info": "rule:project_viewer",
+    "v2:project:info": "rule:'v2:domain:info' or rule:'v2:project_user'",
+    "v2:project:report": "rule:'v2:domain:report' or (rule:'v2:project_scope' and rule:'v2:project_user')",
     "v2:project:commitment_create": "rule:project_editor",
 
-    "v2:domain:info": "rule:domain_viewer",
+    "v2:domain:info": "rule:'v2:cluster:info' or rule:'v2:project_user'",
```
majewsky: Copy-paste error?
```suggestion
    "v2:domain:info": "rule:'v2:cluster:info' or rule:'v2:domain_user'",
```

wagnerd3: yes indeed... fixed

internal/core/service_info_cache.go
```diff
@@ -28,6 +28,8 @@ import (
 // ServiceInfoFilter is a set of possible attributes by which [ServiceInfoSnapshot]
 // can be filtered. Most of the time, these come directly or indirectly from an API
 // requests' scope, e.g. query filters or attributes of a processed commitment.
+// The Category filter is special, because the category can be option.None.
+// Therefore, one can filter for empty categories with the empty string "".
```
majewsky: This description is confusing. The implementation of `isEmpty` suggests that `filter.Category = None` means "no restriction", but then what does `Some("")` do? What does "empty categories" mean?

I remember that we talked about this a while back, and I have already forgotten what the conclusion was. What I do remember is that when I put `CategoryName` as an `Option[~string]` in LIQUID, the intent was that it would be normalized with `.UnwrapOr(serviceType)` on the Limes level, so that every resource has a non-empty category name. What was the reason why we did not end up doing that? If there was one, we should document that somewhere. If there wasn't one, we should apply this normalization in SaveServiceInfoToDB and remove the `Option[...]` wrapping from all the category stuff (up to and including making `resources.category_id` a `NOT NULL` field).

Edit: I saw in the testcase that it says "filtering by `Some("")` returns all resources with `None` category". I remember something like that, and I still maintain that this is not a useful filter in practice. Compared to the alternative where `category = None` is coerced into `category.Name = service.Type`, the only additional query this makes possible is "all uncategorized resources from across all services", which is not a useful filter given that "uncategorized" can mean wildly different things across services.

internal/core/service_info_cache.go
```diff
@@ -353,45 +362,36 @@ type FilteredServiceInfoSnapshot struct {
...
+
+// GetFilteredResourceName returns the filters ResourceName property.
+func (f FilteredServiceInfoSnapshot) GetFilteredResourceName() Option[liquid.ResourceName] {
+	return f.filter.ResourceName
+}
+
+// GetFilteredRateName returns the filters RateName property.
+func (f FilteredServiceInfoSnapshot) GetFilteredRateName() Option[liquid.RateName] {
+	return f.filter.RateName
 }
```
majewsky: All these methods do not appear to be used. If we do need them, I would rather just have `GetFilter()` for simplicity.

internal/util/test.go
majewsky, line 1: This should be in `package test`. I don't want these strings baked into prod binaries.

internal/api/reports_v2/scope_test.go
```diff
@@ -0,0 +1,317 @@
...
+		assert.Equal(t, token.Context.Request["project_uuid"], expected)
+	} else {
+		assert.Equal(t, token.Context.Request["project_uuid"], "")
+	}
+
+	if expected, eOK := e.requestDomainUUID.Unpack(); eOK {
+		assert.Equal(t, token.Context.Request["domain_uuid"], expected)
+	} else {
+		assert.Equal(t, token.Context.Request["domain_uuid"], "")
+	}
```
majewsky:
```suggestion
	assert.Equal(scope.Project, e.project)
	assert.Equal(scope.Domain, e.domain)
	assert.Equal(t, token.Context.Request["project_uuid"], e.requestProjectUUID.UnwrapOr(""))
	assert.Equal(t, token.Context.Request["domain_uuid"], e.requestDomainUUID.UnwrapOr(""))
```

internal/api/reports_v2/scope_test.go
```diff
@@ -0,0 +1,317 @@
...
+	// the real validator fills the Request, so we have to emulate this here
+	maps.Copy(token.Context.Request, mux.Vars(i.r))
+	scope, err := reports_v2.NewScope(i.forProject, i.r, i.queryDomainUUID, token, i.s.DB)
+
+	if expected, ok := e.err.Unpack(); ok {
+		assert.Equal(t, err != nil, true)
+		assert.Equal(t, err.Error(), expected.Error())
+	} else {
+		assert.Equal(t, err, nil)
+	}
```
majewsky:
```suggestion
	assert.ErrEqual(t, err, e.err.UnwrapOr(nil))
```

internal/api/reports_v2/scope.go
```diff
@@ -0,0 +1,158 @@
...
+// NewScope obtains the project and domain from the database.
+// When forProject=true, this should be called before evaluating the tokens authorization
+// so that the project-->domain relation of domain admins is checked properly.
+// It returns the validated scope information or an error.
+func NewScope(forProject bool, r *http.Request, queryDomainUUID Option[string], token *gopherpolicy.Token, dbm *gorp.DbMap) (s Scope, err error) {
+	urlDomainUUID := mux.Vars(r)["domain_uuid"]
+	urlProjectUUID := mux.Vars(r)["project_uuid"]
+
+	isDomainUser := token.DomainScopeUUID() != ""
+	isProjectUser := token.ProjectScopeUUID() != "" && token.ProjectScopeName() != "cloud_admin" && token.ProjectScopeDomainUUID() != "" && token.ProjectScopeDomainName() != "ccadmin"
```
majewsky: We should not hardcode our specific cloud-admin project, when cloud-admin-ness could just as well be encoded via system-scoped tokens or any other scope (including in our own infrastructure at a later point).

I have not considered all the consequences of this particular suggestion, but the way this ought to be done is to defer to the policy rules, so I would imagine it looking something like:

```suggestion
	isDomainUser := token.Check("v2:domain:report")
	isProjectUser := !token.Check("v2:cluster:report")
```

The second variable name does not make sense anymore after this, but that's vaguely how it ought to be structured, if we can make it fit with all the other constraints.

For example, I realize that matching on the full `:report` rule might be an overreach, because it might misclassify someone as e.g. "not a domain user" if they have a domain-scoped token, but are missing a required role assignment there. Maybe we need to specify additional policy rules `v2:cluster:scope` and so on, that just say whether you could be a cloud admin (and so on) with the right roles.

From this would also follow that the mock useridentity strings in (currently) `internal/util/test.go` can be removed, because the respective mocking can be done by selecting outcomes in the MockEnforcer.

internal/api/reports_v2/scope.go
```diff
@@ -0,0 +1,158 @@
...
+
+	// a project user for project API without URL param will get an authentication error from the token check, so we just return here
+	if forProject && isProjectUser && urlProjectUUID == "" {
+		return s, nil
```
majewsky: For defense in depth, we could return a dummy project to ensure that, even if the policy accepts this request for some reason, they would not match any of the projects in the DB:
```suggestion
		return Scope{Project: Some(db.Project{ID: -1})}, nil
```

internal/api/api_v2/domain.go
```diff
@@ -0,0 +1,106 @@
...
+	none := resourcesv2.DomainGetResponse{}
+
+	err = token.Enforce("v2:domain:report")
+	if err != nil {
+		return none, err
+	}
+	_, err = reports_v2.NewScope(false, r, None[string](), token, p.DB)
+	if err != nil {
+		return none, err
+	}
```
majewsky: In the case of getting a report for a single domain, this fails because the policy (as written in `docs/example-policy.json`) relies on `token.Context.Request["domain_uuid"]` being set, but nobody sets this in time. NewScope runs too late.

internal/api/api_v2/domain.go
```diff
@@ -0,0 +1,106 @@
...
```
majewsky: In the case of listing domains as the admin of a single domain, this auth check always fails because the example policy uses `domain_id:%(domain_uuid)s`, but there is no value for the `domain_uuid` parameter.

We should consider whether we need separate policy checks for "can this user get a domain report at all" and "can this user get a domain report for $DOMAIN". Same for projects.

majewsky (dismissed):
Comments for the first commit in the chain. I have not read all of `scope.go` yet, given that my critique of the first part is pretty fundamental.

internal/core/service_info_cache.go
wagnerd3: W.r.t. the normalization part (that we have `Option[Category]` in the database), I remember we wanted to keep the logic on the reader-side. E.g. we were unsure whether we wanted to leave out the category instead of it having the `serviceType` value in some cases. I would personally stick to that serialization for now and would offer to put a separate PR later to change this everywhere to keep this PR a little smaller.

If filtering for uncategorized resources is not a thing, I will adjust the implementation now for that so that `""` is not a special case anymore.

internal/api/reports_v2/scope.go
wagnerd3: I already have `v2:domain_scope` and `v2:project_scope` in the `example-policy`, I will adjust them to fit the naming scheme you suggested.

Yes, you are right, with that I think I can get rid of the identity stuff.

internal/api/api_v2/domain.go
wagnerd3: Ah, you are right. I wanted to add the snippet from the v1 API to the handler func.
`t.Context.Request = mux.Vars(r)`

internal/api/api_v2/domain.go
wagnerd3: This is in line with the documentation, I specified
> // A project-scoped token cannot access this path.
> // A domain-scoped token can only access this path, if its domain_uuid is set in the URL parameter.
> // A cloud-admin token can receive data for requests with and without domain_uuid.

I don't know whether it's super beneficial to allow calling the endpoint for listing the one single domain. Same for project. Do we really want that?

wagnerd3:
All changes for the resolved comments are included now.

internal/core/service_info_cache.go
majewsky: Okay, makes sense. Can we please add:

- a TODO about maybe changing this normalization to take place at ingestion time, and
- a test where we filter by a category name equal to the service type, and check that we retrieve resources with `category = None`?

internal/api/api_v2/domain.go
majewsky: I suppose my confusion stems from `v2:domain:report` being both for reporting on a bunch of domains and on a single domain, where both cases have different variables set in `token.Context.Request`. And same for projects. Would it make things any clearer if we were to split this into two different policy rules, e.g. `v2:domain:report_multiple` and `v2:domain:report_single`? Same for projects.

internal/core/service_info_cache.go
wagnerd3: * TODO added
* Test was already added

internal/api/api_v2/domain.go
wagnerd3: Yes, can do that.

internal/api/api_v2/core.go
```diff
@@ -58,8 +58,19 @@ func (p *v2Provider) AddTo(r *mux.Router) {
 
 	tv := p.tokenValidator
 	resRouter.Methods("GET").Path("/info").HandlerFunc(handlerFunc(http.StatusOK, tv, p.handleGetResourcesInfo))
-	ratesRouter.Methods("GET").Path("/info").HandlerFunc(handlerFunc(http.StatusOK, tv, p.handleGetRatesInfo))
+	resRouter.Methods("GET").Path("/cluster").HandlerFunc(handlerFunc(http.StatusOK, tv, p.handleGetResourcesCluster))
+	resRouter.Methods("GET").Path("/domains").HandlerFunc(handlerFunc(http.StatusOK, tv, p.handleGetResourcesDomains))
+	resRouter.Methods("GET").Path("/domains/{domain_uuid}").HandlerFunc(handlerFunc(http.StatusOK, tv, p.handleGetResourcesDomain))
+	resRouter.Methods("GET").Path("/projects").HandlerFunc(handlerFunc(http.StatusOK, tv, p.handleGetResourcesProjects))
+	resRouter.Methods("GET").Path("/projects/{project_uuid}").HandlerFunc(handlerFunc(http.StatusOK, tv, p.handleGetResourcesProject))
```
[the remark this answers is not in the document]
wagnerd3: These routes are not accessible on production yet, so this is not relevant.

internal/api/reports_v2/project.go
```diff
@@ -0,0 +1,167 @@
...
+	services := filter.GetServices()
+	categories := filter.GetCategories()
+
+	for _, serviceType := range slices.Sorted(maps.Keys(services)) {
+		rates, _ := filter.GetRatesForType(serviceType) // can have no resources
+		for _, rateName := range slices.Sorted(maps.Keys(rates)) {
+			rate := rates[rateName]
+			if rate.ID != rateID {
+				continue
+			}
```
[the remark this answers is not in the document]
wagnerd3: The comment is right in theory, but we will not build a second data structure just for this access to be faster. If this was a problem on large reports, we could still change this.

internal/api/reports_v2/domain.go
```diff
@@ -0,0 +1,141 @@
...
+	services := filter.GetServices()
+	categories := filter.GetCategories()
+
+	for _, serviceType := range slices.Sorted(maps.Keys(services)) {
+		rates, _ := filter.GetRatesForType(serviceType) // can have no resources
+		for _, rateName := range slices.Sorted(maps.Keys(rates)) {
+			rate := rates[rateName]
+			if rate.ID != rateID {
+				continue
+			}
```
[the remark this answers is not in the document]
wagnerd3: same as above.

internal/api/reports_v2/cluster.go
```diff
@@ -0,0 +1,119 @@
...
+	services := filter.GetServices()
+	categories := filter.GetCategories()
+
+	for _, serviceType := range slices.Sorted(maps.Keys(services)) {
+		rates, _ := filter.GetRatesForType(serviceType) // can have no resources
+		for _, rateName := range slices.Sorted(maps.Keys(rates)) {
+			rate := rates[rateName]
+			if rate.ID != rateID {
+				continue
+			}
```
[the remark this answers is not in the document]
wagnerd3: same as above.

internal/api/api_v2/core.go
```diff
@@ -71,6 +82,7 @@ func (p *v2Provider) AddTo(r *mux.Router) {
 func handlerFunc[T any](successCode int, tv gopherpolicy.Validator, action func(*http.Request, *gopherpolicy.Token) (T, error)) http.HandlerFunc {
 	return func(w http.ResponseWriter, r *http.Request) {
 		t := tv.CheckToken(r)
+		t.Context.Request = mux.Vars(r)
```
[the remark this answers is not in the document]
wagnerd3: I think it's nicer to handle this on write. The map could always be nil, so it's better to check when writing.

internal/api/reports_v2/filter.go
```diff
@@ -0,0 +1,150 @@
...
+// is the name of the entity whose ID-column values are used.
+// It supports service_id, resource_id and rate_id.
+// On unknown keywords it will panic.
+func (f Filter) ExpandServiceFilters(originalQuery string, originalArgs ...any) (query string, args []any) {
+	// get current highest index
+	var err error
+	i := 0
+	queryVariables := regexp.MustCompile(`\$(\d{1,2})`)
+	matches := queryVariables.FindAllString(originalQuery, -1)
+	if len(matches) > 0 {
```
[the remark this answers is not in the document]
wagnerd3: I don't see how the `len(matches)` can be skipped, an index of -1 will panic.

internal/api/api_v2/project.go
```diff
@@ -0,0 +1,107 @@
...
+		return none, err
+	}
+	return none, nil
+}
+
+// handleGetResourcesProject handles GET /resources/v2/projects/:project_id.
+func (p *v2Provider) handleGetResourcesProject(r *http.Request, token *gopherpolicy.Token) (_ resourcesv2.ProjectGetResponse, err error) {
+	httpapi.IdentifyEndpoint(r, "/resources/v2/projects/:project_id")
+	none := resourcesv2.ProjectGetResponse{}
+
```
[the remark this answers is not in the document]
wagnerd3: I could swear I had fixed this earlier on...

internal/api/api_v2/project.go
```diff
@@ -0,0 +1,107 @@
...
+func (p *v2Provider) handleGetRatesProjects(r *http.Request, token *gopherpolicy.Token) (_ ratesv2.ProjectGetResponse, err error) {
+	httpapi.IdentifyEndpoint(r, "/rates/v2/projects")
+	return p.commonHandleGetRatesProject(r, token, "v2:project:report_multiple")
+}
+
+// handleGetRatesProject handles GET /rates/v2/projects/:project_id.
+func (p *v2Provider) handleGetRatesProject(r *http.Request, token *gopherpolicy.Token) (_ ratesv2.ProjectGetResponse, err error) {
+	httpapi.IdentifyEndpoint(r, "/rates/v2/projects/:project_id")
+	return p.commonHandleGetRatesProject(r, token, "v2:project:report_single")
+}
```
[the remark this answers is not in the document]
wagnerd3: I could swear I had fixed this earlier on...

internal/api/api_v2/project.go
```diff
@@ -0,0 +1,107 @@
...
+	none := resourcesv2.ProjectGetResponse{}
+
+	// important: validate scope before token.Enforce, so that projects domain_uuid is written back to token scope
+	options, err := opts.ParseQueryString[common.ProjectResourceReportOpts](r.URL.Query())
+	if err != nil {
+		return none, err
+	}
+	_, err = reports_v2.NewScope(true, r, options.DomainUUID, token, p.DB)
+	if err != nil {
+		return
```
VoigtS: This is inconsistent. Some NewScope calls return `none, err`, while others like this do a naked return.
F.e. `commonHandleGetRatesDomain` handles it that way.  This should probably be consistent.

internal/api/api_v2/domain.go
```diff
@@ -0,0 +1,106 @@
...
+	_, err = reports_v2.FilterFromResourceOpts(p.Cluster, options.ResourceReportOpts)
+	if err != nil {
+		return none, err
+	}
+	return none, nil
+}
+
+var none = ratesv2.DomainGetResponse{}
```
VoigtS: Why is this a package level variable?
`handleGetResourcesDomains` even defines it locally.

VoigtS (changes requested):
I only have a couple of small remarks. I think I understand the structure now. Essentially the broad cluster snapshots gets filtered down by query params and the Scope. 

As hinted in the go-bits PR I don't like the jq JSON modifications for fixture files. 
Trying to build a diff in the head with statements like:
`"del(.cluster_report.service_areas.first.services.first.categories.first)"`
is harder than having a scan over a prebuild fixture.
However, this remark should not hinder the merge.

internal/api/api_v2/domain.go
wagnerd3: I wanted to change the architecture of the functions at some point, this is a leftover. I removed that.

VoigtS (approved):

PR #905, opened 2026-06-23
config: replace CatalogURL with APIDomainNames

majewsky:
I want to set up the API such that v1 endpoints can only be queried with the v1 domain name (the one that says "Limes" in it) and v2 endpoints can only be queried with a separate domain name (which will say "Limitas" instead). Therefore, this new configuration section, unlike CatalogURL before it, is required, and the bulk of the diff is to add it to all test configs.

The actual change to inspect the Host/X-Forwarded-Host of requests will follow in a later branch. I'm only setting up infrastructure for now.

Also, I want to get rid of or at least rework the `GET /` endpoint, which is currently also used as the target for the liveness/readiness probes of API pods. To untangle this, this PR adds the conventional `GET /healthcheck` endpoint.

**Please do not merge without the respective helm-charts and secrets PR reviewed and approved.**

### DIFF
--- a/docs/operators/config.md
+++ b/docs/operators/config.md
@@ -30,6 +30,8 @@ on the top left corner of this document to get to a specific section of this gui
 
 | Variable | Default | Description |
 | --- | --- | --- |
+| `LIMES_API_DOMAIN_NAME_V1` | *(required)* | Domain name of Limes v1 API as it appears in the Keystone service catalog for this cluster. The public endpoint URL in the catalog entry for service type `resources` must be `https://THIS_VALUE`. |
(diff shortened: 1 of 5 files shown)


wagnerd3 (changes requested):
The change itself is fine and consistent, but I would like to raise the question whether it would be good to do this via mandatory env variable instead of the config?
I kind of dislike this being in the config as it adds bloat to the tests and provides no value to the application logic. So far, everything in the config has been app-logic related and not related to the server setup, which I liked.
Maybe I am missing something, why this kind of validation that we need to do is not possible with env variable?

wagnerd3 (approved):

PR #909, opened 2026-06-29
api: simplify VersionProvider, enforce configured domain names for v1/v2 API

majewsky:
The domain name enforcement ensures that the v1 API is served from the "Limes"-named domain name only, and v2 only from the "Limitas"-named domain name.

Regarding the VersionProvider simplification: The initial implementation of `GET /` and `GET /v1/`, nearly a decade ago, was based on my cargo-culting of generic recommendations for REST APIs, but the discoverability angle is significantly less important in OpenStack because APIs are discovered via the Keystone catalog. Really the only thing that we could be said to need is a link from the base URL (as entered in the Keystone catalog) to the API documentation, so we do just that: straight redirects without any JSON payload shenanigans.

### DIFF
--- a/internal/api/api_v2/core.go
+++ b/internal/api/api_v2/core.go
@@ -21,6 +21,7 @@ import (
 	"github.com/sapcc/go-bits/gopherpolicy"
 	"github.com/sapcc/go-bits/httpapi"
 	"github.com/sapcc/go-bits/respondwith"
+	. "go.xyrillian.de/gg/option"
 
 	"github.com/sapcc/limes/internal/core"
 	"github.com/sapcc/limes/internal/db"
@@ -29,6 +30,7 @@ import (
 type v2Provider struct {
 	Cluster        *core.Cluster
 	DB             *gorp.DbMap
+	DomainNames    Option[DomainNames]
 	tokenValidator gopherpolicy.Validator
 	auditor        audittools.Auditor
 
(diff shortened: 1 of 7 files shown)


internal/api/api_v2/version_provider.go
```diff
@@ -141,3 +143,29 @@ func CollectDomainNamesFromEnv() (result DomainNames, err error) {
...
+// EnforceDomainName returns a middleware that rejects requests where
+// the hostname in the request URL does not match the given value.
+func EnforceDomainName(domainName string) mux.MiddlewareFunc {
+	return func(inner http.Handler) http.Handler {
+		return http.HandlerFunc(func(w http.ResponseWriter, r *http.Request) {
+			hostName := getHostNameFor(r)
+			if hostName == domainName {
+				inner.ServeHTTP(w, r)
+			} else {
+				msg := fmt.Sprintf("endpoint %s cannot be accessed on %s", r.URL.EscapedPath(), hostName)
```
SuperSandro2000, line 99: Should we mention what domain should be used instead? I think this could avoid a ticket or two.

majewsky: I was conservative with the error message to avoid accidentally disclosing too much information to someone who maybe ought not know it.

If the support tickets materialize, we can add more details. But I really don't think it will happen, given that API endpoint URLs are usually only ever constructed from what is in the catalog.

internal/api/api_v2/core.go
```diff
@@ -39,16 +41,25 @@ type v2Provider struct {
 // NewV2API creates an httpapi.API that serves the Limes v2 API.
 // It also returns the VersionData for this API version which is needed for the
 // version advertisement on "GET /".
-func NewV2API(cluster *core.Cluster, tokenValidator gopherpolicy.Validator, auditor audittools.Auditor, timeNow func() time.Time) httpapi.API {
-	return &v2Provider{Cluster: cluster, DB: cluster.DB, tokenValidator: tokenValidator, auditor: auditor, timeNow: timeNow}
+func NewV2API(cluster *core.Cluster, domainNames Option[DomainNames], tokenValidator gopherpolicy.Validator, auditor audittools.Auditor, timeNow func() time.Time) httpapi.API {
```
wagnerd3, line 44: I would have personally just made this parameter non-`Option`, set the value statically in `setup` and just check it in the 2 tests. But I do get, that it's nice to read the param in the test setup and not in setup centrally. It's only `Option` so that it can be omitted for the tests, that's a bit counterintuitive for me.

majewsky: That's not the only intent. If I made this a non-`Option`, I would have to adjust every test because e.g. `GET /v2/info` is not `GET http://limitas.example.com/v2/info` and thus gets rejected by the middleware. The intent is to skip the host-matching middleware during most tests to avoid line noise.

wagnerd3: 💡ah okay, that makes sense - thank you!

wagnerd3 (approved):

PR #917, opened 2026-07-14
Less cross join

SuperSandro2000:
This was tested by running both sql queries, sorting the output (diff cannot handle 512M well and starts to show a diff for everything at some point) and running a diff over it. Only minor differences in values which looked very expected where observed. The actual structure and amount of rows stayed the same.


wagnerd3 (dismissed):
NICE

wagnerd3 (approved):

PR #918, opened 2026-07-17
GET/ DELETE /v2/commitments 

wagnerd3:
I will open this branch here as draft, so that the progress does not get lost or cause double work.
I had to switch back to the v2 report PR in the middle of this, so this is not finished.

What has been done:
* I cherry-picked the necessary documentation and types from the `v2-commitment-api` branch
* implemented `DELETE` API and tests
* implemented `GET` API and tests
* switch all v2 API responses to RFC3339 encoded times
* some more cleanup for the scope-related functions


docs/example-policy.json
majewsky, line 1: The list of policy rules in `docs/operators/config.md` should be extended appropriately.

internal/api/api_v2/commitment_get.go
```diff
@@ -0,0 +1,191 @@
...
+		return nil
+	})
+
+	if err != nil {
+		return none, err
+	}
+	return result, nil
+}
+
+func checkOptions(token *gopherpolicy.Token, options common.CommitmentListOpts) error {
```
majewsky: Name is too generic to be useful:
```suggestion
func checkCommitmentListOpts(token *gopherpolicy.Token, options common.CommitmentListOpts) error {
```

internal/api/api_v2/commitment_delete.go
```diff
@@ -0,0 +1,144 @@
...
+	"github.com/sapcc/limes/internal/datamodel"
+	"github.com/sapcc/limes/internal/db"
+	"github.com/sapcc/limes/internal/util"
+)
+
+var findDeletableCommitmentQuery = sqlext.SimplifyWhitespace(db.ExpandEnumPlaceholders(`
+	SELECT * FROM project_commitments
+	WHERE uuid = $1
+	AND status NOT IN ({{liquid.CommitmentStatusSuperseded}}, {{liquid.CommitmentStatusExpired}}, {{util.CommitmentStatusDeleted}})
+`))
```
majewsky: Style nitpick: This can use a PreparedSelectQuery.

```suggestion
var findDeletableCommitmentQuery = db.ProjectCommitmentStore.MustPrepareSelectQueryWhere(
	sqlext.SimplifyWhitespace(db.ExpandEnumPlaceholders(`
		`uuid = $1 AND status NOT IN ({{liquid.CommitmentStatusSuperseded}}, {{liquid.CommitmentStatusExpired}}, {{util.CommitmentStatusDeleted}})`,
	)),
))
```

internal/api/api_v2/commitment_delete.go
```diff
@@ -0,0 +1,144 @@
...
+	if !ok {
+		// defense in depth, the referenced AZResource should exist
+		return nil, errInvalidResourceReference
+	}
+
+	// check auth
+	scope, err := p.checkProjectAccessByID(ctx, token, c.ProjectID, "v2:project:commitment_delete")
+	if err != nil {
+		return nil, err
+	}
```
majewsky: The SELECT query + checkProjectAccessByID will be the same in all other non-read commitment endpoints (merge, renew, patch, etc.), so we can have this as a reusable helper function, with just a different choice of policy rule per invocation.

internal/api/api_v2/commitment_get.go
```diff
@@ -0,0 +1,191 @@
...
+func (p *v2Provider) handleGetCommitmentSingle(r *http.Request, token *gopherpolicy.Token) (resourcesv2.Commitment, error) {
+	httpapi.IdentifyEndpoint(r, "/resources/v2/commitments/:commitment_uuid")
+	var (
+		none resourcesv2.Commitment
+		ctx  = r.Context()
+		sis  = p.Cluster.SIC.GetSnapshot()
+	)
+
+	cUUID := liquid.CommitmentUUID(token.Context.Request["commitment_uuid"])
+	c, err := db.ProjectCommitmentStore.SelectOneWhere(ctx, p.DB, `uuid = $1`, cUUID)
```
majewsky, line 37: The AuthZ check for read access to superseded/expired/deleted commitments here should be consistent with the AuthZ check for `?with=deleted` on commitment list.

internal/api/api_v2/commitment_get.go
```diff
@@ -0,0 +1,191 @@
...
+var findCommitmentsQuery = sqlext.SimplifyWhitespace(db.ExpandEnumPlaceholders(`
+	SELECT p.uuid as project_uuid, pc.* FROM project_commitments pc
+	JOIN projects p ON pc.project_id = p.id
+	JOIN domains d ON p.domain_id = d.id
+	WHERE {{pc.az_resource_id = ANY($az_resource_id)}} 
+	AND {{d.id = ANY($domain_id)}}
+	AND {{p.id = ANY($project_id)}}
+	$with_public{{AND pc.transfer_status = {{limesresources.CommitmentTransferStatusPublic}}}}
+	AND {{pc.updated_at >= $updated_after}}
+	AND pc.status NOT IN ({{liquid.CommitmentStatusSuperseded}}, {{liquid.CommitmentStatusExpired}} $without_deleted{{, {{util.CommitmentStatusDeleted}}}})
```
majewsky: I can see the case for why `?with=deleted` exists, but it should also cover `superseded` and `expired`, esp. because commitment-get shows superseded/expired commitments as well.

internal/api/api_v2/commitment_get.go
```diff
@@ -0,0 +1,191 @@
...
+	// note: the domain/ project scope is checked where we assemble the scope later, not here
+	if options.OnlyPublic {
+		err := token.Enforce("v2:project:commitment_get_public")
+		if err != nil {
+			return err
+		}
+	}
+	if options.WithDeleted && !isAdmin {
+		return respondwith.CustomStatus(http.StatusForbidden, errors.New(`"with=deleted" requires special permissions`))
+	}
```
majewsky: The `commitment_get_admin` policy rule is doing at least two different things, which is confusing. The permission to use `?with=deleted` should be its own policy rule, similar to the other `...:with_foo` policy rules.

internal/api/api_v2/commitment_get.go
```diff
@@ -0,0 +1,191 @@
...
+			return none, err
+		}
+	} else if projectUUID, ok := options.ProjectUUID.Unpack(); ok {
+		scope, err = p.checkProjectAccess(ctx, token, projectUUID, "v2:project:commitment_get")
+		if err != nil {
+			return none, err
+		}
+	} else {
+		scope = reports_v2.ClusterScope{}
+	}
```
majewsky: I would like for this second paragraph to be inlined into checkOptions, which thus needs to return the chosen `scope`. I may have thoughts on how to rearrange the validation/authorization steps, and that will be easier to play with when they are next to each other in the same function body.

internal/api/api_v2/commitment_get.go
```diff
@@ -0,0 +1,191 @@
...
+	err = oblast.MustNewStore[record](oblast.PostgresDialect()).Select(ctx, p.DB, query, args...).Foreach(func(c record) error {
+		// display form
+		// obtain service ref
+		azRes, ok := sis.GetAZResourceForID(c.AZResourceID)
+		if !ok {
+			// defense in depth, the referenced AZResource should exist
+			return errInvalidResourceReference
+		}
+		canBeDeleted := datamodel.CanDeleteCommitment(token, c.ProjectCommitment, p.timeNow)
+		result.Commitments = append(result.Commitments, convertCommitmentToDisplayForm(c.ProjectCommitment, azRes.Path, c.ProjectUUID, canBeDeleted))
```
majewsky: This can leak project UUIDs to users from non-privileged scopes querying for public commitments.

I suggest that we add `DomainUUID` to `type record` via an extra join, and then `token.Check("v2:project:commitment_get")` using the project and domain UUID of the commitment. If the check comes back negative, we should remove the ProjectUUID attribute from display (i.e. give `""` and set `omitempty`).

The repeated policy evaluations might be a bit expensive, but we could cache the results of `token.Check()` calls per project UUID.

majewsky (changes requested):
This is all that I managed to find before having to leave for the day:

docs/example-policy.json
wagnerd3: added in 8ec361c2

internal/api/api_v2/commitment_get.go
wagnerd3: Implemented in 6d16ae49

internal/api/api_v2/commitment_get.go
wagnerd3: Implemented in 69910fd1

internal/api/api_v2/commitment_delete.go
wagnerd3: Implemented in 69910fd1

internal/api/api_v2/commitment_delete.go
wagnerd3: Implemented in 69910fd1

internal/api/api_v2/commitment_get.go
wagnerd3: I renamed to `with=inactive` and extended the scope as described. Implemented in 9c91d888

internal/api/api_v2/commitment_get.go
wagnerd3: I splitted this into 2 rules `v2:project:with_inactive` and `v2:project:commitment_get_unscoped`. Implemented in 9c91d888

internal/api/api_v2/commitment_get.go
wagnerd3: Implemented in 959c2a7e

internal/api/api_v2/commitment_get.go
wagnerd3: Implemented in 959c2a7e

internal/apideclarations/apiv2/common/opts.go
```diff
@@ -95,3 +97,19 @@ type ProjectRateReportOpts struct {
...
+	OnlyPublic  bool                       `q:"public"`       // list all commitments in all projects that have transfer_status = "public" (for marketplace usecase)
+	DomainUUID  Option[string]             `q:"domain_uuid"`  // if given, must be below authenticated scope; if not given, shows all commitments within authenticated scope (except if OnlyPublic = true, see there)
+	// the service/resource related filters are also considered main filters; if category or resource are given, service is mandatory
+	ServiceType  Option[db.ServiceType]      `q:"service"`
+	Category     Option[liquid.CategoryName] `q:"category"`
+	ResourceName Option[liquid.ResourceName] `q:"resource"`
+
+	// extra filters
+	UpdatedAfter Option[time.Time] `q:"updated_after,format:RFC3339"` // any change on the commitment is considered an update
+	WithInactive bool              `q:"with,value:inactive"`          // returns deleted, superseded and expired commitments; only allowed for users with certain permissions
```
majewsky: "Inactive" could be construed to also include "planned" or "pending" commitments since those are not active _yet_ (for certain definitions of "active"). I suggest `?with=obsolete` instead, which has the clearer connotation of "used to be relevant, but is not anymore".

internal/apideclarations/apiv2/common/time_test.go
```diff
@@ -0,0 +1,74 @@
...
+	buf, err := json.Marshal(u)
+	if err != nil {
+		t.Fatal(err.Error())
+	}
+
+	actual := string(buf)
+	expected := `"2024-03-15T12:30:45Z"`
+	if actual != expected {
+		t.Fatalf("expected %#v to serialize as %s, but got %s", u, expected, actual)
+	}
```
majewsky: We can use gg/assert here. It's also a dependency in go-api-declarations, so using it will not block the move later.

```suggestion
	if assert.ErrEqual(t, err, nil) {
		assert.Equal(t, string(buf), `"2024-03-15T12:30:45Z"`)
	}
```
Same on the other tests below.

docs/example-policy.json
```diff
@@ -42,10 +42,16 @@
...
     "v2:project:with_historical_usage": "rule:v2_cluster_viewer",
+    "v2:project:with_inactive": "rule:cluster_viewer",
+    "v2:project:with_subresources": "rule:v2_cluster_viewer",
     "v2:project:with_timing": "rule:v2_cluster_viewer",
     "v2:project:commitment_create": "rule:v2_project_admin",
+    "v2:project:commitment_delete": "rule:cluster_admin or (rule:'v2:domain_scope' and role:resource_admin) or (rule:'v2:project_scope' and role:admin)",
+    "v2:project:commitment_delete_admin": "rule:cluster_admin",
+    "v2:project:commitment_get": "rule:cluster_viewer or (rule:'v2:domain_scope' and role:resource_viewer) or (rule:'v2:project_scope' and role:member)",
+    "v2:project_commitment_get_public": "rule:cluster_viewer or role:resource_viewer or role:member",
+    "v2:project_commitment_get_unscoped": "rule:v2_cluster_viewer",
```
majewsky: The rule expressions appear to not have been updated to reflect recent refactors of the example policy.

```suggestion
    "v2:project:commitment_delete": "rule:v2_project_admin",
    "v2:project:commitment_delete_admin": "rule:v2_cluster_admin",
    "v2:project:commitment_get": "rule:v2_project_viewer",
    "v2:project_commitment_get_public": "rule:v2_cluster_viewer or role:resource_admin or role:resource_viewer",
    "v2:project_commitment_get_unscoped": "rule:v2_cluster_viewer",
```

docs/example-policy.json
```diff
@@ -42,10 +42,16 @@
     "v2:project:info": "rule:v2_project_viewer",
     "v2:project:report_single": "rule:v2_project_viewer",
     "v2:project:report_multiple": "rule:v2_domain_viewer",
-    "v2:project:with_subresources": "rule:v2_cluster_viewer",
     "v2:project:with_historical_usage": "rule:v2_cluster_viewer",
+    "v2:project:with_inactive": "rule:cluster_viewer",
```
majewsky:
```suggestion
    "v2:project:with_inactive": "rule:v2_cluster_viewer",
```

internal/api/api_v2/commitment_helpers.go
```diff
@@ -208,3 +214,56 @@ func analyzeCommitmentChangeResponse(resp liquid.CommitmentChangeResponse) error
...
+	// check auth
+	scope, err := p.checkProjectAccessByID(ctx, token, c.ProjectID, policyRule)
+	if err != nil {
+		return
+	}
+	canBeDeleted := isDeletable(token, c, p.timeNow)
+	if !canBeDeleted {
+		err = respondwith.CustomStatus(http.StatusForbidden, errNotDeletable)
+		return
+	}
```
majewsky: This check only applies to handleDeleteCommitment, not to any other prospective users of this helper, and should be moved hither.

internal/api/api_v2/commitment_delete.go
```diff
@@ -0,0 +1,116 @@
...
+	// validate request contents
+	cUUID := mux.Vars(r)["commitment_uuid"]
+	tx, err := p.DB.BeginTx(ctx, &sql.TxOptions{Isolation: sql.LevelRepeatableRead})
+	if err != nil {
+		return nil, err
+	}
+	defer sqlext.RollbackUnlessCommitted(tx)
+	c, azRes, scope, err := p.selectCommitmentIfPermitted(ctx, tx, sis, token, "v2:project:commitment_delete", liquid.CommitmentUUID(cUUID))
+	if err != nil {
+		return nil, err
```
majewsky, line 50: Proper REST etiquette suggests that DELETE should be idempotent, so we should respond with 204 if the commitment is already deleted:

```suggestion
		if errors.Is(err, errNoSuchCommitment) {
			return nil, nil // respond with 204 if commitment already deleted (DELETE should be idempotent)
		}
		return nil, err
```

internal/api/api_v2/commitment_delete.go
```diff
@@ -0,0 +1,116 @@
...
+	)
+
+	// validate request contents
+	cUUID := mux.Vars(r)["commitment_uuid"]
+	tx, err := p.DB.BeginTx(ctx, &sql.TxOptions{Isolation: sql.LevelRepeatableRead})
+	if err != nil {
+		return nil, err
+	}
+	defer sqlext.RollbackUnlessCommitted(tx)
+	c, azRes, scope, err := p.selectCommitmentIfPermitted(ctx, tx, sis, token, "v2:project:commitment_delete", liquid.CommitmentUUID(cUUID))
```
majewsky: Upon reading this, I was confused why we don't check for superseded/expired/deleted commitments. Of course, that's in this helper, which is fine, but the name could hint at it:
```suggestion
	c, azRes, scope, err := p.selectCommitmentIfAliveAndPermitted(ctx, tx, sis, token, "v2:project:commitment_delete", liquid.CommitmentUUID(cUUID))
```

majewsky (changes requested):
A few comments. Once again, I did not get through everything before having to leave for the day.

internal/apideclarations/apiv2/common/time_test.go
wagnerd3: Oops, somehow I was of the opinion we had to avoid this for go-api-declarations. But that must have been on a different repo or package or something...

docs/example-policy.json
wagnerd3: o.O Somehow I screwed up the rebase.

internal/api/api_v2/commitment_helpers.go
wagnerd3: Oh you are right about the single use, that's wrong, but the check must be re-used - I just forgot to build the re-use.
You started using `CanDeleteCommitment`, which uses the old policies. So I built this one here to replace it everywhere for v2 but forgot 😬

docs/operators/config.md
```diff
@@ -73,6 +78,7 @@ The v2 API uses the policy rules listed below.
 | `v2:project:report_single` | `domain_uuid`, `project_uuid` | `GET /{rates,resources}/v2/projects/:project_uuid` |
 | `v2:cluster:validation` | none | `GET /admin/mail/render` |
 | `v2:project:with_historical_usage` | none | `?with=historical_usage` in resource reports |
+| `v2:project:with_obsolete` | none | `with=inactive` in commitment APIs |
```
majewsky:
```suggestion
| `v2:project:with_obsolete` | none | `?with=obsolete` in commitment APIs |
```

internal/api/api_v2/project.go
```diff
@@ -132,12 +132,7 @@ func (p *v2Provider) checkAuthZForReportWithMultipleProjects(r *http.Request, to
 		if token.Context.Request == nil {
 			token.Context.Request = make(map[string]string, 1)
 		}
-		token.Context.Request["domain_uuid"] = queryDomainUUID
-		err := token.Enforce("v2:project:report_multiple")
-		if err != nil {
-			return nil, err
-		}
-		return reports_v2.NewDomainScope(r.Context(), queryDomainUUID, p.DB)
+		return p.checkDomainAccess(r.Context(), token, queryDomainUUID, "v2:domain:report_single")
```
majewsky: Copy-paste error:
```suggestion
		return p.checkDomainAccess(r.Context(), token, queryDomainUUID, "v2:domain:report_multiple")
```

internal/api/api_v2/commitment_get.go
```diff
@@ -0,0 +1,214 @@
...
+		if !ok {
+			token.Context.Request["domain_uuid"] = c.DomainUUID
+			token.Context.Request["project_uuid"] = string(c.ProjectUUID)
+			authorized = token.Check("v2:project:commitment_get")
+			authByProject[c.ProjectUUID] = authorized
+		}
+		if authorized {
+			return nil
+		}
+		c.ProjectUUID = ""
```
majewsky: I can see that there is a test for this, but I don't follow the logic of the implementation. We have already written the display form to `result.Commitments[-1]` at this point, and now we're redacting at the source in `c`? How does this work?

internal/api/api_v2/commitment_get.go
```diff
@@ -0,0 +1,214 @@
...
+		}
+	} else if projectUUID, ok := options.ProjectUUID.Unpack(); ok {
+		scope, err = p.checkProjectAccess(ctx, token, projectUUID, "v2:project:commitment_get")
+		if err != nil {
+			return nil, err
+		}
+	} else {
+		scope = reports_v2.ClusterScope{}
+	}
+	return scope, nil
```
majewsky: As a general style note, I see a lot of very strict adherence to the "either return a result or an error, never both" rule that I don't find all that necessary, given that callsites basically always do `if err != nil` immediately after calling a function that can return an error. This snippet is a particularly egregious example of how much boilerplate can be avoided by being more lenient in this regard:
```suggestion
	if domainUUID, ok := options.DomainUUID.Unpack(); ok {
		return p.checkDomainAccess(ctx, token, domainUUID, "v2:project:commitment_get")
	} else if projectUUID, ok := options.ProjectUUID.Unpack(); ok {
		return p.checkProjectAccess(ctx, token, projectUUID, "v2:project:commitment_get")
	} else {
		return reports_v2.ClusterScope{}, nil
	}
```

internal/api/api_v2/commitment_get_test.go
```diff
@@ -0,0 +1,252 @@
...
+		CreatedAt:           s.Clock.Now(),
+		UpdatedAt:           s.Clock.Now(),
+		ConfirmedAt:         Some(s.Clock.Now()),
+		CreatorUUID:         "dummy",
+		CreatorName:         "dummy",
+		ExpiresAt:           expiresAt,
+		CreationContextJSON: must.Return(json.Marshal(db.CommitmentWorkflowContext{Reason: db.CommitmentReasonCreate})),
+		Status:              liquid.CommitmentStatusConfirmed,
+	})
+	s.Clock.StepBy(time.Hour)
```
majewsky: I would prefer that this be inserted via the API instead (using `jsonmatch.CaptureField()` for the randomly generated UUID). Manually manipulating DB records like this carries the risk of testing states that are unreachable in practice, when new columns are added to the DB without updating all `MustDBInsert` calls in tests throughout the codebase.

This was the reason why I started with the commitment-create endpoint. I expected it to be required for all other testcases.

internal/api/api_v2/commitment_get_test.go
```diff
@@ -0,0 +1,252 @@
...
+
+	// error: permission denied
+	s.TokenValidator.Enforcer.AllowCommitmentGet = false
+	s.Handler.RespondTo(s.Ctx, "GET /resources/v2/commitments/00000000-0000-0000-0000-000000000001").
+		ExpectText(t, http.StatusForbidden, "Forbidden\n")
+	s.TokenValidator.Enforcer.AllowCommitmentGet = true
+
+	// error: permission denied as regular user for obsolete commitment
+	s.TokenValidator.Enforcer.ForbidWithObsolete = true
+	s.MustDBExec(`UPDATE project_commitments SET status = 'deleted' WHERE uuid = $1`, uuidOne)
```
majewsky: Following the same theme as the previous comment, this can be replaced by an actual API call to the DELETE endpoint. Matching my reasoning from the previous comment, this does move the commitment into an inconsistent state in the DB because several other fields (`deleted_at`, `updated_at`, etc.) are not updated appropriately. It just happens to not be relevant here, but that could change in the future.

internal/api/api_v2/commitment_get_test.go
```diff
@@ -0,0 +1,252 @@
...
+	s.Clock.StepBy(time.Hour)
+	testCommitment.CreatedAt = s.Clock.Now()
+	testCommitment.UpdatedAt = s.Clock.Now()
+	testCommitment.ExpiresAt = s.Clock.Now().AddDate(1, 0, 0).UTC()
+	testCommitment.ID = 0
+	testCommitment.UUID = "00000000-0000-0000-0000-000000000003"
+	testCommitment.ProjectID = projectDresdenID
+	testCommitment.Amount = 5
+	testCommitment.TransferStatus = limesresources.CommitmentTransferStatusPublic
+	testCommitment.TransferToken = Some(test.GenerateDummyTransferToken(1))
```
majewsky: Most of this paragraph can also be done via PUT+DELETE calls, though this bit will have to be done with direct DB manipulation initially because the respective PATCH endpoint is not implemented yet. We can leave a TODO in this spot to clean this up once the endpoint is implemented in a subsequent changeset.

internal/api/api_v2/commitment_helpers.go
majewsky: Unresolved because this check still lives in `selectCommitmentIfPermittedAndAlive` instead of `handleDeleteCommitment`.

majewsky (changes requested):
I also unresolved a previous comment that was not resolved properly.

internal/api/api_v2/commitment_helpers.go
wagnerd3: I would have noticed on the next API to re-use this method 🤞🏻

internal/api/api_v2/project.go
wagnerd3: we recognized this should be `v2:project:report_multiple`, because we are on a project API but want multiple projects.

internal/api/api_v2/commitment_get.go
wagnerd3: Spoiler: It does not 😅 I don't have a check in `JQModifiableJSONFixture` that a modification cannot be of no effect because the API of the package we are using for that is incapable of that. And I deleted `domain_uuid` instead of `domain_id` (because it's named inconsistently)...
I moved it around, so it's correct now.

internal/api/api_v2/commitment_get.go
wagnerd3: Maybe this is really something for an `agents.md` together with some other styling instructions that we adhere to? Not only for agents to read, also for humans?
Do we want to have a central `agents.md` for all of our repo's maybe, or one generated from `gomm`?

internal/api/api_v2/commitment_get_test.go
wagnerd3: Note: Some states I cannot produce via API yet, because the API is not there yet.

majewsky:
Please rebase, #944 has caused a merge conflict.

internal/api/api_v2/commitment_delete.go
```diff
@@ -0,0 +1,126 @@
...
+		sis = p.Cluster.SIC.GetSnapshot()
+	)
+
+	// validate request contents
+	cUUID := mux.Vars(r)["commitment_uuid"]
+	tx, err := p.DB.BeginTx(ctx, &sql.TxOptions{Isolation: sql.LevelRepeatableRead})
+	if err != nil {
+		return nil, err
+	}
+	defer sqlext.RollbackUnlessCommitted(tx)
```
majewsky, line 44: FYI, gg@v1.15.0 adds a `TxOptions` argument to `DB.WithinTransaction()`, so it can be used here instead: BeginTx, RollbackUnlessCommitted and Commit can all be replaced by
```go
err := p.DB.WithinTransaction(ctx, &sql.TxOptions{Isolation: sql.LevelRepeatableRead}, func(tx *sql.Tx) error {
	// ... everything between BeginTx and tx.Commit() ...
})
if err != nil {
	return nil, err
}
```

**Edit:** Since this ended up being the only note on this PR, I will approve and merge this to unblock future work. If you like, you can take the replacement of all instances of Begin/RollbackUnlessCommited/Commit with WithinTransaction as a Friday afternoon task. This is prep for pgx, in the same vein as #949.

majewsky (approved):

PR #919, opened 2026-07-17
V2 resource apis

wagnerd3:
Current state:
* I introduced some simplifications compared to `/rates`, e.g. direct access to entities by ID from `ServiceInfoCache`
* all apis implemented, tests written
* I think that It can serve all data as we imagined when brainstorming a few months ago and that all `?with=` options exist that we needed
* the performance is as follows: `/resources/v2/project` with our largest domain in our largest region is 10s computation, with all projects of the region it's 30s computation. That's not great, but acceptable, considering that in v1 the iteration through all domains produces some overhead that we save in v2
* my main idea was to have 1 SQL result per innermost result object so that we don't have to loop over the result multiple times - this makes the SQL quite long, possibly it could be shortened through clever application of theory of sets?

TODO:
* look at streaming the results of `resources/v2/project`, because this is quite a large data amount. the streaming implementation of v1 only works with list currently, not with object, but we can most likely make this work
* please look at the oslo policy, whether that is how we want to check for special with-flags in `resources/v2/project`
* check that I did not forget any functionality (params, returned data)
* probably, some logic can be deduplicated when looking at it long enough. E.g. the `setIn...Report` functions have a lot of duplication. Maybe we should implement a generic `func Set(value any)` on all structs and do this recursive? That does not help for values which are set on a higher level in the structure though, like `scraped_at, forbid_autogrowth, max_quota, ...`...
* look at SQL for optimization potential

### DIFF
--- a/docs/example-policy.json
+++ b/docs/example-policy.json
@@ -37,6 +37,9 @@
     "v2:project:info": "rule:'v2:domain:info' or rule:'v2:project:role'",
     "v2:project:report_single": "rule:'v2:project:report_multiple' or (rule:'v2:project:scope' and rule:'v2:project:role')",
     "v2:project:report_multiple": "rule:cluster_viewer or (rule:'v2:domain:scope' and rule:'v2:domain:role')",
+    "v2:project:with_subresources": "rule:cluster_admin or rule:cluster_viewer",
+    "v2:project:with_historical_usage": "rule:cluster_admin or rule:cluster_viewer",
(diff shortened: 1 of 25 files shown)


majewsky (approved):
We will take this as-is, and iterate on it in subsequent PRs. I have taken tons of notes that I will put into followup PRs, and which mostly align with what Daniel has pointed out in the PR description. I have made some adjustments with direct commits on this branch, so someone else will have to give a second stamp.

SuperSandro2000 (approved):

PR #926, opened 2026-08-04
replace Gorp with Oblast

majewsky:
For once, we don't have any huge refactorings in flight, so once again:

meme

At +579/-511 (excluding vendor), this is not quite as large as the respective Keppel change (+826/-730), and it's mostly more of the same. Most of the changes are a straight rewrite from Gorp functions into the
corresponding Oblast functions and methods, with the following exceptions:

- In tests, when a `ctx` argument was needed, I sometimes used `t.Context()` and sometimes the equivalent `s.Ctx`, each time keeping with the style used by the test file in question. At some point, we should harmonize this.
- In tests, I sometimes had to remove explicit ID values from record literals that are passed to `s.MustDBInsert()`. Gorp is lenient about ID fields being prefilled and just overwrites them with the actual ID, but Oblast refuses to do so in order to surface Insert() calls that were meant to be Update() and vice versa.

I also noticed a tiny implementation bug in the legacy v1 API, in a part that is not even used, but for completeness, I fixed this in a separate commit on this branch. Please forgive me for not adding a testcase for the fix.


internal/test/setup.go
```diff
@@ -459,7 +459,30 @@ func (s Setup) MustDBExec(query string, args ...any) {
...
+		must.SucceedT(s.t, db.ProjectServiceStore.Insert(s.Ctx, s.DB, r))
+	case *db.Rate:
+		must.SucceedT(s.t, db.RateStore.Insert(s.Ctx, s.DB, r))
+	case *db.Resource:
+		must.SucceedT(s.t, db.ResourceStore.Insert(s.Ctx, s.DB, r))
+	case *db.Service:
+		must.SucceedT(s.t, db.ServiceStore.Insert(s.Ctx, s.DB, r))
+	default:
+		panic(fmt.Sprintf("please extend the implementation of test.Setup.MustDBInsert() to support type %T", r))
+	}
```
SuperSandro2000, line 485: Can't we make this generic?

SuperSandro2000 (approved):

internal/test/setup.go
majewsky: Not until Go 1.27.

PR #938, opened 2026-08-20
api_v2: rework AuthZ and type Scope

majewsky:
Please look at the individual commit messages for reasoning.


docs/operators/config.md
```diff
@@ -48,6 +48,41 @@ These audit events can be sent to a RabbitMQ server which can then forward them
...
+| `v2:project:report_single` | `domain_uuid`, `project_uuid` | `GET /{rates,resources}/v2/projects/:project_uuid` |
+| `v2:cluster:validation` | none | `GET /admin/mail/render` |
+| `v2:project:with_historical_usage` | none | `?with=historical_usage` in resource reports |
+| `v2:project:with_subresources` | none | `?with=subresources` in resource reports |
+| `v2:project:with_timing` | none | `?with=timing` in resource reports |
+| `v2:meta:no_error_obfuscation` | none | see unobfuscated error messages in HTTP 5xx responses (usually restricted to cloud admins) |
+
+Note that:
+
+- The `v2:domain_info` rule may be evaluated multiple times if multiple `domain_uuid` options apply (based on the token presented by the requester).
```
wagnerd3:
```suggestion
- The `v2:domain:info` rule may be evaluated multiple times if multiple `domain_uuid` options apply (based on the token presented by the requester).
```

internal/api/reports_v2/info.go
```diff
@@ -32,27 +35,45 @@ var findAllowedResourcesQuery = sqlext.SimplifyWhitespace(`
...
+	originalRequest := token.Context.Request
+	defer func() {
+		token.Context.Request = originalRequest
+	}()
+
+	// case 2: domain admin (evaluated multiple times, see documentation on policy rules for details)
+	for _, domain := range []core.KeystoneDomain{
+		{Name: token.DomainScopeName(), UUID: token.DomainScopeUUID()},
+		{Name: token.ProjectScopeDomainName(), UUID: token.ProjectScopeDomainUUID()},
+		{Name: token.UserDomainName(), UUID: token.UserDomainUUID()},
```
wagnerd3: Oh I think this would be against the practice that unscoped tokens cannot be used for anything but generating other tokens. Or am I missing how you could get into this case with a valid, scoped token?
https://docs.openstack.org/keystone/latest/admin/tokens-overview.html

internal/api/api_v2/project.go
```diff
@@ -70,52 +76,74 @@ func (p *v2Provider) commonHandleGetResourcesProject(r *http.Request, token *gop
...
+		}
+		return reports_v2.NewDomainScope(r.Context(), queryDomainUUID, p.DB)
+	} else {
+		return reports_v2.ClusterScope{}, token.Enforce("v2:project:report_multiple")
+	}
+}
+
+// checkAuthZForReportWithSingleProject handles AuthZ for GET /{resources,rates}/v2/projects/:project_uuid.
+func (p *v2Provider) checkAuthZForReportWithSingleProject(ctx context.Context, token *gopherpolicy.Token) (reports_v2.Scope, error) {
+	return p.checkProjectAccess(ctx, token, liquid.ProjectUUID(token.Context.Request["project_uuid"]), "v2:project:report_single")
```
wagnerd3, line 148: We refer here to `checkProjectAccess`, which then does a select on `db.Project` on it's own, assembling a `ProjectScope`. `NewProjectScope` is not used anywhere. Should we call `NewProjectScope` from `checkProjectAccess` or get rid of it? I know in `checkProjectAccess` we need the domain-select individually to do the `token.Enforce` shenanigans but leaving that duplication feels a bit strange.

internal/api/reports_v2/scope.go
```diff
@@ -4,157 +4,136 @@
...
+	if err != nil {
+		if errors.Is(err, sql.ErrNoRows) {
+			return none, fmt.Errorf("referential integrity error: project %s references unknown domain with ID = %d", project.UUID, project.DomainID)
 		}
+		return none, err
 	}
 
-	return s, nil
+	if domainUUID.IsSomeAnd(is.DifferentFrom(domain.UUID)) {
+		return none, fmt.Errorf("inconsistent NewScope() invocation: got domainUUID = %q and projectUUID = %q, but that project actually belongs to domain %q with UUID = %q",
```
wagnerd3: Should this be a `CustomStatus` of `BadRequest` or something?

wagnerd3 (changes requested):

internal/api/reports_v2/info.go
majewsky: Mostly it's that I don't see large risk from allowing this, since policies are usually written to be very narrow in what they approve. And if this case never matches what the others don't, whatever.

internal/api/api_v2/project.go
majewsky: NewProjectScope will likely be needed for commitment list, where in the project-scoped case, `projectUUID` must be given and `domainUUID` might be given. So I think it's useful to keep around.

wagnerd3 (approved):

PR #944, opened 2026-08-28
replace deep cloning in ServiceInfoCache with ConstMap

wagnerd3:
includes the following improvements, too:
* some missing defense-in-depth checks for race conditions where we select in project-entities and try to find the referenced service-info entity
* remove multi-level-return methods from ServiceInfoReader interface, because subsequent calls are cheap now without clones and therefore we can just chain the calls. E.g. getResources()=getServices()+getResourcesForType()
* remove unnecessary "ok" bools from signature of constMap-returning entries of ServiceInfoReader interface, because most of the time we just iterate the result, and the ok variable can be easily replaced. E.g. ok=getResourcesForType().Len()==0

### DIFF
--- a/internal/collector/scrape.go
+++ b/internal/collector/scrape.go
@@ -474,11 +474,7 @@ func (c *Collector) writeRateScrapeResult(ctx context.Context, task projectScrap
 	defer sqlext.RollbackUnlessCommitted(tx)
 
 	// For updating project_rates, we need to find the project_rates' rate_name
-	rateSlice := make([]db.Rate, 0, rates.Len())
-	for _, rate := range rates.All() {
-		rateSlice = append(rateSlice, rate)
-	}
-	ratesByID := db.RateByIDIndex.Index(rateSlice)
+	ratesByID := db.RateByIDIndex.Index(slices.Collect(rates.Values()))
 
 	// update existing project_rates entries
 	rateExists := make(map[liquid.RateName]bool)
--- a/internal/collector/scrape.go
+++ b/internal/collector/scrape.go
@@ -118,7 +118,7 @@ func (c *Collector) discoverScrapeTask(ctx context.Context, labels prometheus.La
 	}
 	// if the above check succeeded, this should never fail because the SIC is updated after the
 	// LiquidConnection is initialized.
-	if !c.Cluster.SIC.HasService(serviceType) {
+	if _, ok := c.Cluster.SIC.GetSnapshot().GetServiceForType(serviceType); !ok {
 		return projectScrapeTask{}, fmt.Errorf("no data found in ServiceInfoCache for type %s", serviceType)
 	}
 
@@ -253,9 +253,8 @@ func extractRateData(report liquid.ServiceUsageReport) (result map[liquid.RateNa
 
 func (c *Collector) writeResourceScrapeResult(ctx context.Context, task projectScrapeTask, serviceType db.ServiceType, dbProject db.Project, resourceData liquid.ServiceUsageReport, sis core.ServiceInfoSnapshot) error {
 	service, sExists := sis.GetServiceForType(serviceType)
-	resources, _ := sis.GetResourcesForType(serviceType)     // can have no resources
-	azResources, _ := sis.GetAZResourcesForType(serviceType) // can have no az_resources
-	if !sExists {                                            // defense in depth: this snapshot is taken immediately after saving the ServiceInfo
+	resources := sis.GetResourcesForType(serviceType) // can have no resources
+	if !sExists {                                     // defense in depth: this snapshot is taken immediately after saving the ServiceInfo
 		return fmt.Errorf("no data found in ServiceInfoCache for type %s", serviceType)
 	}
 
@@ -325,12 +324,6 @@ func (c *Collector) writeResourceScrapeResult(ctx context.Context, task projectS
 	}
 
 	// For inserting the project_az_resources, we need to find the project_az_resources availabilityZone
-	azResourcesByID := make(map[db.AZResourceID]db.AZResource)
-	for _, azResourcesByAZ := range azResources {
-		for _, azr := range azResourcesByAZ {
-			azResourcesByID[azr.ID] = azr
-		}
-	}
 	projectAZResourcesByAZResourceID, err := db.ProjectAZResourceByAZResourceIDIndex.IndexFrom(
 		db.ProjectAZResourceStore.Select(ctx, tx,
 			`SELECT pazr.* FROM project_az_resources pazr JOIN az_resources azr ON PAZR.az_resource_id = azr.id JOIN resources r ON azr.resource_id = r.id WHERE r.service_id = $1 AND pazr.project_id = $2`,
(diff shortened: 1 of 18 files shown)


internal/util/constmap_test.go
```diff
@@ -0,0 +1,91 @@
...
+
+func TestConstMapAll(t *testing.T) {
+	m := NewConstMap(map[string]int{"a": 1, "b": 2, "c": 3})
+
+	seen := maps.Collect(m.All())
+
+	assert.Equal(t, len(seen), 3)
+	assert.Equal(t, seen["a"], 1)
+	assert.Equal(t, seen["b"], 2)
+	assert.Equal(t, seen["c"], 3)
```
majewsky: Regular reminder that `assert.Equal` is not limited to `comparable` types anymore:
```suggestion
	assert.Equal(t, maps.Collect(m.All()), map[string]int{"a": 1, "b": 2, "c": 3})
```

internal/core/service_info_cache.go
```diff
@@ -693,71 +561,108 @@ func (s *ServiceInfoCache) signalInvalidation() {
...
+			if sType != st {
+				rates[sType] = maps.Collect(rateByName.All())
+			}
+		}
+		for sType, catByID := range s.data.categories.All() {
+			if sType != st {
+				categories[sType] = maps.Collect(catByID.All())
+			}
+		}
 	}
```
majewsky: If `serviceType.IsSome()`, this does unnecessary work unwrapping all the lower-level ConstMaps, only to then rewrap them later on. We should first fill with the data from the query, and then merge the existing ConstMap values for the service types that are not invalidated.

internal/core/service_info_cache.go
```diff
@@ -693,71 +561,108 @@ func (s *ServiceInfoCache) signalInvalidation() {
...
 		return fmt.Errorf("while reading services for type(s) %v: %w", serviceType, err)
 	}
-	maps.Copy(s.data.services, servicesByType)
-	for _, svc := range servicesByType {
-		s.data.servicesByID[svc.ID] = svc
+	for _, dbService := range dbServices {
+		services[dbService.Type] = dbService
 	}
 
-	// TODO: simplify queries below by using the ServiceID from the record that we just pulled
```
majewsky, line 716: This TODO should stay because it was not addressed.

internal/collector/scrape.go
```diff
@@ -478,7 +474,11 @@ func (c *Collector) writeRateScrapeResult(ctx context.Context, task projectScrap
 	defer sqlext.RollbackUnlessCommitted(tx)
 
 	// For updating project_rates, we need to find the project_rates' rate_name
-	ratesByID := db.RateByIDIndex.Index(slices.Collect(maps.Values(rates)))
+	rateSlice := make([]db.Rate, 0, rates.Len())
+	for _, rate := range rates.All() {
+		rateSlice = append(rateSlice, rate)
+	}
+	ratesByID := db.RateByIDIndex.Index(rateSlice)
```
majewsky:
```suggestion
	ratesByID := db.RateByIDIndex.Index(slices.Collect(rates.Values()))
```

majewsky (changes requested):

internal/core/service_info_cache.go
wagnerd3: I removed the TODO, because I wanted to implement the optimization, but then I forgot 🙃 Will include it now and leave out the TODO.

internal/core/service_info_cache.go
wagnerd3: I decided to work with a normal `map` on level 0, for easy collection, with `ConstMap` on the lower levels, so I can copy over the old entries 1:1 where possible.

internal/core/service_info_cache.go
```diff
@@ -567,119 +566,125 @@ func (s *ServiceInfoCache) InvalidateService(ctx context.Context, serviceType Op
...
+		if currentService != path.ServiceType {
+			// flush
+			resources[currentService] = util.NewConstMap(resourcesForCurrentService)
+			resourcesForCurrentService = make(map[liquid.ResourceName]db.Resource)
+			currentService = path.ServiceType
 		}
-		resources[path.ServiceType][path.ResourceName] = dbResource
+		resourcesForCurrentService[path.ResourceName] = dbResource
 	}
+	resources[currentService] = util.NewConstMap(resourcesForCurrentService)
```
majewsky: This breaks if `len(dbResources) == 0`, since then we would fill `resources[""]`.
```suggestion
	if currentService != "" {
		resources[currentService] = util.NewConstMap(resourcesForCurrentService)
	}
```
Same for all the other similar loops below. I'm assuming that the tests do not catch this because we only ask for specific service types, so the entry for `""` is never seen.

majewsky (changes requested):

internal/core/service_info_cache.go
wagnerd3: I built a check for this case and fixed it.

internal/core/service_info_cache.go
```diff
@@ -693,113 +560,197 @@ func (s *ServiceInfoCache) signalInvalidation() {
...
+	for st, liquidConfiguration := range s.config.Liquids {
+		areaMappingPlain[st] = liquidConfiguration.Area
 	}
 
-	// now we fill the cache for the invalidated services again
-	servicesByType, err := db.ServiceByTypeIndex.IndexFrom(db.ServiceStore.SelectWhere(ctx, s.DB, `type = $1 OR $1 IS NULL`, serviceType))
+	// we start with empty maps, get the data we want from the database and then possibly copy over the old values
+	services := make(map[db.ServiceType]db.Service)
+	resources := make(map[db.ServiceType]util.ConstMap[liquid.ResourceName, db.Resource])
+	azResources := make(map[db.ServiceType]util.ConstMap[liquid.ResourceName, util.ConstMap[liquid.AvailabilityZone, db.AZResource]])
```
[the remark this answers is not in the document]
wagnerd3: Actually, `liquid.AvailabilityZone` is more correct here, because we are dealing internally with data from liquid - this is not customer facing. So I switched everything to `liquid` now.

internal/core/service_info_cache.go
```diff
@@ -693,113 +560,197 @@ func (s *ServiceInfoCache) signalInvalidation() {
...
+		services[dbService.Type] = dbService
+		serviceIDs = append(serviceIDs, dbService.ID)
 	}
 
-	// TODO: simplify queries below by using the ServiceID from the record that we just pulled
-
-	resourcesByPath, err := db.ResourceByPathIndex.IndexFrom(
-		db.ResourceStore.Select(ctx, s.DB, `SELECT r.* FROM resources r JOIN services s ON r.service_id = s.id WHERE s.type = $1 OR $1 IS NULL`, serviceType),
-	)
+	dbResources, err := db.ResourceStore.Select(ctx, s.DB, `SELECT r.* FROM resources r WHERE r.service_id = ANY($1) OR CARDINALITY($1) = 0 ORDER BY path`, pq.Array(serviceIDs)).Collect()
```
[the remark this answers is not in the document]
wagnerd3: This is comment is wrong. The number of services cannot change without a limes restart and initial `saveServiceInfoToDB`, because there is multiple jobs which get instantiated once per service. If we leave out the `CARDINALITY() = 0` snippet, we make it impossible to reload all serviceTypes, which happens from time to time when the connection to the database breaks.

internal/core/service_info_cache.go
```diff
@@ -693,113 +560,197 @@ func (s *ServiceInfoCache) signalInvalidation() {
...
+		resourcesForCurrentService   map[liquid.ResourceName]db.Resource
+		azResourcesForCurrentService map[liquid.ResourceName]map[liquid.AvailabilityZone]db.AZResource
+		ratesForCurrentService       map[liquid.RateName]db.Rate
+		categoriesForCurrentService  map[db.CategoryID]db.Category
+	)
+	currentService = ""
+	for _, dbResource := range dbResources {
+		path := dbResource.Path
+		if currentService != path.ServiceType {
+			// flush
```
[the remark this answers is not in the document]
wagnerd3: I added a check to prevent these actions on the first run.

majewsky (approved):

PR #945, opened 2026-09-01
implement complex commitment conversions

wagnerd3:
This branch allows for complex unit conversions. The functionality is described in the `config.md`.

tl;dr:
* per resource, multiple conversion rules are now possible
* conversion weights now get a unit. The new lemma is that when converting commitments, `weight * unit * amount = const`. Because the weight has a unit itself, conversions from a byte-based to a non-byte based unit and vise versa are possible.
* conversion direction can be limited, a rule can be set to only be source for conversion
* the old `/commitment` APIs signature is not changed, we could/should probably expose the units in the v2 API to make things more transparent

### DIFF
--- a/internal/api/commitment_test.go
+++ b/internal/api/commitment_test.go
@@ -59,7 +59,7 @@ var testCommitmentsJSONWithoutMinConfirmDate = testCommitmentsJSON.
 	Modify("del(.liquids.first.commitment_behavior_per_resource)").
 	Modify(`.liquids.second.commitment_behavior_per_resource[0].value.durations_per_domain[0].value += ["3 hours"]`).
 	Modify("del(.liquids.second.commitment_behavior_per_resource[0].value.min_confirm_date)")
-var testConvertCommitmentsJSON = httptest.NewJQModifiableJSONString(`
+var testConvertCommitmentsJSON = httptest.NewJQModifiableJSONString(test.RemoveCommentsFromJSON(`
 	{
 		"areas": { "third": { "display_name": "Third" }, "fourth": { "display_name": "Fourth" }},
 		"liquids": {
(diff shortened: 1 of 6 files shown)


internal/api/commitment_test.go
```diff
@@ -59,7 +59,7 @@ var testCommitmentsJSONWithoutMinConfirmDate = testCommitmentsJSON.
 	Modify("del(.liquids.first.commitment_behavior_per_resource)").
 	Modify(`.liquids.second.commitment_behavior_per_resource[0].value.durations_per_domain[0].value += ["3 hours"]`).
 	Modify("del(.liquids.second.commitment_behavior_per_resource[0].value.min_confirm_date)")
-var testConvertCommitmentsJSON = httptest.NewJQModifiableJSONString(`
+var testConvertCommitmentsJSON = httptest.NewJQModifiableJSONString(test.RemoveCommentsFromJSON(`
```
majewsky: Nitpick: Judging by the diff, RemoveCommentsFromJSON does not seem to be used.

internal/api/commitment_test.go
```diff
@@ -2301,6 +2340,101 @@ func Test_ConvertCommitments(t *testing.T) {
...
+	// with sourceAmount=3: (3 * 12) / 5 = 36 / 5 = 7 (rounded down from 7.2)
+	oldassert.HTTPRequest{
+		Method:       http.MethodPost,
+		Path:         "/v1/domains/uuid-for-germany/projects/uuid-for-berlin/commitments/7/convert",
+		Body:         req("third", "capacity_does_not_fit", 3, 7),
+		ExpectStatus: http.StatusAccepted,
+	}.Check(t, s.Handler)
+
+	convertedCommitment = must.ReturnT(db.ProjectCommitmentStore.SelectOneWhere(s.Ctx, s.DB, `id = 10`))(t)
+	assert.Equal(t, convertedCommitment.Amount, uint64(7))
```
majewsky: Do we also want to check that no split commitment was created for the leftover? We could require that `id = 11` yields `sql.ErrNoRows`.

I suppose that there is no way to have a leftover even theoretically because rounding down from `7.2` to `7` on the target side corresponds to only trace amounts of the original amount `3`.

We could change these commitments by a factor of 20, i.e. have 60 on the source and 140 on the target instead, which would still constitute some rounding (the precise conversion would be 144), but then there could theoretically be a leftover (`60 - 140/2.4 = 1.666...`).

internal/core/commitment_behavior.go
```diff
@@ -45,32 +47,32 @@ func (b CommitmentBehavior) Validate(path string, occupiedConversionIdentifiers
 			errs.Addf("invalid value: %s.until_percent may not be bigger than 100", path)
 		}
 	}
-	if conversionRule, ok := b.ConversionRule.Unpack(); ok {
-		identifier = conversionRule.Identifier
-		if slices.Contains(occupiedConversionIdentifiers, conversionRule.Identifier) {
-			errs.Addf("invalid value: %s.conversion_rule.identifier values must be restricted to a single serviceType, but %q is already used by another serviceType", path, conversionRule.Identifier)
+	for identifier := range b.ConversionRules {
+		if slices.Contains(occupiedConversionIdentifiers, identifier) {
+			errs.Addf("invalid value: %[1]s.conversion_rules[%[2]q] identifier must be restricted to a single serviceType, but %[2]q is already used by another serviceType", path, identifier)
```
majewsky: We need to remove this restriction to allow commitment conversions from `swift/capacity` to `keppel/capacity`. This is not related to the main change in this PR, but if it's a quick and easy change that does not break anything anywhere else, I would like for us to do it right now.

internal/core/commitment_behavior.go
```diff
@@ -135,39 +137,102 @@ func (b ScopedCommitmentBehavior) ForV2API(now time.Time) Option[resourcesv2.Com
 	return Some(result)
 }
 
+// ConversionRuleIdentifier is an explicit type to increased readability of the code
```
majewsky:
```suggestion
// ConversionRuleIdentifier is an explicit type to increase readability of the code.
// It appears in type [CommitmentBehavior].
```

internal/api/commitment_test.go
wagnerd3: I get the point about the check and will built that in. but it's not possible to achieve this situation (a leftover of more than 1 source unit) with the weights the test is currently using. I will alter the weights to prove this check.

majewsky (changes requested):
Request from UI: The `allow_rounding` attribute needs to be visible in the GetCommitmentConversions endpoint, so the UI can perform appropriate validation.

**Edit:** I have been told this was a misunderstanding. The UI does not need it.

majewsky (approved):

PR #949, opened 2026-09-08
replace sqlext.ForeachRow cases with oblast.Select

wagnerd3:
Note: I did not replace the code inside of the `SelectSeveralValues` function, that I copied from Keppel, because AFAIK there is no suitable replacement available in oblast yet.


internal/api/commitment.go
```diff
@@ -38,6 +38,11 @@ import (
 	"github.com/sapcc/limes/internal/util"
 )
 
+type azResourceLoc struct {
+	ID   db.AZResourceID   `db:"id"`
+	Path db.AZResourcePath `db:"path"`
+}
```
majewsky, line 44: If we have this type on the package level, we can also have the Store next to it, thus avoiding the small cost of pulling the plan from the cache each time:

```suggestion
type azResourceLoc struct {
	ID   db.AZResourceID   `db:"id"`
	Path db.AZResourcePath `db:"path"`
}

var azResourceLocStore = oblast.MustNewStore[azResourceLoc](oblast.PostgresDialect())
```

internal/collector/consistency.go
```diff
@@ -60,36 +60,30 @@ var (
 	`)
 )
 
-func (c *Collector) checkConsistency(_ context.Context, _ prometheus.Labels) error {
+type missingProjectServiceRecord struct {
+	ProjectID db.ProjectID `db:"project_id"`
+	ServiceID db.ServiceID `db:"service_id"`
+}
```
majewsky, line 66: Same here re: keeping the Store in a global variable.

internal/collector/metrics.go
```diff
@@ -85,29 +84,22 @@ func (c *AggregateMetricsCollector) Collect(ch chan<- prometheus.Metric) {
...
-		}
-
-		if c.Cluster.SIC.GetSnapshot().GetResourcesForType(serviceType).Len() > 0 {
+	type scrapedAtAggregateRecord struct {
+		ServiceType  db.ServiceType `db:"service_type"`
+		MinScrapedAt *time.Time     `db:"min_scraped_at"`
+		MaxScrapedAt *time.Time     `db:"max_scraped_at"`
+	}
+	err := oblast.MustNewStore[scrapedAtAggregateRecord](oblast.PostgresDialect()).Select(context.Background(), c.DB, scrapedAtAggregateQuery).Foreach(func(r scrapedAtAggregateRecord) error {
+		if c.Cluster.SIC.GetSnapshot().GetResourcesForType(r.ServiceType).Len() > 0 {
```
majewsky: Only tangentially related to the change in this PR, but particularly from looking at the red side of this diff, this function looks like it was not correctly updated when we merged resource scraping and rate scraping. This condition should be removed so that we can see `limes_{newest,oldest}_scraped_at` for services with only rates and no resources. I'm assuming that we deleted the branch for writing the respective metrics for `rates_scraped_at`, without considering that the other branch would need to be updated.

internal/collector/metrics.go
```diff
@@ -147,8 +139,8 @@ type CapacityCollectionMetricsCollector struct {
 // CapacityCollectionMetricsInstance describes a single project service for which collection
 // metrics are submitted. It appears in type CapacityCollectionMetricsCollector.
 type CapacityCollectionMetricsInstance struct {
-	ServiceType       db.ServiceType
-	SerializedMetrics string
+	ServiceType       db.ServiceType `db:"service_type"`
+	SerializedMetrics string         `db:"serialized_metrics"`
 }
```
majewsky, line 142: Another instance of a type that can have its Store in a global variable.

internal/collector/metrics.go
```diff
@@ -245,8 +233,8 @@ type UsageCollectionMetricsCollector struct {
 // metrics are submitted. It appears in type UsageCollectionMetricsCollector.
 type QuotaCollectionMetricsInstance struct {
 	Project           core.KeystoneProject
-	ServiceType       db.ServiceType
-	SerializedMetrics string
+	ServiceType       db.ServiceType `db:"type"`
+	SerializedMetrics string         `db:"serialized_metrics"`
 }
```
majewsky, line 238: And another one.

internal/collector/sync_quota_to_backend.go
```diff
@@ -119,65 +120,59 @@ func (c *Collector) performQuotaSync(ctx context.Context, srv db.ProjectService,
 	startedAt := c.MeasureTime()
 
 	sis := c.Cluster.SIC.GetSnapshot()
-	if _, sExists := sis.GetServiceForType(serviceType); !sExists {
+	resources := sis.GetResourcesForType(serviceType)
+	if resources.Len() == 0 {
 		return fmt.Errorf("no data found in ServiceInfoCache for %s", serviceType)
 	}
```
majewsky, line 126: This will throw errors for services without resources. I'm fine with having a shortcut to avoid useless work in this job when there is nothing to do because there are no resources, but we should still clear away the quota-sync flag and return a nil error.

internal/datamodel/resource_demand.go
```diff
@@ -44,44 +45,40 @@ var (
 	`))
 )
 
+type resourceDemandRecord struct {
+	AZ                 limes.AvailabilityZone `db:"az"`
+	Usage              uint64                 `db:"usage"`
+	ActiveCommitments  uint64                 `db:"confirmed"`
+	PendingCommitments uint64                 `db:"pending"`
+	Topology           liquid.Topology        `db:"topology"`
+}
```
majewsky, line 54: Another type that can have its Store as a global variable.

internal/reports/cluster.go
```diff
@@ -43,15 +44,39 @@ var clusterReportQuery1 = sqlext.SimplifyWhitespace(db.ExpandEnumPlaceholders(`
 	 GROUP BY s.type, r.name, azr.az
 `))
 
+type clusterResourceUsageRecord struct {
```
majewsky, line 47: I'm not gonna flag all of these in this package, but the "Store in global variable" thing applies to all of these.

internal/reports/domain.go
```diff
@@ -114,81 +106,74 @@ func GetDomains(cluster *core.Cluster, domainID *db.DomainID, now time.Time, dbi
...
-			&domainID, &dbServiceType, &dbResourceName, &az,
-			&quota, &usage, &unusedCommitments, &uncommittedUsage,
-			&backendQuota, &infiniteBackendQuota, &physicalUsage, &showPhysicalUsage, &minScrapedAt, &maxScrapedAt,
-		)
-		if err != nil {
-			return err
-		}
-		if domains[domainID] == nil {
+
+	type resourceRecord struct {
```
majewsky: `reports/cluster.go` declares its record types in global scope, and this file does so in function scope. It should be consistent across all files of the same package. I prefer the former style because it groups the record type declaration with the respective report query.

internal/reports/project.go
```diff
@@ -86,9 +86,12 @@ func GetProjectResources(ctx context.Context, cluster *core.Cluster, domain db.D
...
 	// but will yield no results at all in the other queries)
-	whereStr, whereArgs := db.BuildSimpleWhereClause(fields, 0)
-	queryStr := `SELECT * FROM projects p WHERE ` + whereStr
-	allProjects, err := db.ProjectStore.Select(ctx, dbi, queryStr, whereArgs...).Collect()
+	selectWhereFields := make(map[string]any, len(fields))
+	for k, v := range fields {
+		selectWhereFields[strings.TrimPrefix(k, "p.")] = v
+	}
+	whereStr, whereArgs := db.BuildSimpleWhereClause(selectWhereFields, 0)
+	allProjects, err := db.ProjectStore.SelectWhere(ctx, dbi, whereStr, whereArgs...).Collect()
```
majewsky: I see what you're going for, but I find it easier to follow before this change. Select() is not bad per se; SelectWhere() is to be preferred because it is usually more succinct and more efficient. In this case, it's neither.

internal/reports/project.go
```diff
@@ -418,9 +411,12 @@ func GetProjectRates(ctx context.Context, cluster *core.Cluster, domain db.Domai
...
```
majewsky: Same as above.

majewsky (changes requested):
A lot of these are the same.

internal/collector/metrics.go
wagnerd3: At some intermediate point in time after merging the 2 scrapes, we had the situation that we were even writing the same value twice - once for rates and once for resources. I think back then this check was still useful, but you are right, now it's obsolete.

internal/collector/sync_quota_to_backend.go
wagnerd3: Mhhh, I think that this constellation cannot happen: `quota_desynced_at` is set from `acpq`, which iterates the resources. So a service without resources cannot end up in `projectsWithUpdatedQuota`, and thus should not be subject to `performQuotaSync`. In the past, the resource join was in the sql statement, so no resources would have led to an empty result there and no entries iterated. But I think actually throwing an error for a service having to perform acpq without any resources is fine due to the mentioned mechanis?

internal/reports/domain.go
wagnerd3: I put it in global now.

internal/reports/project.go
wagnerd3: I reverted to the old style.

internal/reports/project.go
wagnerd3: I reverted to the old style.

internal/collector/sync_quota_to_backend.go
```diff
@@ -119,65 +120,59 @@ func (c *Collector) performQuotaSync(ctx context.Context, srv db.ProjectService,
...
 		if !ok {
-			// race condition: selected projectResource while the resource got deleted
-			// other projectResources/ resources might still exist, so we don't abort!
+			// race condition: az_resource deleted while the select was executed
 			return nil
 		}
+		resource := must.BeOK(sis.GetResourceForID(azResource.ResourceID))
 		if !resource.HasQuota {
 			return nil
 		}
```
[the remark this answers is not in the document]
majewsky: Not relevant. We will always have a snapshot that's internally consistent, so if the az_resources record was already missing from it, we skipped it cleanly and will not fail on a missing resources record.

majewsky (approved):

PR #952, opened 2026-09-11
PATCH /resources/v2/commitment

wagnerd3:
reviewing by commit is easiest, because I improved the other commitment tests a bit in the 2nd one.

### DIFF
--- a/docs/operators/config.md
+++ b/docs/operators/config.md
@@ -68,6 +68,7 @@ The v2 API uses the policy rules listed below.
 | `v2:project:commitment_get` | `domain_uuid`, `project_uuid` | `GET /resources/v2/commitments(/:commitment_uuid)?` as regular user |
 | `v2:project:commitment_get_public` | none | `GET /resources/v2/commitments?only_public=true` |
 | `v2:project:commitment_get_unscoped` | none | `GET /resources/v2/commitments` without `domain_uuid` and `project_uuid` filters |
+| `v2:project:commitment_patch` | `domain_uuid`, `project_uuid` | `PATCH /resources/v2/commitments/:commitment_uuid` |
 | `v2:cluster:info` | none | `GET /{rates,resources}/v2/info` for whole cluster and `?with=info` in cluster-level reports |
 | `v2:domain:info` | `domain_uuid` | `GET /{rates,resources}/v2/info` for domain scope and `?with=info` in domain-level reports (based on domain of token scope, supporting both project-scoped and domain-scoped tokens) |
 | `v2:project:info` | `domain_uuid` | `GET /{rates,resources}/v2/info` for project scope and `?with=info` in project-level reports |
--- a/internal/api/api_v2/commitment_delete_test.go
+++ b/internal/api/api_v2/commitment_delete_test.go
@@ -4,7 +4,6 @@
 package api_v2_test
 
 import (
-	"encoding/json"
 	"errors"
 	"net/http"
 	"testing"
@@ -14,20 +13,19 @@ import (
 	"github.com/sapcc/go-bits/httptest"
 	. "go.xyrillian.de/gg/option"
 
-	limesresources "github.com/sapcc/go-api-declarations/limes/resources"
 	"github.com/sapcc/go-api-declarations/liquid"
 	"github.com/sapcc/go-bits/easypg"
 	"github.com/sapcc/go-bits/must"
 	"go.xyrillian.de/gg/assert"
+	"go.xyrillian.de/gg/jsonmatch"
 
 	"github.com/sapcc/limes/internal/db"
 	"github.com/sapcc/limes/internal/test"
 )
 
 // Helper function for a successful commitment deletion.
-func deleteCommitmentAndExpectSuccess(t *testing.T, s test.Setup, tr *easypg.Tracker, liquidHandlesCommitments bool, ccr liquid.CommitmentChangeRequest) {
+func deleteCommitmentAndExpectSuccess(t *testing.T, s test.Setup, tr *easypg.Tracker, liquidHandlesCommitments bool, uuid string, dbID int, ccr liquid.CommitmentChangeRequest) {
 	t.Helper()
-	uuid := "00000000-0000-0000-0000-000000000001"
 	path := "/resources/v2/commitments/" + uuid
 	s.Handler.RespondTo(s.Ctx, "DELETE "+path).ExpectStatus(t, http.StatusNoContent)
 
@@ -37,7 +35,7 @@ func deleteCommitmentAndExpectSuccess(t *testing.T, s test.Setup, tr *easypg.Tra
 	}
 
 	// assertions
-	tr.DBChanges().AssertEqualf("UPDATE project_commitments SET status = 'deleted', deleted_at = %[1]d, updated_at = %[1]d WHERE id = 1 AND uuid = '00000000-0000-0000-0000-000000000001' AND transfer_token = NULL;", s.Clock.Now().Unix())
+	tr.DBChanges().AssertEqualf("UPDATE project_commitments SET status = 'deleted', deleted_at = %[1]d, updated_at = %[1]d WHERE id = %[2]d AND uuid = '%[3]s' AND transfer_token = NULL;", s.Clock.Now().Unix(), dbID, uuid)
 	s.Auditor.ExpectEvents(t, cadf.Event{
 		Action:      "delete",
 		Outcome:     "success",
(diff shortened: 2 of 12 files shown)


internal/api/api_v2/commitment_patch.go
```diff
@@ -0,0 +1,187 @@
...
+		InfoVersion: must.BeOK(sis.GetServiceForType(azRes.Path.ServiceType)).LiquidVersion,
+		ByProject: map[liquid.ProjectUUID]liquid.ProjectCommitmentChangeset{
+			scope.Project.UUID: {
+				ProjectMetadata: datamodel.LiquidProjectMetadataFromDBProject(scope.Project, scope.Domain),
+				ByResource: map[liquid.ResourceName]liquid.ResourceCommitmentChangeset{
+					azRes.Path.ResourceName: {
+						TotalConfirmedBefore:  stats.TotalConfirmed,
+						TotalConfirmedAfter:   stats.TotalConfirmed,
+						TotalGuaranteedBefore: stats.TotalGuaranteed,
+						TotalGuaranteedAfter:  stats.TotalGuaranteed, // TODO: change when introducing "guaranteed" commitments
```
majewsky:
```suggestion
						TotalGuaranteedAfter:  stats.TotalGuaranteed,
```

internal/api/api_v2/commitment_patch.go
```diff
@@ -0,0 +1,187 @@
...
+		c.ExpiresAt = newExpiresAt
+	}
+	c.UpdatedAt = now
+
+	// checking the response is only relevant for extending durations
+	stats, err := getCommitmentStats(p.DB, c.ProjectID, c.AZResourceID)
+	if err != nil {
+		return none, err
+	}
+	ccr := liquid.CommitmentChangeRequest{
```
majewsky, line 118: We should skip this entirely for `TransferStatus.IsSome()` since there is no LIQUID-relevant change.

internal/api/api_v2/commitment_patch_test.go
```diff
@@ -0,0 +1,370 @@
...
+		mockLiquid.LastCommitmentChangeRequest = liquid.CommitmentChangeRequest{}
+	}
+
+	// we only expect the audit event and call to the liquid here
+	// (the DB effect is not checked here; the caller will take care of that afterwards)
+	path := "/resources/v2/commitments/" + uuid
+	s.Handler.RespondTo(ctx, "PATCH "+path, httptest.WithJSONBody(request)).
+		ExpectJSON(t, http.StatusAccepted, expected)
+	if noop {
+		s.Auditor.ExpectEvents(t)
```
majewsky: To clarify that the variadic argument is empty, I recommend writing
```suggestion
		s.Auditor.ExpectEvents(t, nil...)
```

internal/api/api_v2/commitment_delete_test.go
```diff
@@ -14,20 +13,19 @@ import (
...
 	"go.xyrillian.de/gg/assert"
+	"go.xyrillian.de/gg/jsonmatch"
 
 	"github.com/sapcc/limes/internal/db"
 	"github.com/sapcc/limes/internal/test"
 )
 
 // Helper function for a successful commitment deletion.
-func deleteCommitmentAndExpectSuccess(t *testing.T, s test.Setup, tr *easypg.Tracker, liquidHandlesCommitments bool, ccr liquid.CommitmentChangeRequest) {
+func deleteCommitmentAndExpectSuccess(t *testing.T, s test.Setup, tr *easypg.Tracker, liquidHandlesCommitments bool, uuid string, dbID int, ccr liquid.CommitmentChangeRequest) {
```
majewsky, line 27: We have a lot of `uuid string` arguments across these tests, which I would all like to use the more specific type:
```suggestion
func deleteCommitmentAndExpectSuccess(t *testing.T, s test.Setup, tr *easypg.Tracker, liquidHandlesCommitments bool, uuid liquid.CommitmentUUID, dbID int, ccr liquid.CommitmentChangeRequest) {
```

internal/api/api_v2/commitment_patch.go
```diff
@@ -0,0 +1,187 @@
...
+		ctx  = r.Context()
+		sis  = p.Cluster.SIC.GetSnapshot()
+	)
+
+	// validate request contents
+	req, err := parseRequestBodyAs[resourcesv2.CommitmentPatchRequest](r)
+	if err != nil {
+		return none, err
+	}
+	cUUID := mux.Vars(r)["commitment_uuid"]
```
majewsky: I vaguely remember that this might also apply to existing handlers, but it is weird that this is not casted into the right type immediately:
```suggestion
	cUUID := liquid.CommitmentUUID(mux.Vars(r)["commitment_uuid"])
```

internal/api/api_v2/commitment_patch_test.go
```diff
@@ -0,0 +1,370 @@
...
+						insertStmt = fmt.Sprintf(`INSERT INTO project_commitments (id, uuid, project_id, az_resource_id, status, amount, duration, created_at, creator_uuid, creator_name, confirmed_at, expires_at, transfer_status, transfer_token, creation_context_json, transfer_started_at, updated_at) VALUES (1, '%[1]s', 3, 2, 'confirmed', 10, '1 hour', 0, 'uuid-for-alice', 'alice@Default', 0, 3600, '%[3]s', '%[4]s', '{"reason": "create"}', %[2]d, %[2]d);
+						`, uuid, s.Clock.Now().UTC().Unix(), newStatus, transferToken)
+					}
+
+					// easypg sorts by key string: "id = 1 AND uuid = '...' AND transfer_token = <val>"
+					if oldTokenSQL < newTokenSQL {
+						tr.DBChanges().AssertEqual(deleteStmt + insertStmt)
+					} else {
+						tr.DBChanges().AssertEqual(insertStmt + deleteStmt)
+					}
```
majewsky, line 252: I understand what is happening and why, but still, yikes. I took note of the suboptimal behavior of easypg.Tracker re: changes to key fields (i.e. the way that it generates INSERT+DELETE in weird situations like this one), but was not motivated to go in and fix it because it was never quite as bad as this.

This is not required to merge this PR, but I'm flagging this here in case you want to take a stab at maybe improving easypg to render an UPDATE instead when the only key column that changes is not part of the primary key.

internal/api/api_v2/commitment_patch_test.go
```diff
@@ -0,0 +1,370 @@
...
+					tr.DBChanges().AssertEmpty()
+
+					oldStatus = newStatus
+				})
+			}
+
+			// check extension of duration
+			s.Clock.StepBy(time.Minute)
+			expectedJSON["duration"] = "2 hours"
+			expectedJSON["expires_at"] = expiresAt.Add(1 * time.Hour).Format(time.RFC3339)
```
majewsky: I was confused by this at first. Maybe rename `expiresAt` to `initialExpiresAt`?

docs/operators/config.md
```diff
@@ -68,6 +68,7 @@ The v2 API uses the policy rules listed below.
 | `v2:project:commitment_get` | `domain_uuid`, `project_uuid` | `GET /resources/v2/commitments(/:commitment_uuid)?` as regular user |
 | `v2:project:commitment_get_public` | none | `GET /resources/v2/commitments?only_public=true` |
 | `v2:project:commitment_get_unscoped` | none | `GET /resources/v2/commitments` without `domain_uuid` and `project_uuid` filters |
+| `v2:project:commitment_patch` | `domain_uuid`, `project_uuid` | `PATCH /resources/v2/commitments/:commitment_uuid` |
```
majewsky: We usually use more standard verbs instead of the HTTP method, e.g. `POST .../commitments/new` has `...:commitment_create` instead of `commitment_post` or `commitment_new`.
```suggestion
| `v2:project:commitment_update` | `domain_uuid`, `project_uuid` | `PATCH /resources/v2/commitments/:commitment_uuid` |
```
This policy can also be reused for split/merge to avoid policy rule sprawl, since I don't see a reason why any particular user would be allowed to access the PATCH endpoint, but not also do a merge or split, or vice versa. In a similar vein, the receive and renew endpoints can reuse the `commitment_create` policy rule.

majewsky (changes requested):

internal/api/api_v2/commitment_patch.go
wagnerd3: It was in the delete-API. I will change it there, too.

internal/api/api_v2/commitment_patch_test.go
wagnerd3: I noted this already, but figured I would ignore it for now. If we were to change this, I would have to touch many existing places and adjust the test-assertions, so I want to do this in a separate PR 100%.

majewsky (approved):

PR #954, opened 2026-09-14
replace Begin/BeginTx with WithinTransaction

majewsky:
All code that is within a transaction gains one level of indentation, so viewing this diff with `?w=1` is urgently advised:

```
$ git diff origin/master... --stat | tail -n1
 15 files changed, 1595 insertions(+), 1723 deletions(-)
$ git diff origin/master... --stat -w | tail -n1
 15 files changed, 251 insertions(+), 379 deletions(-)
```

**Update:** This PR was initially just for v2 API, but I ended up also changing all other instances. The change to the v1 commitment API moves several transaction starts closer to the first time the transaction is actually used, in order to reduce the number of variables that need to be predeclared outside of the WithinTransaction callback.


internal/api/commitment.go
majewsky, line 1906: Flagging this for completeness sake: We do `p.DB.QueryRow()` here even though we have an open `tx` on the red side of this diff. I chose to not change that because this query is very unlikely to yield a nonrepeatable read, and instead applied the same logic as with most other transaction starts: moving the BeginTransaction call to right in front of the first actual use of `tx`.

wagnerd3 (approved):

PR #955, opened 2026-09-16
/v2/commitments/split and /merge

wagnerd3:
These 2 APIs _should_ be quite straight forward to review, because there is not as much logic involved.


internal/api/api_v2/commitment_merge.go
```diff
@@ -0,0 +1,206 @@
...
+			return err
+		}
+
+		// in this loop we do 3 things to avoid unnecessary iterations:
+		// * validate status of mergeable commitments
+		// * build merged commitment: sum amounts, use latest expiresAt and its duration
+		// * collect IDs for context
+		commitmentIDs := make([]db.ProjectCommitmentID, len(commitments))
+		commitmentUUIDs := make([]liquid.CommitmentUUID, len(commitments))
+		liquidCommitments := make([]liquid.Commitment, 0, len(commitments)+1)
```
majewsky: Tiny naming nitpick: IMO it reads better to describe what these representations are for. "Liquid commitment" does make sense, but it tripped me up.
```suggestion
		commitmentsForCCR := make([]liquid.Commitment, 0, len(commitments)+1)
```

internal/api/api_v2/commitment_merge.go
```diff
@@ -0,0 +1,206 @@
...
+		creationContext := db.CommitmentWorkflowContext{
+			Reason:                 db.CommitmentReasonMerge,
+			RelatedCommitmentIDs:   commitmentIDs,
+			RelatedCommitmentUUIDs: commitmentUUIDs,
+		}
+		creationBuf, err := json.Marshal(creationContext)
+		if err != nil {
+			return err
+		}
+		mergedCommitment := &db.ProjectCommitment{
```
majewsky: Avoid pointer types unless truly needed. In this case, we have a lot of instances where `*mergedCommitment` is used below, which can be simplified.
```suggestion
		mergedCommitment := db.ProjectCommitment{
```

internal/api/api_v2/commitment_split.go
```diff
@@ -0,0 +1,216 @@
...
+			return err
+		}
+		c := commitments[0]
+		deletable := isDeletable(token, c, now)
+
+		// validate sum of amounts
+		newSum := uint64(0)
+		for _, amount := range req.Amounts {
+			newSum += amount
+		}
```
majewsky, line 65: Should check for overflow.
```suggestion
		newSum := uint64(0)
		for _, amount := range req.Amounts {
			if newSum + amount < newSum {
				kaboom()
			}
			newSum += amount
		}
```
You should be able to construct a failing test on the current implementation where a commitment with amount `10` can be split into two halfs of size `5 + (1 << 63)` (though it is possible that it would then later be denied e.g. while judging the CommitmentChangeRequest against available capacity).

internal/api/api_v2/commitment_split.go
```diff
@@ -0,0 +1,216 @@
...
+		if c.TransferStatus != limesresources.CommitmentTransferStatusNone {
+			return respondwith.CustomStatus(http.StatusBadRequest, errNoTransferSplit)
+		}
+
+		// prep modifications to commitments
+		splitCommitments, err := buildSplitCommitments(c, req.Amounts, now)
+		if err != nil {
+			return err
+		}
+		splitLiquidCommitments := make([]liquid.Commitment, len(splitCommitments))
```
majewsky: Same as above:
```suggestion
		splitCommitmentsForCCR := make([]liquid.Commitment, len(splitCommitments))
```

internal/api/api_v2/commitment_split.go
```diff
@@ -0,0 +1,216 @@
...
+		c.SupersedeContextJSON = Some(json.RawMessage(buf))
+		c.Status = liquid.CommitmentStatusSuperseded
+		c.SupersededAt = Some(now)
+		c.UpdatedAt = now
+		err = db.ProjectCommitmentStore.Update(ctx, tx, c)
+		if err != nil {
+			return err
+		}
+
+		return nil
```
majewsky:
```suggestion
		return db.ProjectCommitmentStore.Update(ctx, tx, c)
```

majewsky (changes requested):
Notes for implementation.

internal/api/api_v2/commitment_split.go
wagnerd3: I got rid of the additional variable for the split commitments and merged all `commitmentsForCCR`

internal/api/api_v2/commitment_split.go
wagnerd3: ~If we do this overflow check, I will also add an overflow check for merging commitments.~
This was a BS idea, because we can never insert enough commitments which would still fit into the capacity of `bigint`.

internal/api/api_v2/commitment_merge_test.go
```diff
@@ -0,0 +1,294 @@
...
+	s.Handler.RespondTo(ctx, "POST "+path, httptest.WithJSONBody(request)).
+		ExpectJSON(t, http.StatusCreated, expected)
+	ae := auditEventFunc()
+	s.Auditor.ExpectEvents(t, cadf.Event{
+		Action:      "create",
+		Outcome:     "success",
+		Reason:      cadf.Reason{ReasonType: "HTTP", ReasonCode: "201"},
+		RequestPath: path,
+		Target:      ae,
+	})
```
majewsky: `auditEventFunc` is an unnecessarily vague name.
```suggestion
	s.Auditor.ExpectEvents(t, cadf.Event{
		Action:      "create",
		Outcome:     "success",
		Reason:      cadf.Reason{ReasonType: "HTTP", ReasonCode: "201"},
		RequestPath: path,
		Target:      getExpectedResource(),
	})
```

majewsky: The way that we provide a function to render the cadf.Resource is a bit convoluted, given that the effort is just to avoid writing the ExpectEvents() call at the callsite. I would rather we make the ExpectEvents() call shorter:

```go
func buildCreateEvent(path string, target cadf.Resource) cadf.Event {
	return cadf.Event{
		Action:      "create",
		Outcome:     "success",
		Reason:      cadf.Reason{ReasonType: "HTTP", ReasonCode: "201"},
		RequestPath: path,
		Target:      target,
	}
}
```

And then after each mergeCommitmentsAndExpectSuccess call, remove the auditEventFunc setup and instead say:

```go
s.Auditor.ExpectEvents(t, buildCreateEvent("/resources/v2/commitments/merge", cadf.Resource {
	...
}))
```

And same for split.

majewsky (changes requested):
One more style nitpick on the tests, then this is good to go.

internal/api/api_v2/commitment_merge_test.go
wagnerd3: That would not work exactly as described because we need the `liquid.CommitmentUUID`s to be captured from the `expectedJSON` to be passed to the `liquid.CommitmentChangeRequest` as well. If we do the audit thingy outside, we would have to do the assertion for the `CCR` also outside. This is possible, but IMHO doing 2 function calls after each action is more complexity than just passing the function?

majewsky: The ExpectEvents call would take place after mergeCommitmentsAndExpectSuccess returns, so CaptureField has already had its effect and this concern does not apply.

majewsky: Ah, I did not see that `ae` is passed there. The names `ae` and `auditEventFunc` are super-weird. Can we go with `target` and `getAuditTarget`?

majewsky (approved):

PR #956, opened 2026-09-17
/v2/commitments/{uuid}/renew

wagnerd3:
@VoigtS To review the styling, it's probably beneficial to look at the split, merge, delete, patch APIs.


internal/api/api_v2/commitment_renew.go
```diff
@@ -0,0 +1,198 @@
...
+	resourcesv2 "github.com/sapcc/limes/internal/apideclarations/apiv2/resources"
+	"github.com/sapcc/limes/internal/audit"
+	"github.com/sapcc/limes/internal/datamodel"
+	"github.com/sapcc/limes/internal/db"
+)
+
+// As per the API spec, commitments can be renewed 90 days in advance at the earliest.
+const commitmentRenewalPeriod = 90 * 24 * time.Hour
+
+func (p *v2Provider) handleRenewCommitment(r *http.Request, token *gopherpolicy.Token) (result resourcesv2.Commitment, _ error) {
```
VoigtS, line 36: The `delete` API returns: `(any, error)`, while the other endpoints return `(resourcesv2.Commitment, error)`. Is there a reason for the delete API to behave differently?
The `delete` tests also don't check for the `nil` or `struct` response type.

internal/api/api_v2/commitment_renew.go
```diff
@@ -0,0 +1,198 @@
...
+	)
+
+	// validate request contents
+	req, err := parseRequestBodyAs[resourcesv2.CommitmentRenewRequest](r)
+	if err != nil {
+		return none, err
+	}
+	err = p.DB.WithinTransaction(ctx, &sql.TxOptions{Isolation: sql.LevelRepeatableRead}, func(tx *gsql.Tx) error {
+		cUUID := liquid.CommitmentUUID(mux.Vars(r)["commitment_uuid"])
+		commitments, azRes, scope, err := p.selectCommitmentsIfPermittedAndAlive(ctx, tx, sis, token, "v2:project:commitment_create", []liquid.CommitmentUUID{cUUID})
```
VoigtS, line 53: Not a PR code critique, but a question.
According to `selectCommitmentsIfPermittedAndAlive`. We have:
```
 case errors.Is(err, sql.ErrNoRows):
			err = respondwith.CustomStatus(http.StatusNotFound, errNoSuchCommitment)
```

Shouldn't we obfuscate if a commitment is not found? I'm thinking of scenarios where users could probe the commitment consumption behaivor or something similar.

VoigtS (changes requested):

internal/api/api_v2/commitment_renew.go
wagnerd3: This is because the handler func does not care about the returned value, when the desired `ResponseCode` is 204. And writing `any` and then using `nil` as value all the time is quite short.

>  The delete tests also don't check for the nil or struct response type.

I don't get this point? 204 implies there is no response body and AFAIK the body you try to write when sending 204 gets ignored automatically.

VoigtS: Ah, yes, I didn't take the `handlerFunc` behavior into account. Fair enough.

internal/api/api_v2/commitment_renew.go
wagnerd3: I discussed this with @majewsky and our stance was that the commitment is identified via UUID in the new API. To guess a UUID for a commitment correctly, you need to try very often. In the old API, you are right - the `ID` used there is in fact very unsafe for such "attacks". But even there the question is, what you want to do with knowing a commitment ID. You can only use such an ID against Limes, not against any other openstack system, and all the APIs for it probe authorization - so there is an attack surface, but it's very small (limited to the one commitment API file).

VoigtS (approved):

